PR #23, opened 2025-07-11
bug(#22): fixed yaml file

maxonfjvipon:
Closes: #22


test/resources/xpath-packs/template-match-does-not-start-with-double-slash.yaml
```diff
@@ -26,7 +26,7 @@ found:
   positions: [ ]
 input: |-
   <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="2.0">
-    <xsl:template match="/objects//o">
+    <xsl:template match="/objects//o/o[1]/o[1]">
```
maxonfjvipon, line 29: @maxonfjvipon it's not good

PR #73, opened 2026-01-29
added use SPDX for license

### LINKED ISSUE #47: use spdx for licence
Let's use SPDX for licence. Check this as an example: https://github.com/objectionary/eo2js/blob/master/.github/workflows/reuse.yml

alexachenko:
@maxonfjvipon please review
fixes #47

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -33,6 +33,7 @@ path = [
     "package.json",
     "patches/grunt-mocha-cli+7.0.0.patch",
     "renovate.json",
+    ".npmignore",
 ]
 precedence = "override"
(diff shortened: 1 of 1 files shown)


LICENSES/MIT.txt
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2025 Max Trunnikov
```
maxonfjvipon: @alexachenko let's fix a year here, should be 2025-2026. I'll add check for year in license later

alexachenko: @maxonfjvipon year fixed, REUSE.toml file added

PR #97, opened 2026-02-02
added rule UnusedNamedTemplate

### LINKED ISSUE #93: UnusedNamedTemplate check is not implemented
Let's do it. See the description here: https://gandhimukul.tripod.com/xslt/xslquality.html
Related to #24

alexachenko:
fixes #93
[Here](https://stackoverflow.com/questions/42317523/how-do-i-import-templates-from-another-stylesheet) I found of information about using templates in another stylesheets and another files with `xsl:import`. Should I rewrite xpath to check it too? Or it's bad practice too to write named templates in one stylesheet and use in the other?


maxonfjvipon:
@alexachenko this check should be implemented in other linting group where we will check multiple files at the same time. Current xpath checker is supposed to find violations only inside a single xsl. I would recommend to make a new ticket with this info you've found. We'll get back to it when we're done with this xpath checker

maxonfjvipon (approved):

PR #117, opened 2026-03-03
changed xsl-file

### LINKED ISSUE #116: test xsl is properly formatted
Check this: https://github.com/maxonfjvipon/xslint/blob/8c84019a1fef963732711fde19423eb6cefa3d7a/test/resources/xsl-packs/xsl-with-some-violations.xsl#L25-L26

In general it's valid, but looks ugly, should be: `<xsl:when ... />`

alexachenko:
@maxonfjvipon please review
fixes #116

### DIFF
--- a/test/xslint.test.js
+++ b/test/xslint.test.js
@@ -27,12 +27,12 @@ describe('xslint', function() {
     const expected = [
       'Processed files: 1, defects found 7',
       '(25:9) Don\'t use empty content for instructions like \'xsl:for-each\' \'xsl:if\' \'xsl:when\' etc. (template-match-empty-content-in-instructions)',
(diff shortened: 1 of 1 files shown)


test/resources/xsl-packs/xsl-with-some-violations.xsl
```diff
@@ -23,22 +23,22 @@
       </xsl:if>
       <xsl:choose>
         <xsl:when test="price &lt; 300">
-                </xsl:when>
+        </xsl:when>
```
maxonfjvipon: @alexachenko there's no need in closing `<\when>` tag since there's nothing in between. Should be just single `<xsl:when test="..."/>`. The same story for all other similar tags

alexachenko:
@maxonfjvipon please review

maxonfjvipon (approved):

PR #150, opened 2026-04-08
added rule BlankNestedIf

### LINKED ISSUE #140: blank nested `<xsl:if>` are not caught
This is what I have:

```
<xsl:if test=".!=''">
  <xsl:if test="position() &gt; 1">
    <xsl:text>foo</xsl:text>
  </xsl:if>
</xsl:if>
```

This code must be replaced by:

```
<xsl:if test=".!='' and position() &gt; 1">
  <xsl:text>foo</xsl:text>
</xsl:if>
```

Let's create a new check for xslint that would catch such situations.

alexachenko:
@maxonfjvipon 
fixes #140


src/resources/template-match-blank-nested-if.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:if[(count(*) = 1) and (count(xsl:if) = 1)]
+severity: warning
+message: Incorrectly using nested 'xsl:if'. Combine nested 'xsl:if' into one or use several nodes in parent's 'xsl:if' node.
```
maxonfjvipon: @alexachenko 

> use several nodes in parent's 'xsl:if' node.

I don't get it, what does it mean?

alexachenko: @maxonfjvipon
if it contains other nodes besides the child 'xsl:if', then the rule does not apply. Therefore, to neutralize the violation, user either need to combine the parent and child 'xsl:if', or add other child nodes to the parent 'xsl:if', if appropriate. for example, like [here](https://github.com/maxonfjvipon/xslint/pull/150/changes#diff-6238a0a570063de42811356fae7abd270257299216c4a2b28ec5ba723f629c91R12-R17)

maxonfjvipon: @alexachenko ok, I got the idea but the message still looks a bit confusing to me. I think this rule does not need to _think_ about other possible options except `if(if(..))`. So let's just keep the message just like this:
'Recommended to combine nested `xsl:if` nodes into single one joining conditions with `and` operator'

alexachenko: @maxonfjvipon can i try to merge it?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

PR #151, opened 2026-04-08
added rule template-match-not-using-output

alexachenko:
@maxonfjvipon please review 
fixes #142


src/resources/template-match-not-using-output.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:stylesheet[count(xsl:output) = 0]
+severity: warning
```
maxonfjvipon: @alexachenko let's use `error` severity here

src/resources/template-match-not-using-output.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:stylesheet[count(xsl:output) = 0]
```
maxonfjvipon: @alexachenko I think it's not necessary to use `//` in front of `xsl:stylesheet` here. I think it's allowed to use `xsl:stylesheet` only at the top level but I can be wrong here. Let's investigate. Maybe we can make a new check from it

maxonfjvipon (changes requested):

src/resources/template-match-not-using-output.yaml
alexachenko: @maxonfjvipon you are right, `xsl:stylesheet` should be outermost
besides, [here](https://www.w3.org/TR/xslt.html#section-Stylesheet-Structure) i found:
>A stylesheet is represented by an xsl:stylesheet element in an XML document. xsl:transform is allowed as a synonym for xsl:stylesheet.

and some examples:
```xslt
<xsl:stylesheet
  id = id
  extension-element-prefixes = tokens
  exclude-result-prefixes = tokens
  version = number>
  <!-- Content: ([xsl:import](https://www.w3.org/TR/xslt.html#element-import)*, top-level-elements) -->
</xsl:stylesheet>
```

```xslt
<xsl:transform
  id = id
  extension-element-prefixes = tokens
  exclude-result-prefixes = tokens
  version = number>
  <!-- Content: ([xsl:import](https://www.w3.org/TR/xslt.html#element-import)*, top-level-elements) -->
</xsl:transform>
```

as far as I remember, we didn't write anywhere in xpath `xsl:transform`

maxonfjvipon: @alexachenko we don't care about `xsl:transform` for now. Let's stick with `xsl:stylesheet` and change xpath. Also let's create a new ticket. We should raise an error if there's `xsl:styleshet` somewhere but not in the root

maxonfjvipon:
@alexachenko there's a conflict to be resolved

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

Issue #137, opened 2026-03-31
False positive: template-match-not-using-schema-types on XSLT 1.0 files


yegor256:
The check `template-match-not-using-schema-types` fires on XSLT 1.0 stylesheets, which by definition cannot use XML Schema types (xs:string, etc.) - those are XSLT 2.0+ features.

Example file triggering the warning:
```xml
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
  <xsl:template match="/">
    <!-- ... -->
  </xsl:template>
</xsl:stylesheet>
```

Warning:
```
The stylesheet is not using any of the built-in Schema types (xs:string etc.), when working in XSLT 2.0 mode. (template-match-not-using-schema-types)
```

The check should either:
1. Not fire on XSLT 1.0 files (check the `version` attribute)
2. Be renamed to clarify it's for XSLT 2.0+ only

alexachenko:
@yegor256 we added new [test](https://github.com/maxonfjvipon/xslint/blob/master/test/resources/xpath-packs/template-match-not-using-schema-types-in-xslt-1-0.yaml) for this [check](https://github.com/maxonfjvipon/xslint/blob/master/src/resources/checks/template-match-not-using-schema-types.yaml).
but we couldn't find this bug.
can you please send your xsl code for comparison? I want to see if the warning can be caused by something else.

yegor256:
@alexachenko maybe I was wrong about the bug

PR #153, opened 2026-04-09
added test not-using-schema-types in xslt 1.0

alexachenko:
@maxonfjvipon please review 
#137

### DIFF
--- a/test/resources/xpath-packs/template-match-not-using-schema-types-in-xslt-1-0.yaml
+++ b/test/resources/xpath-packs/template-match-not-using-schema-types-in-xslt-1-0.yaml
@@ -7,12 +7,11 @@ found:
   positions: [ ]
 input: |-
   <?xml version="1.0"?>
-  <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" version="2.0">
+  <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
     <xsl:template match="/objects/o/o[1]/o[1]">
       <xsl:copy-of select="."/>
     </xsl:template>
     <xsl:template match="/objects/o/o[1]/o[2]">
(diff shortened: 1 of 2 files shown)


maxonfjvipon (commented):
@alexachenko I'm not sure I understand the purpose of the PR. How's the PR related to the issue? What do you test here?

alexachenko:
@maxonfjvipon it was written in the issue that when using `version="1.0"` violation is triggered, but I added a test with:
```xslt
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
```
and the violation does not appear (as it should be).

maxonfjvipon:
@alexachenko maybe the xml in the issue was not really correct and does not contain places that can be replaced with schema types. Let's add some to the test

maxonfjvipon:
@alexachenko this check `template-match-not-using-schema-types` check if we use any of built in schema types when working with XSLT 2.0
```xpath
/xsl:stylesheet[@version = '2.0'][not(some $x in .//@* satisfies contains($x, 'xs:'))]
```

However, this check is triggered in XSL with `<xsl:stylesheet version="1.0">`. Let's add test with `version="1.0"` if it does not exis

alexachenko:
@maxonfjvipon i added this test in the PR: https://github.com/maxonfjvipon/xslint/pull/153/changes#diff-43b8be1142015489705125fc443b6027faf66c86f6492b30f842fedebf6b670d
or do you mean different kind of test?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

alexachenko:
@maxonfjvipon 
also [here](https://github.com/maxonfjvipon/xslint/issues/137#issue-4185771013) is written
> The check template-match-not-using-schema-types fires on XSLT 1.0 stylesheets, which by definition cannot use XML Schema types (xs:string, etc.) - those are XSLT 2.0+ features.

we can add a new ticket about prohibiting the use of schema types in xslt 1.0, right?

maxonfjvipon:
@rultor merge

PR #162, opened 2026-04-16
bug(#154): rewrite all violation messages to follow a consistent two-sentence pattern

### LINKED ISSUE #154: violation message does not follow any pattern
It would be better all the violation messages follow the same pattern. It's better to avoid the word `recommend` because it brings ambiguity. I think the final message should be strong. `It's must be...`, `it's prohibited...`, something like this. So in general we should have 2 sentences. In the first one we say that something is wrong or there's a problem. In the second one we should say how this problem must be solved.

maxonfjvipon:
## Summary

- Each message now follows a strict two-sentence pattern: **[What is wrong.]** **[What must be done.]**
- Removed weak language: "Recommended", "You might be", "Please relook", "You can", "Don't"
- Fixed formatting issues: missing periods, trailing spaces, inconsistent capitalisation
- Updated the one integration test that asserted exact message strings

## Test plan

- [x] All 150 tests pass (`grunt`)

Closes #154

🤖 Generated with [Claude Code](https://claude.com/claude-code)


maxonfjvipon:
@rultor merge

PR #173, opened 2026-04-18
bug(#12): check xsl:import sort order by @href

### LINKED ISSUE #12: the order of `<xsl:import href=""/>` is not checked
The order of imports is not checked:

```xsl
<?xml version="1.0" encoding="UTF-8"?>
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:foo="https://foo.com" version="2.0">
  <xsl:import href="/z.xsl"/>
  <xsl:import href="/x.xsl"/>
  <xsl:import href="/a.xsl"/>
   ...
</xsl:stylesheet>
```

Currently, `xslint` does not complain about it. I suggest to to make it `WARNING`, if imports are not sorted alphabetically.

maxonfjvipon:
## Summary

- Adds `template-match-unsorted-imports` rule that warns when `<xsl:import>` elements are not sorted alphabetically by `@href`
- Adds motive doc with incorrect/correct XSL examples
- Adds two test fixtures: one with unsorted imports (2 violations) and one with sorted imports (0 violations)

Closes #12

## Test plan

- [ ] `npm test` passes with both new fixtures
- [ ] Running xslint against an XSL with unsorted imports produces a `warning`
- [ ] Running xslint against an XSL with sorted imports produces no output for this rule

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/resources/checks/template-match-unsorted-imports.yaml
+++ b/src/resources/checks/template-match-unsorted-imports.yaml
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:import[preceding-sibling::xsl:import/@href > @href]
+severity: warning
+message: The 'xsl:import' elements are not sorted alphabetically by '@href'. Sort imports to improve readability.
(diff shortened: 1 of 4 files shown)


maxonfjvipon:
@rultor merge

PR #174, opened 2026-04-19
added check for recursion used

### LINKED ISSUE #167: False negative: `template-match-unused-function` in recursion
for example:
```xslt
<?xml version="1.0"?>
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" version="2.0">
  <xsl:output encoding="UTF-8" method="xml"/>
  <xsl:function name="my:factorial" as="xs:integer">
    <xsl:param name="number" as="xs:integer"/>
    <xsl:sequence
            select="if ($number=0) then 1 else $number * my:factorial($number-1)"/>
  </xsl:function>
</xsl:stylesheet>
```
we do not have `template-match-unused-function` violation here, although we do not use `my:factorial`. This happens because `my:factorial` is a recursion, and we use it in the function definition itself.

it is necessary to change the `xpath` so that it excludes the definition of the function itself from the calculation of its use.

alexachenko:
@maxonfjvipon please review
fixes #167

### DIFF
--- a/src/resources/checks/template-match-unused-function.yaml
+++ b/src/resources/checks/template-match-unused-function.yaml
@@ -4,4 +4,3 @@
 xpath: //xsl:function[not(some $x in //(@match | @select) except (./(@match | @select) | .//*/(@match | @select)) satisfies contains($x, concat(@name,'(')))]
 severity: warning
 message: A stylesheet function is never called in any match or select expression. Remove it or call it.
-
--- a/src/resources/checks/template-match-unused-function.yaml
+++ b/src/resources/checks/template-match-unused-function.yaml
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //xsl:function[not(some $x in //(@match | @select) satisfies contains($x, @name))]
+xpath: //xsl:function[not(some $x in //(@match | @select) except (./(@match | @select) | .//*/(@match | @select)) satisfies contains($x, concat(@name,'(')))]
 severity: warning
 message: A stylesheet function is never called in any match or select expression. Remove it or call it.
+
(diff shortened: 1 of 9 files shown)


src/resources/checks/template-match-unused-function.yaml
```diff
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //xsl:function[not(some $x in //(@match | @select) satisfies contains($x, @name))]
+xpath: //xsl:function[not(some $x in //(@match | @select) except (./(@match | @select) | .//*/(@match | @select)) satisfies contains($x, concat(@name,'(')))]
```
maxonfjvipon: The `except` clause is more complex than needed. `xsl:function` never carries `@match` or `@select` itself, so `./(@match | @select)` always yields an empty sequence — dead weight. The whole clause simplifies to:

```xpath
except .//(@match | @select)
```

Making the full expression:

```xpath
//xsl:function[not(some $x in //(@match | @select) except .//(@match | @select) satisfies contains($x, concat(@name,'(')))]```

File: test/resources/xpath-packs/template-match-unused-recursion.yaml [hunk 2]
```diff
@@ -0,0 +1,20 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-unused-function
+found:
+  amount: 1
```
maxonfjvipon, line 6: Good test for direct recursion. Mutual recursion is not covered: if `my:a` calls `my:b` and `my:b` calls `my:a`, neither called from outside, both should be flagged — but the current XPath won't catch them because each name appears in the other function's `@select`. Worth a separate issue/test.

maxonfjvipon (changes requested):

test/resources/xpath-packs/template-match-unused-recursion.yaml
alexachenko: Ok! i added this: https://github.com/maxonfjvipon/xslint/issues/175#issue-4292144217

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

PR #178, opened 2026-04-22
added `xcop` tests for motives

### LINKED ISSUE #170: `xsl` parts in motives are not checked by `xcop`
Similar to [this](https://github.com/maxonfjvipon/xslint/issues/108) ticket we can make a new unit test that would take XSLs from motives and pass them through `xcop` in order to keep the XSL quality

alexachenko:
@maxonfjvipon please review
fixes #170 
also i have a question
can we leave  `console.log(ex.stdout.toString())` in code?
```js
const execCmd = function(command, args, print) {
  try {
    return execSync(
      `${command} ${args.join(' ')}`,
      {
        timeout: 120000,
        windowsHide: true,
        stdio: print ? null : 'ignore'
      }
    ).toString()
  } catch (ex) {
    console.log(ex.stdout.toString())
    throw ex
  }
}
```
it can be useful for correcting identified errors after launching motives-check.test.js and xpath-linter.test.js

UPD: I decided to delete this because grunt detected [warnings](https://github.com/maxonfjvipon/xslint/actions/runs/24788365913/job/72538626494?pr=178#step:6:301)  when working due to these changes. xslint.test.js


maxonfjvipon:
@alexachenko you can leave it, you can just not to print these warnings into regular console pipeline. Or print them on demand, but it's up to you. I would recommend to leave it

test/motives-check.test.js
```diff
@@ -0,0 +1,38 @@
...
+ * Yaml test packs.
+ * @type {Array<String>}
+ */
+const MOTIVES = allFilesFrom(path.resolve(__dirname, '../src/resources/motives'))
+
+describe('motives-check', function() {
+  MOTIVES.forEach((motive) => {
+    const lines = fs.readFileSync(`${motive}`, 'utf8').split(/\r?\n/);
+    if (cmdAvailable('xcop')) {
+      it(`should find 0 errors in xsl in ${path.basename(motive)}`, function() {
```
maxonfjvipon: @alexachenko let's use `should find no errors...`

test/motives-check.test.js
```diff
@@ -0,0 +1,38 @@
...
+const {runXcop, cmdAvailable} = require('./helpers')
+const fs = require('fs')
+
+/**
+ * Yaml test packs.
+ * @type {Array<String>}
+ */
+const MOTIVES = allFilesFrom(path.resolve(__dirname, '../src/resources/motives'))
+
+describe('motives-check', function() {
```
maxonfjvipon: @alexachenko let'use `xcop motives check`

test/motives-check.test.js
```diff
@@ -0,0 +1,38 @@
...
+    if (cmdAvailable('xcop')) {
+      it(`should find 0 errors in xsl in ${path.basename(motive)}`, function() {
+        for (let i = 0; i < lines.length; i++) {
+          if (lines[i].startsWith('```xsl')) {
+            const xsl = path.resolve(__dirname, 'temp-motives-check.xsl')
+            fs.writeFileSync(xsl, '<?xml version="1.0"?>\n');
+            while (!lines[i+1].startsWith('```')) {
+              fs.appendFileSync(xsl, lines[i+1] + '\n')
+              i++
+            } const stdout = runXcop(xsl)
```
maxonfjvipon: @alexachenko I believe this `const stdout = runXcop(xsl)` should be on the next line

test/xpath-linter.test.js
```diff
@@ -59,7 +59,7 @@ describe('xpath-linter', function() {
     })
     if (cmdAvailable('xcop')) {
       it(`should find 0 errors in xsl in ${path.basename(pack)}`, function() {
```
maxonfjvipon: @alexachenko let's use `..find no errors..`

maxonfjvipon (changes requested):

src/resources/checks/template-match-empty-content-in-instructions.yaml
```diff
@@ -2,5 +2,5 @@
 # SPDX-License-Identifier: MIT
 ---
 xpath: //*[name()='xsl:for-each' or name()='xsl:if' or name()='xsl:otherwise' or name()='xsl:when'][count(node()) = count(text()) and normalize-space() = ""]
-severity: warning
+severity: error
```
maxonfjvipon: @alexachenko I think this change is not related to this task

test/helpers.js
```diff
@@ -26,6 +26,7 @@ const execCmd = function(command, args, print) {
       }
     ).toString()
   } catch (ex) {
+    console.error(ex.stdout.toString())
```
maxonfjvipon: @alexachenko let's add todo here. I think we need to mock `console` here. By default we should use default `console` provided by `node`. In the tests we should use a custom console so we don't see the warnings in the real output, but tests will see it

maxonfjvipon (changes requested):

alexachenko:
@maxonfjvipon I changed packs a bit [(replaced warn with error)](https://github.com/maxonfjvipon/xslint/pull/178/changes#diff-39513b24cb141a8ae30e989e82729c7461eb3a4180f7994a4eeee9b7f2258bc5), assuming that [action](https://github.com/maxonfjvipon/xslint/actions/runs/24799463262/job/72577906987?pr=178#step:6:362) will fail. but this did not happen, despite the fact that the error is visible in the logs. Then is it possible to leave like it is?

test/motives-check.test.js
```diff
@@ -0,0 +1,39 @@
...
+
+describe('xcop motives check', function() {
+  MOTIVES.forEach((motive) => {
+    const lines = fs.readFileSync(`${motive}`, 'utf8').split(/\r?\n/);
+    if (cmdAvailable('xcop')) {
+      it(`should find no errors in xsl in ${path.basename(motive)}`, function() {
+        for (let i = 0; i < lines.length; i++) {
+          if (lines[i].startsWith('```xsl')) {
+            const xsl = path.resolve(__dirname, 'temp-motives-check.xsl')
+            fs.writeFileSync(xsl, '<?xml version="1.0"?>\n');
```
maxonfjvipon: `while (!lines[i+1].startsWith('```'))` has no bounds check. If a ````xsl` block is ever left unterminated (or is the last line of the file with no trailing newline), `lines[i+1]` becomes `undefined` and this throws `Cannot read properties of undefined`. Worth guarding, e.g. `while (i + 1 < lines.length && !lines[i+1].startsWith('```'))`.

Minor: the snippet is assembled into a single shared `temp-motives-check.xsl`. Since the per-motive `it` runs sequentially that's fine, but a unique temp name (or os.tmpdir) would be more robust against parallel runs.

---

test/helpers.js
maxonfjvipon: `ex.stdout` can be `null`/`undefined` (e.g. if the failure isn't an `execSync` non-zero exit but a spawn error), in which case `.toString()` throws and masks the original error. Safer: `console.error(ex.stdout ? ex.stdout.toString() : ex.message)`. Also confirm this extra stderr output doesn't trip the grunt warning check that you mentioned in the PR description.

---

maxonfjvipon (commented):
Good idea — extending xcop coverage to the XSL snippets inside motive docs is worthwhile, and the bulk reformatting (wrapping snippets in a full `xsl:stylesheet`, consistent 2-space indentation, `xmlns:` before `version`) is the right direction. A few things before merge:

**1. Rebase needed — stale base.** This PR targets `master` at `cb8d8a16`, but `master` has since moved motives into subfolders: the files you reformat at `src/resources/motives/*.md` now live under `src/resources/motives/xpath/` (with `corpus/` and `validation/` siblings). This will conflict heavily and/or touch paths that no longer exist. Please rebase onto latest `master`. The good news: `motives-check.test.js` uses `allFilesFrom(...)` which recurses, so it'll still discover the snippets after they move into `xpath/`.

**2. `pr-size` CI check is failing** on this PR (it's large). That may just need a label/exception, but worth resolving.

**3. Robustness nits inline** — unbounded `while` loop in the snippet extractor, shared temp filename, and the `ex.stdout.toString()` in `helpers.js`.

**4. Docs:** per `CLAUDE.md`, behavior changes should keep docs in sync. Adding a new motives xcop test is worth a one-line mention next to the existing `test/xcop.test.js` description.

Solid contribution overall; mainly needs the rebase.

---

maxonfjvipon:
Thanks! Two things before this can merge: (1) `mytext2.txt` looks like a stray file — please remove it; (2) the debug `console.log(ex.stdout.toString())` should go (or be routed through the logger) rather than stay in the code. Once those are cleaned up it's good to go.

maxonfjvipon:
Thanks @alexachenko, and apologies this sat so long.

Closing as unrebasable rather than unwanted. #258 dropped the `template-match-` prefix from every rule name, so all 24 motives this touches have been renamed out from under the branch, and it stands 702 commits behind master.

#170 stays open, because the gap is real — motives still carry XSL that xcop never reads — and it is on the #644 board in the motive and docs batch. Worth doing again against master when that batch comes up.

PR #179, opened 2026-04-27
changed template-match-incorrect-use-of-boolean-constants

### LINKED ISSUE #138: False positive: template-match-incorrect-use-of-boolean-constants on string comparisons
The check `template-match-incorrect-use-of-boolean-constants` fires when comparing XML element values to the strings `'true'` or `'false'`, which is legitimate when the source XML contains boolean-like text content.

Example triggering the warning:
```xml
<xsl:if test="toggles/read-only='true'">
```

This compares the text content of `<read-only>true</read-only>` to the string `'true'`. This is correct XPath - we're not trying to use XSLT boolean constants, we're comparing against string values from XML.

The warning incorrectly assumes any `='true'` or `='false'` is an attempt to use boolean constants.

Workaround: use `string(toggles/read-only) = 'true'` to make the string comparison explicit.

alexachenko:
@maxonfjvipon please review
fixes #138

### DIFF
--- a/src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
+++ b/src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //*[@*[(contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))]]
+xpath: //*[@*[((contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))) and not(contains(., 'string('))]]
 severity: warning
-message: "Boolean constants 'true' and 'false' are non-empty strings, not booleans. Use true() and false() instead."
+message: "Boolean constants 'true' and 'false' are non-empty strings, not booleans. Use true() and false() instead if you want to use boolean. Use string() if you want to use strings."
--- a/src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
+++ b/src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //*[@*[((contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))) and not(matches(., "string\(.*\)\s?=\s?('true'|'false')"))]]
+xpath: //*[@*[((contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))) and not(replace(replace(., 'b', 'd'), "(\s|^)string\(.*\)\s?=\s?('true'|'false')", 'b')=replace(replace(., 'b', 'd'), "(\s|^)\S+\s?=\s?('true'|'false')", 'b'))]]
 severity: warning
 message: "Boolean constants 'true' and 'false' are non-empty strings, not booleans. Use true() and false() instead if you want to use boolean. Use string() if you want to use strings."
--- a/src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
+++ b/src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //*[@*[((contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))) and not(contains(., 'string('))]]
+xpath: //*[@*[((contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))) and not(matches(., "string\(.*\)\s?=\s?('true'|'false')"))]]
 severity: warning
 message: "Boolean constants 'true' and 'false' are non-empty strings, not booleans. Use true() and false() instead if you want to use boolean. Use string() if you want to use strings."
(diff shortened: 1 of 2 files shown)


src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
```diff
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //*[@*[(contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))]]
+xpath: //*[@*[((contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))) and not(contains(., 'string('))]]
```
maxonfjvipon: @alexachenko I'm not sure that this last condition is enough. What about such xpath:
```xpath
string(toggles) = 'hi' and @ready='true'
```

we have `string(` but we still have `'true'` as a string.

maxonfjvipon (changes requested):
@alexachenko check the comment below. please try again to solve the issue using xpath. If no luck - we can try to solve it when we'll be able to parse xpath and build AST for it. I'll make a ticket soon

src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
alexachenko: @maxonfjvipon please review
**UPD:** now I discovered that with `string(toggles) = 'true' and @ready='true'` this xpath does not work.

maxonfjvipon: @alexachenko try this: `string(toggles)='true' and non-string(ready)='false'`

alexachenko: @maxonfjvipon please review 
first we replace 'b' on 'd' so that it doesn't interfere with us further when comparing strings. These letters were chosen because they are not contained in the words 'true', 'false' and 'string'. 
next, we replace the expressions corresponding to the pattern with 'b'. Now all the 'b's' in the strings are our `(\s|^)string\(.*\)\s?=\s?('true'|'false')` (sth like `  @ready='true'`) and`(\s|^)\S+\s?=\s?('true'|'false')` (sth like ` string(toggles)='true'`).
If all `(\s|^)\S+\s?=\s?('true'|'false')` are `(\s|^)string\(.*\)\s?=\s?('true'|'false')` , then the strings are equal to each other

maxonfjvipon: @alexachenko that's dirty business

alexachenko: @maxonfjvipon 
UPD: I find the way with `tokenize()` (it have pushed now), but i don't know, if it can be called clean

alexachenko: @maxonfjvipon you can take a look, please. I will be glad if it can be left:)

test/resources/xpath-packs/template-match-boolean-like-text-content.yaml
```diff
@@ -0,0 +1,39 @@
...
+      <xsl:if test="string(toggles) = 'true' and @ready='true'">
+        <p>not OOO</p>
+      </xsl:if>
+      <xsl:if test="string(toggles)='true' and non-string(ready)='false'">
+        <p>not OOO</p>
+      </xsl:if>
+      <xsl:if test="non-string(bool)='false' and string(toggles)='true'">
+        <p>not OOO</p>
+      </xsl:if>
+    </xsl:template>
```
maxonfjvipon, line 34: @alexachenko let's also test such approach:
```xsl
  <xsl:if test="string(a) = 'true' and string(b) = 'false'">
    <p>not OOO</p>
  </xsl:if>
```

It seems it will be false positive also here

src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml
```diff
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //*[@*[(contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))]]
+xpath: //*[@*[((contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))) and not(count(tokenize(., "(\s|^)string\(.*\)\s?=\s?('true'|'false')"))=count(tokenize(., "(\s|^)\S+\s?=\s?('true'|'false')")))]]
```
maxonfjvipon, line 4: This XPath is very hard to follow and looks fragile:

- The `count(tokenize(., "...string\(.*\)...")) = count(tokenize(., "...\S+..."))` trick infers "every `= 'true'/'false'` comparison is a `string(...)=...` comparison" by comparing split-counts. It's clever but obscure — anyone maintaining this will struggle.
- `string\(.*\)` uses a **greedy** `.*`, so in an attribute with several comparisons it spans across them (e.g. `string(toggles)='true' and non-string(ready)='false'`), which makes the match-count reasoning brittle. It happens to land on the expected 5/0/2 in the test fixtures, but I'd be cautious about real-world expressions (multiple comparisons, nested parens, `'true'`/`'false'` substrings inside literals, whitespace variants).

Two requests:
1. Add a brief comment in the YAML (or the motive) explaining the intent, since the expression is non-obvious.
2. Consider whether a simpler formulation is possible. If the tokenize approach stays, please add fixtures for trickier cases (e.g. a literal containing the word `true`, and `count(... = 'true')` style function calls) so the brittleness is pinned down.

---

maxonfjvipon (commented):
The intent here is good — not flagging legitimate `string(...) = 'true'` string comparisons while still catching bare `foo='true'` — and the tests (5/0/2) pass on this branch. Two things before merge:

**1. Rebase needed — this PR is on a stale base.** It targets `master` at `cb8d8a16`, but `master` has since reorganized checks and motives into subfolders. The file you edit here, `src/resources/checks/template-match-incorrect-use-of-boolean-constants.yaml`, now lives at `src/resources/checks/xpath/template-match-incorrect-use-of-boolean-constants.yaml` on current `master` (likewise the new test pack and the motive). As-is this won't apply cleanly / will recreate the file in the old location. Please rebase onto latest `master` and move the edits into `checks/xpath/`.

**2. The XPath is fragile and unreadable** — see inline comment. At minimum add an explanatory comment; ideally simplify and add fixtures for the tricky cases.

The message improvement ("Use string() if you want to use strings") is a nice touch and reads well.

---

maxonfjvipon:
@rultor merge

PR #184, opened 2026-05-20
added check text-outside-xsl-text

### LINKED ISSUE #141: text without `<xsl:text>` is not caught
This is my code:

```
<xsl:if test=".!=''">
  Hello!
</xsl:if>
```

It must be written as such, instead:

```
<xsl:if test=".!=''">
  <xsl:text>Hello!</xsl:text>
</xsl:if>
```

Any text (except spaces) outside of `<xsl:text>` must be prohibited.

alexachenko:
@maxonfjvipon please review
fixes #141 

I also left here an ability to write text in html tags and when declaring variables and params. Is it correct?


src/resources/checks/template-match-text-outside-node.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //*[count(text()[normalize-space()])>0 and (not(name() = 'xsl:text') and (not((name() = 'xsl:variable' or name() = 'xsl:param' or name() = 'xsl:with-param') and not(@select)))) and not(name() = ('div', 'p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'section', 'article', 'header', 'footer', 'main', 'aside', 'nav', 'blockquote', 'pre', 'figure', 'figcaption', 'details', 'form', 'fieldset', 'legend', 'span', 'a', 'strong', 'em', 'b', 'i', 'u', 's', 'small', 'mark', 'code', 'kbd', 'samp', 'var', 'sub', 'sup', 'abbr', 'time', 'data', 'q', 'cite', 'dfn', 'bdi', 'bdo', 'ruby', 'rt', 'rp', 'wbr', 'li', 'dt', 'dd', 'td', 'th', 'caption', 'colgroup', 'video', 'audio', 'iframe', 'object', 'canvas', 'svg', 'math'))]
```
maxonfjvipon: The host-language whitelist is the wrong shape for this rule. Two problems:

1. The list is incomplete — `title`, `option`, `label`, `button`, `summary`, `textarea`, `output`, `meter`, `progress`, `address`, `del`, `ins`, table elements beyond `td`/`th`, plus the entire SVG/MathML/custom-vocabulary world are all absent.
2. `name() = (...)` compares the *qualified* name. `<html:p>` or any prefixed literal result element won't match the unprefixed names in the list.

The rule's job is to police XSL, not to know what HTML is. Suggest flipping the predicate so it only inspects elements in the XSL namespace, e.g.:

```xpath
//xsl:*[
  count(text()[normalize-space()]) > 0
  and not(local-name() = ('text','attribute','comment','processing-instruction','message'))
  and not(local-name() = ('variable','param','with-param') and not(@select))
]
```

This removes the HTML list entirely, fixes the prefix problem, and stops flagging literal result elements of any vocabulary.

maxonfjvipon: The variable/param exemption is the right idea, but several other XSL elements legitimately carry text content and are not exempted: `xsl:attribute`, `xsl:comment`, `xsl:processing-instruction`, `xsl:message`. Of these, `xsl:attribute` matters most — `<xsl:attribute name="class">container</xsl:attribute>` is idiomatic XSLT and is currently flagged (see the updated `template-match-not-creating-element-correctly` pack).

maxonfjvipon: Naming nit: the rule file is `template-match-text-outside-node`, but the branch and motive talk about text outside `xsl:text`. "Outside the node" is unclear. Suggest renaming to `template-match-text-outside-xsl-text.yaml` (and the motive/test packs accordingly) so the filename matches the actual check.

test/resources/xpath-packs/template-match-not-creating-element-correctly.yaml
```diff
@@ -3,8 +3,8 @@
 ---
 pack: template-match-not-creating-element-correctly
 found:
-  amount: 1
-  positions: [ [ 5, 5 ] ]
+  amount: 2
+  positions: [ [ 5, 5 ], [ 6, 7, 'template-match-text-outside-node' ] ]
```
maxonfjvipon: This new expected defect at `[6, 7, 'template-match-text-outside-node']` points at `<xsl:attribute name="class">container</xsl:attribute>` on line 6 of the fixture. That's idiomatic XSLT for computing an attribute value and should not be a defect — the rule needs to exempt `xsl:attribute` (see the rule-file comment).

Also, piggy-backing one rule's coverage onto another rule's fixture muddies the negative case for `template-match-not-creating-element-correctly`. Once the rule is fixed to not fire here, please revert this fixture and keep the new rule's coverage in its own packs.

test/resources/xpath-packs/template-match-text-in-node.yaml
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-text-outside-node
```
maxonfjvipon: Filename doesn't follow the project's pack naming. Existing negative-case packs are named after the rule with a `not-` infix (e.g. `template-match-blank-nested-if` ↔ `template-match-not-blank-nested-if`). Suggest renaming this file to `template-match-text-inside-xsl-text.yaml` (or `template-match-no-text-outside-xsl-text.yaml`) to match the convention.

src/resources/motives/template-match-text-outside-node.md
```diff
@@ -0,0 +1,20 @@
+# Text is outside the node
```
maxonfjvipon: Title "Text is outside the node" is ambiguous (text is always inside some node). Suggest `Literal text outside xsl:text` — and please also document the exceptions (HTML / literal result elements of any vocabulary, `xsl:variable` / `xsl:param` / `xsl:with-param` without `@select`, `xsl:attribute`, etc.) so users understand the boundary.

test/resources/reports/xsl-with-some-violations.xsl
```diff
@@ -29,15 +29,24 @@
     </p>
   </xsl:template>
   <xsl:template match="//title">
-    Title: <span style="color:#ff0000"><xsl:value-of select="."/></span>
+    <xsl:text>Title:</xsl:text>
+    <span style="color:#ff0000">
```
maxonfjvipon, line 33: The original was `Title: <span>...</span>` (with a trailing space after the colon). The rewrite becomes `<xsl:text>Title:</xsl:text>` — the space is gone. That's a semantic change in the fixture's expected output. If you want to preserve it, use `<xsl:text>Title: </xsl:text>`. Same applies to the corresponding lines in `xsl-packs-3/xsl-with-some-violations.xsl` and `xsl-packs/xsl-with-some-violations.xsl`.

maxonfjvipon (commented):
**On your question about HTML.** I'd argue this project should not have any HTML-specific knowledge. xslint lints XSL, but the *output* of an XSL transform can be HTML, SVG, MathML, XSL-FO, custom XML, or anything else. Hardcoding an HTML element list:
- is inherently incomplete (the current list misses `title`, `option`, `label`, `button`, `summary`, table elements beyond `td`/`th`, etc., and any custom/future element);
- punishes anyone outputting non-HTML (SVG `<text>`, MathML, custom vocabularies would all be flagged);
- conflates "literal result element" (an XSL concept — any non-XSL element in a template body) with "HTML element" (a particular vocabulary).

The cleaner model: **the rule should only inspect elements in the XSL namespace**. Literal result elements of any vocabulary are simply out of scope — XSL doesn't care what vocabulary you're outputting. Inside XSL, exempt the elements that legitimately carry text content: `xsl:text`, `xsl:variable`/`xsl:param`/`xsl:with-param` without `@select`, `xsl:attribute`, `xsl:comment`, `xsl:processing-instruction`, `xsl:message`. That shrinks the XPath to something readable and removes a whole class of false positives.

**On scope.** As currently written, the rule also fires on `<xsl:attribute name="x">value</xsl:attribute>` (see the updated `template-match-not-creating-element-correctly` pack — the new defect at row 6 col 7 lands on `<xsl:attribute>`). That's the idiomatic way to compute attribute values in XSLT 1.0; flagging it as a defect is incorrect.

**Missing test coverage.** Please add packs for:
- text inside `<xsl:attribute>` — should NOT fire;
- text inside `<xsl:variable name="x">value</xsl:variable>` (no `@select`) — should NOT fire;
- text inside `<xsl:comment>`/`<xsl:processing-instruction>`/`<xsl:message>` — should NOT fire;
- text inside a non-HTML literal result element (e.g. `<svg:text>foo</svg:text>` or a custom namespace element) — should NOT fire;
- text in an HTML element currently missing from the whitelist (e.g. `<title>`, `<option>`) — under the suggested redesign should NOT fire.

alexachenko:
@maxonfjvipon [here](https://github.com/maxonfjvipon/xslint/pull/184#pullrequestreview-4328592090) you wrote this:
> text inside <xsl:variable name="x">value</xsl:variable> (no @select) — should NOT fire;

but later we add new [issue](https://github.com/maxonfjvipon/xslint/issues/193)
so, can i ignore it, because it will be a part of another check?

**UPD:** while i added some changes with ignoring this part of check

src/resources/checks/template-match-text-outside-node.yaml
alexachenko: see [here](https://github.com/maxonfjvipon/xslint/pull/184#issuecomment-4757603294)

maxonfjvipon:
@alexachenko in general yes, each check should be isolated and validate only its part and don't complain about other similar problems

test/resources/xpath-packs/template-match-text-inside-non-XSL-element.yaml
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-text-outside-xsl-text
+found:
+  amount: 0
+  positions: [ ]
+input: |-
+  <?xml version="1.0"?>
+  <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:ex=https://example.com" xmlns:xs="http://www.w3.org/2001/XMLSchema" version="2.0" id="style1">
```
maxonfjvipon: 🐛 This line is malformed XML — `xmlns:ex=https://example.com"` is missing the opening quote (should be `xmlns:ex="https://example.com"`). I ran this pack locally and xmldom emits:

```
[xmldom warning] attribute "xmlns:ex" missed start quot(")!!
[xmldom error]  element parse error: Error: invalid tagName:<
```

The test still reports `amount: 0` and "passes", but only because the document fails to parse, so the linter never sees the `<ex:text>foo</ex:text>` element you intended to test. In other words this fixture is currently passing vacuously and is **not** verifying that text inside a non-XSL (custom-namespace) element is ignored. It slips through CI because the inline pack XML isn't run through xcop in CI (only committed `.xsl` files are). Please fix the quote so the fixture actually exercises the rule.

---

src/resources/checks/xpath/template-match-text-outside-xsl-text.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:*[count(text()[normalize-space()]) > 0 and not(local-name() = ('text','attribute','comment','processing-instruction','message','variable','param','with-param'))]
+severity: warning
+message: Text element is outside the node. Write it in 'xsl:text'.
```
maxonfjvipon: Nit: the message reads a little awkwardly ("Text element is outside the node"). Consider something clearer, e.g.:

```yaml
message: Literal text appears directly inside an xsl:* instruction. Wrap it in xsl:text.
```

Also a scope note: this rule excludes `variable`/`param`/`with-param` unconditionally, whereas text in those is only valid when there's no `@select` (a `select` + text body is itself an error). Fine to leave for now, just flagging the gap.

---

maxonfjvipon (commented):
Took a look — the rule itself is sound and the approach (flag `xsl:*` instructions that carry non-whitespace text, excluding the elements where text is the value) is reasonable, and I confirmed the 12 text-* pack assertions pass locally. The reformatting of the report/stylesheet/template fixtures into `xsl:text` is consistent.

One real issue and a couple of nits inline:
- **`template-match-text-inside-non-XSL-element.yaml` has malformed XML** (missing quote), so that fixture passes vacuously and doesn't actually test the custom-namespace case — please fix (details inline).
- Minor: message wording, and the `variable`/`param`/`with-param` exclusion is unconditional.
- Trivia: the new motive `.md` is missing a trailing newline.

Once the malformed fixture is fixed (and ideally re-checked that it still reports 0 for the right reason), this looks good to merge.

---

alexachenko:
@maxonfjvipon please review

src/resources/checks/xpath/template-match-text-outside-xsl-text.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:*[count(text()[normalize-space()]) > 0 and not(local-name() = ('text','attribute','comment','processing-instruction','message','variable','param','with-param'))]
```
maxonfjvipon, line 4: The `variable`/`param`/`with-param` exemption works against the rule's own rationale. Loose text in `xsl:if`/`xsl:template` (which you do flag) only adds stray output whitespace, but loose text in a variable body silently becomes part of the *value*:

```xsl
<xsl:variable name="qw">
  ooooo
</xsl:variable>
```

`$qw` is not `"ooooo"` — XSLT strips only whitespace-*only* text nodes from a stylesheet, and `"\n  ooooo\n"` is not whitespace-only, so the indentation leaks into the value. That makes this the case the rule arguably most needs to catch. Suggest either flagging these too, or exempting them only when the body is a single `xsl:text`. The `comment`/`processing-instruction`/`message` exemptions are fine — text there is the payload.

src/resources/checks/xpath/template-match-text-outside-xsl-text.yaml
```diff
@@ -0,0 +1,6 @@
...
+severity: warning
+message: Literal text appears directly inside an xsl:* instruction. Wrap it in xsl:text.
```
maxonfjvipon: `inside an xsl:* instruction` leaks the selector syntax into a user-facing message; the other rules name concrete elements (e.g. "such as xsl:for-each, xsl:if, or xsl:when"). Suggest: `Literal text appears directly inside an xsl: instruction element. Wrap it in xsl:text.`

src/resources/motives/xpath/template-match-text-outside-xsl-text.md
```diff
@@ -0,0 +1,82 @@
+# Literal text outside `xsl:text`
+
+Text element can't be outside the node. Any text (except spaces) must be written
```
maxonfjvipon: Three small things on this file: the opening "Text element can't be outside the node." reads awkwardly, there is a trailing space at the end of this line, and the file has no trailing newline (every other motive ends with one). Maybe open with: "Literal text (other than whitespace) must not sit directly inside an `xsl:` instruction — wrap it in `xsl:text`."

test/resources/templates/xsl-with-no-violations.xsl
```diff
@@ -49,5 +49,14 @@
       <xsl:with-param name="color" select="'#00ffff'"/>
     </xsl:call-template>
   </xsl:template>
-  <xsl:template name="string"><xsl:param name="name" select="'Unknown'"/><xsl:param name="color" select="'#000000'"/><xsl:value-of select="$name"/>: <span style="color:{$color}"><xsl:value-of select="."/></span><br/></xsl:template>
+  <xsl:template>
```
maxonfjvipon: This dropped `name="string"`, so `<xsl:template>` now has neither `match` nor `name` (invalid XSLT), and the three `<xsl:call-template name="string">` above (lines 35, 41, 47) no longer resolve. The `reports/` and `stylesheets/` copies kept `name="string"`, and all three were identical before this PR — looks like a slip during reformatting. The `Processed files: N` assertions do not catch it. Please restore `name="string"`.

test/resources/xpath-packs/template-match-text-outside-xsl-text.yaml
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-text-outside-xsl-text
+found:
+  amount: 1
```
maxonfjvipon: The only violating case in the suite is `xsl:if`. Could you add a positive case for loose text directly in `xsl:template` — that is the exact #141 scenario and what the shared fixtures were reformatted around — and ideally one more container such as `xsl:for-each`? Right now the headline case is not asserted directly.

test/resources/xpath-packs/template-match-text-inside-html-element.yaml
```diff
@@ -0,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-text-outside-xsl-text
```
maxonfjvipon, line 4: This file and `template-match-text-inside-non-XSL-element.yaml` use uppercase in their names; every other pack is lowercase-kebab. Suggest renaming to `...-html-element.yaml` and `...-non-xsl-element.yaml` for consistency.

src/resources/checks/xpath/template-match-text-outside-xsl-text.yaml
alexachenko: @maxonfjvipon can we put this in a separate check? User can write it in one line, fully understanding what he is doing. for example:
```xsl
<xsl:variable name="para-font-size" as="xs:string">12pt</xsl variable>
```
we could put this in a warning explaining to the user that the `variable/param/with-param` he has entered will be saved with all line breaks and other possibly unwanted characters.

alexachenko: also if we change this rule this way we should change all rules with `variable/param/with-param` where like examples we use smth like this:
```xsl
<xsl:variable name="qw">
  ooooo
</xsl:variable>
```

maxonfjvipon: @alexachenko agreed, let's make it a separate check. Keep this rule focused on loose text inside `xsl:if`/`xsl:template`/`xsl:for-each` etc., and add a new warning-level check for text content directly inside `xsl:variable`/`xsl:param`/`xsl:with-param`, with a message explaining that the value will keep all line breaks and surrounding whitespace. That way this PR doesn't need to touch the examples in the other variable/param rules — those stay valid as-is.

---

maxonfjvipon:
@alexachenko thanks for the work here. Here's the to-do list to get this PR ready for a final review:

- [x] **Split out the variable/param case into a new check** (as agreed above): keep `template-match-text-outside-xsl-text` focused on loose text inside `xsl:if`/`xsl:template`/`xsl:for-each` etc., and add a new **warning**-level check for text content directly inside `xsl:variable`/`xsl:param`/`xsl:with-param`, with a message noting the value keeps all line breaks and surrounding whitespace. New YAML + motive + test packs.
- [x] **Fix the dropped `name="string"`** in `test/resources/templates/xsl-with-no-violations.xsl` — the template currently has neither `match` nor `name`, and the three `xsl:call-template name="string"` references above it no longer resolve. Restore it (the `reports/` and `stylesheets/` copies still have it).
- [x] **Add the headline positive cases** — assert loose text directly inside `xsl:template` (the exact #141 scenario) and ideally one more container such as `xsl:for-each`. Right now only `xsl:if` is covered.
- [x] **Message wording** — avoid the `xsl:*` selector syntax in the user-facing message; name concrete elements like the other rules do.
- [x] **Motive polish** — reword the awkward opening sentence, drop the trailing space, and add a trailing newline (every other motive ends with one).
- [x] **Pack filenames** — rename `template-match-text-inside-HTML-element.yaml` and `template-match-text-inside-non-XSL-element.yaml` to lowercase-kebab (`...-html-element.yaml`, `...-non-xsl-element.yaml`) to match the convention.

Once these are in, ping me for a re-review.

---

alexachenko:
@maxonfjvipon please review

src/resources/checks/xpath/template-match-text-outside-xsl-text.yaml
alexachenko: open [here](https://github.com/maxonfjvipon/xslint/issues/230)

src/resources/motives/xpath/template-match-text-outside-xsl-text.md
```diff
@@ -0,0 +1,90 @@
...
+<ul>
+  <li>William Shakespeare</li>
+  <li>Robert Burns</li>
+  <li>Lord Byron</li>
+</ul>
+```
+
+or:
+```xsl
+<xsl:msg>
```
maxonfjvipon: This "Correct" example is actually flagged by the rule, so it contradicts itself. `<xsl:msg>` is in the XSLT namespace (the `xsl:` prefix), and `msg` isn't in the exemption list, so `//xsl:*[count(text()[normalize-space()]) > 0 and not(local-name() = (…))]` matches it. I verified locally — it returns 1 defect, not 0.

You probably meant to illustrate a *non-XSL* element. Either drop this block (the `<ul><li>…` example above already covers literal result elements) or replace it with a genuinely non-XSL element, e.g.:

```xsl
<my:note xmlns:my="https://example.com">
  some text
</my:note>
```

Also on line 7: "in another non-xsl elements" → "in other non-XSL elements".

---

maxonfjvipon (commented):
One thing to fix before merge (inline): the motive's `<xsl:msg>…</xsl:msg>` block is listed as **Correct** but the rule actually flags it (I verified — 1 defect). Swap it for a real non-XSL element or drop it.

On the separate variable/param text check we agreed on: I see this PR keeps `variable`/`param`/`with-param` exempted and leaves that as follow-up work — that's fine by me. Could you open a tracking issue for the new warning-level check so it isn't lost? Then once the `xsl:msg` example is fixed, this is good to merge.

src/resources/motives/xpath/template-match-text-outside-xsl-text.md
alexachenko: @maxonfjvipon 
issue was opened #230
fixed `template-match-text-outside-xsl-text.md`

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

PR #202, opened 2026-05-30
Added required attribute test

### LINKED ISSUE #198: `xsl:if` and `xsl:when` should have `test` attribute
[here](https://www.w3.org/TR/xslt20) i read
> The [xsl:if](https://www.w3.org/TR/xslt20/#element-if) element has a mandatory test attribute, which specifies an [expression](https://www.w3.org/TR/xslt20/#dt-expression).

>Each [xsl:when](https://www.w3.org/TR/xslt20/#element-when) element has a single attribute, test, which specifies an [expression](https://www.w3.org/TR/xslt20/#dt-expression).

so, this is incorrect:
```xsl
<xsl:when>
  <!-- body -->
</xsl:when>
```

```xsl
<xsl:if>
  <!-- body -->
</xsl:if>
```

i think, we should catch it with check with severity "error"

alexachenko:
@maxonfjvipon please review
fixed #198


maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

PR #207, opened 2026-06-18
bug(#106): cross-file corpus linter for unused named templates

### LINKED ISSUE #106: Should implement new check for Using NamedTemplate in another files
There is information [here](https://stackoverflow.com/questions/42317523/how-do-i-import-templates-from-another-stylesheet) about import templates in another stylesheets and another files with `xsl:import`.
The xpath (`//xsl:template[not(@match)][@name][not(//xsl:call-template/@name = @name)]`) from this [file](https://github.com/maxonfjvipon/xslint/blob/master/src/resources/template-match-unused-named-template.yaml) in PR #97 (fixes #93) can't check another files for using the created named templates. 
It is necessary to implement checks for the case when the named template does not use in this stylesheet but uses in another file.

maxonfjvipon:
Closes #106.

## Problem

`template-match-unused-named-template` ran file by file, so a named template
defined in one stylesheet but invoked from another (via `xsl:import` /
`xsl:include`) was wrongly flagged as unused. XPath cannot see across files,
so this needs a pass over the whole set of linted stylesheets.

## Approach

A **corpus** — every linted stylesheet parsed once and held together — is now
handed to each linter. The unified contract is `(corpus, suppressions) => defects`,
where `corpus` is `[{file, xsl}]` and every defect is tagged with its `file`.

- **`src/corpus-linter.js`** — data-driven cross-file checks. A declaration is
  a defect only when its name is used by no stylesheet in the corpus, so a
  template called from another file is not reported.
- **`src/xpath.js`** — extracted shared fontoxpath environment (prefixes,
  custom functions, node/string evaluators). The two linters now depend only
  on it, not on each other.
- **Suppression validation** moved to the orchestrator, validating against the
  names every linter exposes (so corpus checks are suppressible too).
- **Directory layout**: `checks/{xpath,corpus}` and `motives/{xpath,corpus}`;
  the doc generator walks both kinds and renders corpus checks'
  `declaration`/`usage`.

## Semantics

Global union: a template is unused only if no file in the lint set calls it.
This resolves the false positive for the realistic case (lint the project
together). It does not parse the `import`/`include` graph — a precision
upgrade left for a follow-up. Lint the whole project so the check sees callers.

## Tests

- New `test/corpus-linter.test.js` with multi-file packs: cross-file call → 0,
  same-file call → 0, defined-but-uncalled across files → 1.
- xcop format check moved into its own `test/xcop.test.js`, now covering both
  xpath and corpus packs (every inline stylesheet).
- Full suite green; ESLint clean.


maxonfjvipon:
@rultor merge

Issue #134, opened 2026-03-31
XPath formatting is not checked


yegor256:
This is my XSL:

```
<xsl:if test="foo(  a) ==         'hello'   ">
  ...
</xsl:if>
```

It's wrong, but xslint doesn't complain.

maxonfjvipon:
## Epic: an XPath formatting linter, built on a real lexer → parser

PR #212 closed the *validity* half of this ticket: XPath that **cannot parse** is
now reported. This epic covers the other half — XPath that parses but is **badly
formatted** — and, as discussed, with *specific* messages ("redundant
whitespace", "redundant parentheses") rather than a binary "yours ≠ perfect".

### Why a real parser — and what "real" has to mean

A textbook parser produces an *abstract* syntax tree, which by design discards
exactly what we want to lint. In fontoxpath `foo(  a)` and `foo(a)` compile to
byte-identical output, and `((a))` ≡ `(a)` ≡ `a`. So to check whitespace **and**
redundant parens we need a **full-fidelity (lossless) parser**: a positioned
token stream plus a tree that keeps every parenthesis and source span. The lexer
is *layer 1* of that parser, not an alternative to it.

### The guard rail that lets us ship in layers

fontoxpath stays the **validity oracle**. The formatting linter is **style-only
with bail-out**: it lints only expressions that already parse, and (once a
grammar exists) skips any construct it does not yet understand. A partial grammar
therefore yields *fewer* lints, never *wrong* ones — so each layer is safe to
release on its own.

### Layers

1. **Lexer + whitespace lints** — token-level, precise positions.
   `redundant-whitespace` (doubled space; leading/trailing space). *This PR.*
2. **Adjacency lints** (lexer only) — space inside `()`/`[]`, space around `,`,
   space after `@`/`$`, space around `::`, spaced path `/`; quote-style
   consistency.
3. **Operator spacing** — `a=b` → `a = b`. Needs XPath's lexical disambiguation
   (`*` multiply vs wildcard, `div`/`and`/`or` operator vs name, unary vs binary
   `-`). Higher false-positive risk, so gated behind layers 1–2.
4. **Grammar + structural lints** — recursive-descent over the token stream into
   a lossless tree: redundant parentheses, axis abbreviations
   (`child::x`→`x`, `self::node()`→`.`, `parent::node()`→`..`,
   `attribute::x`→`@x`), `/descendant-or-self::node()/`→`//`, needless
   `boolean(...)`.

### Deferred puzzles

[...]

PR #212, opened 2026-06-19
bug(#134): validate XPath expressions and XSL well-formedness

maxonfjvipon:
Part of #134. Implements the **validity** half of the ticket — checking that XPath expressions are syntactically valid (and that stylesheets are well-formed XML). The **formatting** half (extra whitespace, redundant braces) remains a follow-up.

## What changed

xslint now runs in two stages: **validators** establish that the input is valid, then **linters** run over the stylesheets that pass.

- **`xsl-validator`** — builds the corpus from raw sources and reports each stylesheet that is not well-formed XML (`malformed-stylesheet`), leaving it out of linting. This also fixes a latent crash: a malformed `.xsl` previously aborted the whole run.
- **`xpath-validator`** — parses every bare-XPath-expression attribute (`select`, `test`, `use`, `value`, `group-by`, `group-adjacent`, and the XSLT 3.0 `key`, `initial-value`, `xpath`, `context-item`, `with-params`, `namespace-context`) with fontoxpath — the same engine that runs the rules — and reports the ones that do not compile (`invalid-xpath-expression`). A permissive resolver isolates genuine syntax errors from unknown prefixes and custom functions.
- **`xslint.js`** orchestrates validators then linters, merging defects.

Gating: a file that does not parse is reported once and skipped; the rest are still linted, so one broken file never hides feedback on the others.

## Deliberately out of scope (follow-ups)

- The **formatting** linter (whitespace, redundant parens) — the other half of #134.
- **Pattern** attributes (`match`, `count`, `from`, `group-starting-with`, `group-ending-with`) — a different grammar.
- **AVT** (`{…}`) expression validation.
- Per-expression lint gating (skip invalid expressions during linting) — no expression-level linter consumes it yet.

## Tests

199 passing. New: `xpath-validator` packs (valid/invalid across all attributes), `xsl-validator` unit tests (corpus partitioning + suppression), and an end-to-end gating test over a temp directory. Docs site regenerated (38 checks); README and CLAUDE.md updated, plus a "Keeping Docs in Sync" rule.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,19 +21,26 @@ npx mocha test/xslint.test.js --grep "sentence"  # Run tests matching a pattern
 
 ## Architecture
 
-**xslint** is a CLI linter for XSL stylesheets. It finds `.xsl` files, parses them all into a *corpus* of XML documents, runs each linter over the whole corpus, and reports defects.
+**xslint** is a CLI linter for XSL stylesheets. It runs in two stages: **validators** first establish that the input is valid (each stylesheet is well-formed XML, and every XPath expression compiles), then **linters** run over the stylesheets that pass, catching stylistic, semantic, and logical problems. A stylesheet that does not parse is reported once and left out of the corpus, so one broken file never hides the feedback on the rest.
 
-Linting flow:
+Flow:
 ```text
 src/index.js (CLI, commander.js)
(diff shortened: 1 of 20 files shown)


maxonfjvipon:
@rultor merge

PR #213, opened 2026-06-20
added check template-match-variable-and-param-without-name

### LINKED ISSUE #195: `xsl:variable` and `xsl:param` should have attribute name:
[here](https://www.w3.org/TR/xslt20/) i read:
> The [xsl:variable](https://www.w3.org/TR/xslt20/#element-variable) element has a required name attribute, which specifies the name of the variable.
>The [xsl:param](https://www.w3.org/TR/xslt20/#element-param) element has a required name attribute, which specifies the name of the parameter
we should write:
```
<xsl:variable name="qw">
    101
</xsl:variable>
```
we can't write:
```
<xsl:variable >
    101
</xsl:variable>
```
i think, we should catch it like error

alexachenko:
@maxonfjvipon please review
fixes #195

### DIFF
--- a/src/resources/checks/xpath/template-match-unused-variable.yaml
+++ b/src/resources/checks/xpath/template-match-unused-variable.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //xsl:variable[every $x in (//xsl:variable/../*//@*) satisfies not(contains($x, concat('$', @name)))]
+xpath: //xsl:variable[@name][every $x in (//xsl:variable/../*//@*) satisfies not(contains($x, concat('$', @name)))]
 severity: warning
 message: A variable is declared but never referenced by $name in the stylesheet. Remove it or use it.
--- a/src/resources/checks/xpath/template-match-variable-or-param-without-name.yaml
+++ b/src/resources/checks/xpath/template-match-variable-or-param-without-name.yaml
@@ -3,4 +3,4 @@
 ---
 xpath: //(xsl:variable | xsl:param)[not(@name)]
 severity: error
-message: xsl:variable or xsl:param haven't @name attribute. Add @name attribute or delete node.
+message: xsl:variable and xsl:param require a @name attribute. Add one or remove the element.
--- a/src/resources/checks/xpath/template-match-variable-or-param-without-name.yaml
+++ b/src/resources/checks/xpath/template-match-variable-or-param-without-name.yaml
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //(xsl:variable | xsl:param)[not(@name)]
+severity: error
+message: xsl:variable or xsl:param haven't @name attribute. Add @name attribute or delete node.
--- a/test/resources/xpath-packs/template-match-variable-and-param-without-name.yaml
+++ b/test/resources/xpath-packs/template-match-variable-and-param-without-name.yaml
@@ -1,0 +1,25 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-variable-or-param-without-name
+found:
+  amount: 2
+  positions: [ [ 5, 5 ], [ 9, 5 ] ]
+input: |-
+  <?xml version="1.0"?>
+  <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" version="2.0" id="style1">
+    <xsl:output encoding="UTF-8" method="xml"/>
+    <xsl:template name="ooo">
+      <xsl:param as="xs:string" select="ooo"/>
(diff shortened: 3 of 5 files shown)


src/resources/checks/xpath/template-match-variable-or-param-without-name.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //(xsl:variable | xsl:param)[not(@name)]
+severity: error
+message: xsl:variable or xsl:param haven't @name attribute. Add @name attribute or delete node.
```
maxonfjvipon: Nit: `haven't @name attribute` is ungrammatical, and the other rule messages read as polished full sentences. Suggest something like:

```yaml
message: xsl:variable and xsl:param require a @name attribute. Add one or remove the element.
```

---

src/resources/checks/xpath/template-match-unused-variable.yaml
```diff
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //xsl:variable[every $x in (//xsl:variable/../*//@*) satisfies not(contains($x, concat('$', @name)))]
+xpath: //xsl:variable[@name][every $x in (//xsl:variable/../*//@*) satisfies not(contains($x, concat('$', @name)))]
```
maxonfjvipon, line 4: 👍 Good defensive guard, and it's not incidental to this PR. Without `[@name]`, an unnamed variable would reach `concat('$', @name)` → `concat('$', ())` → `'$'`, so `contains($x, '$')` could mis-fire this "unused variable" warning on a node that is really a *different* defect (the new without-name rule). The negative test's "2 defects by all checks" assertion is what proves the two rules no longer overlap. Nicely done.

---

test/resources/xpath-packs/template-match-variable-and-param-without-name.yaml
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-variable-or-param-without-name
```
maxonfjvipon, line 4: Nit: filename uses `...-variable-and-param-...` but the check and the `pack:` value below are `...-variable-or-param-...` ("and" vs "or"). Harmless (packs are auto-discovered by directory and keyed by the `pack:` field), but consider renaming both fixtures to match the check name for consistency.

---

test/resources/xpath-packs/template-match-variable-and-param-without-name.yaml
```diff
@@ -0,0 +1,25 @@
...
+    <xsl:template name="ooo">
+      <xsl:param as="xs:string" select="ooo"/>
+      <p>$o1</p>
+    </xsl:template>
+    <xsl:template match="/objects/o/o[1]/o[2]">
+      <xsl:variable as="xs:string" select="ooo"/>
+      <p>ooo</p>
+    </xsl:template>
+    <xsl:template match="/objects/o/o[1]/o[3]">
+      <xsl:call-template name="ooo">
```
maxonfjvipon, line 21: Scope note (non-blocking): this `xsl:with-param` has no `@name`, which is also invalid XSLT, but it's intentionally not flagged since this rule targets only `xsl:variable`/`xsl:param`. That's a reasonable boundary for issue #195 — just flagging it as a natural follow-up (along with other required-`name` elements like `xsl:attribute`, `xsl:key`). Worth a separate issue rather than expanding this PR.

---

maxonfjvipon (commented):
Reviewed — looks good overall. Correctly implements #195 (both `xsl:variable` and `xsl:param` require `@name`, reported as `error`), all 16 CI checks are green, and I confirmed the full xpath-linter suite (172 tests) passes locally including the two new packs.

Highlights:
- The `[@name]` guard added to `template-match-unused-variable` is the right fix and cleanly separates the two rules (see inline comment).
- No doc wiring needed — `generate-docs.js` auto-scans `motives/xpath`, and the new fixtures land in the already-registered `xpath-packs/` dir so xcop covers them.

Only minor, non-blocking nits inline: message grammar, fixture filename/`pack` naming consistency, and a scope note on `xsl:with-param`. None block merge.

---

test/resources/xpath-packs/template-match-variable-and-param-without-name.yaml
alexachenko: @maxonfjvipon i think we need use `...-variable-and-param-...` in the test-pack name because we test both `xsl:variable` and `xsl:param` in it. At the same time, check-name use `...-variable-or-param-...` because it will catch `xsl:variable` or `xsl:param` separately.

src/resources/checks/xpath/template-match-variable-or-param-without-name.yaml
alexachenko: @maxonfjvipon pushed

test/resources/xpath-packs/template-match-variable-and-param-without-name.yaml
alexachenko: open #214

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

PR #215, opened 2026-06-21
added rule with-param-use-in-invalid-parent-node

### LINKED ISSUE #199: `xsl:with-param` is allowed within xsl:call-template, xsl:apply-templates, xsl:apply-imports, and xsl:next-match.
[here](https://www.w3.org/TR/xslt20/#element-with-param) i read:
> [xsl:with-param](https://www.w3.org/TR/xslt20/#element-with-param) is allowed within [xsl:call-template](https://www.w3.org/TR/xslt20/#element-call-template), [xsl:apply-templates](https://www.w3.org/TR/xslt20/#element-apply-templates), [xsl:apply-imports](https://www.w3.org/TR/xslt20/#element-apply-imports), and [xsl:next-match](https://www.w3.org/TR/xslt20/#element-next-match).

so we can't use `xsl:with-param` outside of these elements
it should be check with severity error

alexachenko:
@maxonfjvipon please review
fixes #199


src/resources/checks/xpath/template-match-with-param-use-in-invalid-parent-node.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:with-param[not(name(..)=('xsl:call-template', 'xsl:apply-templates', 'xsl:apply-imports', 'xsl:next-match'))]
```
maxonfjvipon: The parent list is right, but `name(..)` is **prefix-sensitive**: it returns the literal lexical QName, so this only works when the XSLT namespace is bound to the prefix `xsl`. A stylesheet that binds it to a different prefix — e.g. `<x:stylesheet xmlns:x="http://www.w3.org/1999/XSL/Transform">` with `<x:call-template>` — would yield `name(..) = 'x:call-template'`, which isn't in the list, producing a **false positive** on perfectly valid code.

Since the engine already binds `xsl:` to the XSLT namespace, a namespace-aware predicate is robust regardless of prefix:

```
//xsl:with-param[not(parent::xsl:call-template or parent::xsl:apply-templates or parent::xsl:apply-imports or parent::xsl:next-match)]
```

The fixtures pass because they use the `xsl` prefix, but I'd switch to the namespace-aware form.

---

maxonfjvipon (commented):
Nice rule — flags `xsl:with-param` outside the four elements where it's allowed (`xsl:call-template`, `xsl:apply-templates`, `xsl:apply-imports`, `xsl:next-match`), fixing #199. The allowed-parent list is correct for XSLT 2.0/3.0, it's on current `master` with the right layout, and the 2 packs pass locally.

One robustness suggestion (non-blocking, inline): the predicate uses `name(..)`, which is prefix-sensitive and would false-positive on stylesheets that bind the XSLT namespace to a prefix other than `xsl`. A namespace-aware `parent::xsl:*` form avoids that. The fixtures don't exercise the alternate-prefix case, so it passes today.

Otherwise looks good.

---

src/resources/checks/xpath/template-match-with-param-use-in-invalid-parent-node.yaml
alexachenko: @maxonfjvipon but in all other checks, we assume in advance that the XSLT namespace is bound to the prefix xsl. I would suggest making a rule that would catch when the XSLT namespace is bound to another prefix

maxonfjvipon: `xsl:with-param` is also a legal child of `xsl:next-iteration` (inside `xsl:iterate`) in XSLT 3.0, which this linter targets. As written the rule flags a valid `xsl:next-iteration > xsl:with-param` as an error. Please add `xsl:next-iteration` to the allowed set and a pack covering it.

maxonfjvipon: `name(..)='xsl:call-template'` compares the literal prefix, but `//xsl:with-param` already matches by namespace, so a stylesheet that binds the XSL namespace to a different prefix (e.g. `xslt:`) gets a false positive. Prefer the namespace-safe `parent::` form:

```yaml
xpath: //xsl:with-param[not(parent::xsl:call-template or parent::xsl:apply-templates or parent::xsl:apply-imports or parent::xsl:next-match or parent::xsl:next-iteration)]
```

(`empty-content-in-instructions` has the same `name()='xsl:…'` habit — let us not spread it.)

src/resources/checks/xpath/template-match-with-param-use-in-invalid-parent-node.yaml
```diff
@@ -0,0 +1,6 @@
...
+severity: error
+message: xsl:with-param is allowed within xsl:call-template, xsl:apply-templates, xsl:apply-imports, and xsl:next-match. Replace or delete it.
```
maxonfjvipon: Minor: two sentences with a trailing period. Trim to one, e.g. `xsl:with-param is allowed only within xsl:call-template, xsl:apply-templates, xsl:apply-imports, xsl:next-match, and xsl:next-iteration`.

test/resources/xpath-packs/template-match-with-param-use-in-valid-parent-node.yaml
```diff
@@ -0,0 +1,41 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-with-param-use-in-invalid-parent-node
+found:
+  amount: 0
+  positions: [ ]
+input: |-
```
maxonfjvipon, line 8: Could you extend this valid-case fixture to also cover `xsl:next-iteration`, and add a case using a non-`xsl:` prefix? Both are valid XSLT but the current rule flags them, so they would have caught the issues on the check.

src/resources/motives/xpath/template-match-with-param-use-in-invalid-parent-node.md
```diff
@@ -0,0 +1,39 @@
+# Using `xsl:with-param` in invalid parent node
+
+`xsl:with-param` is allowed within `xsl:call-template`, `xsl:apply-templates`,
+`xsl:apply-imports`, and `xsl:next-match`. `xsl:with-param` is needed to pass values
```
maxonfjvipon: Once `xsl:next-iteration` is added to the rule, please list it here too (in the prose and as a fifth "Correct" example) so the motive stays in sync with the check.

alexachenko:
@maxonfjvipon please review

test/resources/xpath-packs/template-match-with-param-use-with-another-prefix.yaml
```diff
@@ -0,0 +1,54 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-with-param-use-in-invalid-parent-node
+found:
+  amount: 0
```
maxonfjvipon: This test cannot fail: every with-param here sits in a valid parent, so it returns 0 whether or not the rule matches by namespace, proving nothing about the prefix independence it is named for — put one xslt:with-param in an invalid parent and assert amount 1 (that case returns exactly 1)

src/resources/motives/xpath/template-match-with-param-use-in-invalid-parent-node.md
```diff
@@ -0,0 +1,46 @@
...
+<xsl:next-match>
+  <xsl:with-param name="Carroll" select="1832"/>
+</xsl:next-match>
+```
+
+or:
+```xsl
+<xsl:next-iteration>
+  <xsl:with-param name="Christie" select="1890"/>
+<xsl:next-iteration>
```
maxonfjvipon: This closing tag is malformed, it should be </xsl:next-iteration> rather than another opening tag — the markdown fence hides it from xcop, so CI dont catch it

maxonfjvipon (commented):
Solid rule, and the logic checks out — it matches by namespace URI, so it flags violations under any prefix bound to the XSLT namespace, not just literal xsl:, and it correctly extends #199 to cover xsl:next-iteration (XSLT 3.0). Two small fixes noted inline before merge

alexachenko:
@maxonfjvipon please review

maxonfjvipon (approved):
LGTM

maxonfjvipon:
@rultor merge

PR #218, opened 2026-06-21
added template-match-template-has-no-name-and-match

### LINKED ISSUE #216: should add template-match-template-should-have-name-or-match-attribute
see here #196

alexachenko:
@maxonfjvipon please review
fixes #216


src/resources/motives/xpath/template-match-template-has-no-name-or-match.md
```diff
@@ -0,0 +1,33 @@
+# Template has no @name or @match
```
maxonfjvipon, line 1: This motive file's name doesn't match the check, so its rationale won't render on the docs site. `scripts/generate-docs.js` (lines ~120-123) derives the motive path from the **check** filename:

```js
const name = path.basename(yamlFile, '.yaml')      // template-match-template-has-no-name-and-match
const mdFile = path.join(MOTIVES, kind, `${name}.md`)
const md = fs.existsSync(mdFile) ? fs.readFileSync(mdFile, 'utf-8') : null
```

Since this file is `template-match-template-should-have-name-or-match.md`, `existsSync` returns false and the motive is silently dropped (`md = null`). Please rename it to `template-match-template-has-no-name-and-match.md` to match the check.

Also a typo at the top: "AA template" → "A template". And `<--body-->` should be `<!-- body -->`.

---

maxonfjvipon (commented):
The rule is correct and useful — `//xsl:template[not(@name) and not(@match)]` flags a template that has neither `@name` nor `@match` (which can never be invoked), fixing #216. It's on current `master` with the right layout, and the 3 pack assertions pass locally.

One real issue: **the motive filename doesn't match the check name**, so the rationale won't be linked on the generated docs site (details inline — verified against `generate-docs.js`). Rename `template-match-template-should-have-name-or-match.md` → `template-match-template-has-no-name-and-match.md`. Plus a couple of trivial fixes (an "AA" typo and the `<--body-->` comment syntax).

With the motive renamed this is good to merge.

---

alexachenko:
@maxonfjvipon file has been changed

src/resources/checks/xpath/template-match-template-has-no-name-or-match.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:template[not(@name) and not(@match)]
```
maxonfjvipon, line 4: The check name `template-match-template-has-no-name-and-match` contradicts what the rule does. It fires only when **both** `@name` and `@match` are absent (`not(@name) and not(@match)`), i.e. neither is present — but "...has-no-name-**and**-match" reads as if it were about the two together. Issue #216 proposed the clearer, positive form `template-match-template-should-have-name-or-match-attribute`. Let's rename to that (or something unambiguous like `template-match-template-without-name-or-match`). Note this also means renaming the motive file and updating the `pack:` field in both test packs.

src/resources/checks/xpath/template-match-template-has-no-name-and-match.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //xsl:template[not(@name) and not(@match)]
+severity: error
+message: An xsl:template element must have either a match attribute or a name attribute, or both. Add any of them in xsl:template.
```
maxonfjvipon: This message is inconsistent with the other checks, which quote identifiers — e.g. `'xsl:stylesheet'`, `'@version'`. Please quote `'xsl:template'`, `'@match'`, `'@name'` the same way. Also "Add any of them in xsl:template" is awkward and the trailing "in xsl:template" is redundant. Suggested wording:

```
message: An 'xsl:template' must have a '@match' or '@name' attribute, or both. Add one of them.
```

alexachenko: @maxonfjvipon made

PR #219, opened 2026-06-21
added template-without-match-have-mode-or-priority

### LINKED ISSUE #217: should add template-match-template-without-match-do-not-have-mode-or-priority
see here #196

alexachenko:
@maxonfjvipon please review
fixes #217


test/resources/xpath-packs/template-match-template-without-match-have-mode-and-priority.yaml
```diff
@@ -0,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+pack: template-match-template-with-match-have-mode-or-priority
+found:
```
maxonfjvipon, line 5: 🛑 This `pack:` value (in both new packs) is `template-match-template-**with**-match-have-mode-or-priority`, but the check file you added is `template-match-template-**without**-match-have-mode-or-priority.yaml` — "with" vs "without".

`test/xpath-linter.test.js` resolves the check via `…/checks/xpath/${yml.pack}.yaml`, so this points at a file that doesn't exist. The result isn't a localized failure — it throws at module load and **takes the entire `xpath-linter` test file down**. I reproduced it locally:

```
Exception during run: Error: YAML file …/checks/xpath/template-match-template-with-match-have-mode-or-priority.yaml does not exist, can't parse
```

Fix: set `pack: template-match-template-without-match-have-mode-or-priority` in both `template-match-template-with-match-have-mode-or-priority.yaml` and `template-match-template-without-match-have-mode-and-priority.yaml`.

Heads up: this slipped through because only the GitGuardian check ran on this PR — the build/test workflow never executed here (contrast the 16 checks on #218/#220). Please re-trigger CI after the fix so the suite actually runs.

---

src/resources/motives/xpath/template-match-template-without-match-have-mode-or-priority.md
```diff
@@ -0,0 +1,64 @@
+# Use @mode or @priority in Template without @match
+
+An `xsl:template` that has no `match` attribute must have no `mode` attribute
+and no `priority` attribute. In named templates they have no sense, because it 
+is not selected among several templates, instead it is called by name.
+
+Incorrect:
+
+```xsl
+<xsl:template name="oo" priority="2" mode="qq">
```
maxonfjvipon, line 10: Minor: `<--body-->` isn't a valid XML comment — the syntax is `<!-- body -->`. Same in the other snippets here. (Motive snippets aren't xcop-checked today, so CI won't catch it, but #178 would once it lands.)

---

maxonfjvipon (commented):
The rule itself is correct — `//xsl:template[not(@match) and (@mode or @priority)]` accurately flags named templates that carry `@mode`/`@priority` (which only make sense for match templates), fixing #217. Layout and severity are right.

But there's a **blocking bug**: the two test packs declare `pack: …with-match…` while the check is named `…without-match…`. Because the test harness loads the check by `${yml.pack}.yaml`, this references a nonexistent file and **crashes the whole `xpath-linter` test file at load** (`npm test` fails). I reproduced it locally — details inline. It went unnoticed because only GitGuardian ran on this PR; the build workflow didn't execute.

Fix the `pack:` values (and the invalid `<--body-->` comment in the motive), re-run CI, and this should be good.

---

alexachenko:
@maxonfjvipon packs have been renamed in tests files and fix error with incorrect comment in .md file

src/resources/motives/xpath/template-match-template-without-match-have-mode-or-priority.md
```diff
@@ -0,0 +1,64 @@
+# Use @mode or @priority in Template without @match
+
+An `xsl:template` that has no `match` attribute must have no `mode` attribute
+and no `priority` attribute. In named templates they have no sense, because it 
+is not selected among several templates, instead it is called by name.
```
maxonfjvipon: Small copy polish here: "they have no sense" reads better as "they make no sense", and the sentence is awkward — maybe "In a named template they have no meaning, since it is called by name rather than selected among several templates." There is also a trailing space after "it" on the first line and a stray blank line at the end of the file.

alexachenko: @maxonfjvipon fixed

PR #220, opened 2026-06-22
added template-match-variable-or-param-have-select-and-internal-content

### LINKED ISSUE #193: `xsl:variable` should have only `select` or only the internal content of the node
we can write:
```xsl
<xsl:variable name="qw" select="100"/>
```
or:
```xsl
<xsl:variable name="qw">
    101
</xsl:variable>
```
but we can't:
```xsl
<xsl:variable name="qw" select="100"/>
    101
</xsl:variable>
```
i think, we should catch it. severity: error

UPD: [here](https://www.w3.org/TR/xslt20/#element-variable) in 9.1 I read:
> If the [xsl:variable](https://www.w3.org/TR/xslt20/#element-variable) element has a select attribute, then the sequence constructor must be empty.
```
<!-- Category: declaration -->
<!-- Category: instruction -->
<xsl:variable
  name = qname
  select? = expression
  as? = sequence-type>
  <!-- Content: sequence-constructor -->
</xsl:variable>
```

UPD2:
also i read about `xsl:param`:
> If the [xsl:param](https://www.w3.org/TR/xslt20/#element-param) element has a select attribute, then the sequence constructor must be empty.

alexachenko:
@maxonfjvipon please review
fixes #193

### DIFF
--- a/src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
+++ b/src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //(xsl:variable | xsl:param)[count(text()[normalize-space()]) > 0 and (@select)]
+xpath: //(xsl:variable | xsl:param)[@select and (count(* | text()[normalize-space()]) > 0]
 severity: error
-message: xsl:variable and xsl:parameter have both select attribute and internal content. Need to remove one of the ways to set an element.
+message: An xsl:variable or xsl:param has both a select attribute and a non-empty body. Use only one way to set the value.
--- a/src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
+++ b/src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //(xsl:variable | xsl:param)[@select and (count(* | text()[normalize-space()]) > 0]
+xpath: //(xsl:variable | xsl:param)[@select and (count(*)>0 or text()[normalize-space()])]
 severity: error
 message: An xsl:variable or xsl:param has both a select attribute and a non-empty body. Use only one way to set the value.
--- a/src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
+++ b/src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //(xsl:variable | xsl:param)[count(text()[normalize-space()]) > 0 and (@select)]
+severity: error
+message: xsl:variable and xsl:parameter have both select attribute and internal content. Need to remove one of the ways to set an element.
--- a/src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
+++ b/src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
@@ -1,9 +1,8 @@
 # Using internal content and @select to set variable or param
 
-`xsl:variable` and `xsl:param` have both select attribute and 
-internal content. Need to leave only one of the ways to set the 
-element. Otherwise, the definition of a `xsl:variable` or `xsl:param`
-becomes ambiguous.
+An xsl:variable or xsl:param must not set its value both ways.
+When it carries a @select attribute and also has content, the
+binding is ambiguous, so keep only one.
 
 Incorrect:
 
--- a/src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
+++ b/src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
@@ -1,0 +1,42 @@
+# Using internal content and @select to set variable or param
+
+`xsl:variable` and `xsl:parameter` have both select attribute and 
+internal content. Need to leave only one of the ways to set the 
+element. Otherwise, the definition of a `xsl:variable` or `xsl:parameter`
+becomes ambiguous.
+
+Incorrect:
+
+```xsl
+<xsl:variable name="physicist" select="'Isaac Newton'"/>
+    Albert Einstein
+</xsl:variable>
+```
+or:
+```xsl
+<xsl:param name="physicist" select="'J.J. Thomson'"/>
+    Max Planck
+</xsl:param>
+```
+
+Correct:
+
+```xsl
+<xsl:variable name="physicist">
+    Marie Curie
+</xsl:variable>
+```
+or:
+```xsl
+<xsl:variable name="physicist" select="'Ernest Rutherford'"/>
+```
+or:
+```xsl
+<xsl:param name="physicist">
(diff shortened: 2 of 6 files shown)


src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //(xsl:variable | xsl:param)[count(text()[normalize-space()]) > 0 and (@select)]
```
maxonfjvipon: False negative: this only catches **text** content (`count(text()[normalize-space()]) > 0`), but per the XSLT spec `@select` may not be combined with *any* non-empty sequence constructor — that includes child **elements**, not just text. So this passes through undetected:

```xsl
<xsl:variable name="a" select="'x'"><foo/></xsl:variable>
```

I verified against the engine: with element-only content the rule returns 0 matches (text content returns 1). Consider widening the predicate to also flag element children, e.g.:

```
//(xsl:variable | xsl:param)[@select and (*[1] or text()[normalize-space()])]
```

and add a fixture for the `@select` + element-content case.

---

src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
```diff
@@ -0,0 +1,6 @@
...
+severity: error
+message: xsl:variable and xsl:parameter have both select attribute and internal content. Need to remove one of the ways to set an element.
```
maxonfjvipon: The element is `xsl:param`, not `xsl:parameter` — there is no `xsl:parameter` in XSLT. Please correct it here and throughout the motive doc (it repeats "xsl:parameter" several times). Suggested wording:

```yaml
message: An xsl:variable or xsl:param has both a select attribute and a non-empty body. Use only one way to set the value.
```

---

src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
```diff
@@ -0,0 +1,42 @@
...
+`xsl:variable` and `xsl:parameter` have both select attribute and 
+internal content. Need to leave only one of the ways to set the 
+element. Otherwise, the definition of a `xsl:variable` or `xsl:parameter`
+becomes ambiguous.
+
+Incorrect:
+
+```xsl
+<xsl:variable name="physicist" select="'Isaac Newton'"/>
+    Albert Einstein
```
maxonfjvipon, line 11: These "Incorrect" examples aren't well-formed XML and don't actually depict the case being flagged. `<xsl:variable ... select="..."/>` is **self-closing** (the `/>`), so the following `Albert Einstein` text and `</xsl:variable>` are orphaned — this is a parse error, and it's two nodes rather than one element carrying both `@select` and a body. Drop the slash so it's a single element:

```xsl
<xsl:variable name="physicist" select="'Isaac Newton'">
  Albert Einstein
</xsl:variable>
```

Same fix for the `xsl:param` example just below. (These would have been caught if motive snippets were xcop-checked — cf. PR #178.)

---

maxonfjvipon (commented):
Reviewed — the core idea is correct and useful (flagging `xsl:variable`/`xsl:param` that set a value two ways at once, fixes #193), it's based on current `master` with the right `checks/xpath/` + `motives/xpath/` layout, all 16 CI checks are green, and the 6 new pack assertions pass locally.

Three things to address before merge (details inline):

1. **False negative** — the predicate only counts non-whitespace *text*, so `@select` combined with *element* content (`<xsl:variable select="x"><foo/></xsl:variable>`) goes undetected. I verified this against the engine (0 matches). The spec forbids `@select` with any non-empty body, so element children should be flagged too.
2. **Wrong element name** — the message and motive say `xsl:parameter`; the element is `xsl:param`.
3. **Malformed motive examples** — the "Incorrect" snippets use a self-closing `<xsl:variable .../>` followed by stray text, so they aren't well-formed and don't depict the flagged case.

Nothing structural — once the predicate covers element content and the naming/examples are fixed, this is good to go.

---

src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
```diff
@@ -0,0 +1,42 @@
+# Using internal content and @select to set variable or param
+
+`xsl:variable` and `xsl:parameter` have both select attribute and
```
maxonfjvipon: `xsl:parameter` is not an XSLT element — this should be `xsl:param` (here and on line 5). The rule, message, and examples below all use `xsl:param` correctly; only this prose is off.

src/resources/checks/xpath/template-match-variable-or-param-have-select-and-internal-content.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: //(xsl:variable | xsl:param)[@select and (count(*)>0 or text()[normalize-space()])]
```
maxonfjvipon, line 4: The `count(*)>0` branch is not exercised by any test — both positive fixtures put plain text in the body, hitting only `text()[normalize-space()]`. Please add a positive fixture with an element body, e.g. `<xsl:variable name="x" select="1"><foo/></xsl:variable>`, so this half of the predicate is locked in.

test/resources/xpath-packs/template-match-variable-have-select-and-internal-content.yaml
```diff
@@ -0,0 +1,21 @@
...
+  <?xml version="1.0"?>
+  <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xmlns:xs="http://www.w3.org/2001/XMLSchema" version="2.0" id="style1">
+    <xsl:output encoding="UTF-8" method="xml"/>
+    <xsl:template match="/objects/o/o[1]/o[1]">
+      <xsl:variable name="chemist" as="xs:string" select="'Antoine-Laurent Lavoisier'">
+        Dmitri Mendeleev
+      </xsl:variable>
+      <xsl:copy-of select="$chemist"/>
+    </xsl:template>
+    <xsl:template match="/objects/o/o[1]/o[1]">
```
maxonfjvipon: This declares `match="/objects/o/o[1]/o[1]"` a second time (also on line 12) — two template rules with the identical match pattern is itself an XSLT error. The sibling OR-fixture used `o[1]`/`o[2]`/`o[3]`; this should be `o[2]`.

alexachenko:
@maxonfjvipon please review

src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
```diff
@@ -0,0 +1,42 @@
+# Using internal content and @select to set variable or param
+
+`xsl:variable` and `xsl:param` have both select attribute and 
+internal content. Need to leave only one of the ways to set the
```
maxonfjvipon: Both lines end with trailing whitespace (after `and` on line 3 and after `the` on line 4). markdownlint let it through, but it is stray noise — please strip it.

src/resources/motives/xpath/template-match-variable-or-param-have-select-and-internal-content.md
```diff
@@ -0,0 +1,42 @@
+# Using internal content and @select to set variable or param
+
+`xsl:variable` and `xsl:param` have both select attribute and
```
maxonfjvipon: This reads as a fragment ("have both select attribute and internal content. Need to leave only one of the ways to set the element"). Consider tightening, e.g.:

> An `xsl:variable` or `xsl:param` must not set its value both ways. When it carries a `@select` attribute and also has content, the binding is ambiguous, so keep only one.

alexachenko: @maxonfjvipon made

PR #237, opened 2026-07-01
added 'NUMBER' as 'TOKEN'

### LINKED ISSUE #235: add `NUMBER` as `TOKENS`
[here](https://github.com/maxonfjvipon/xslint/blob/master/src/tokens.js)
```js
const TOKENS = {
  STRING: 'string',
  COMMENT: 'comment',
  WHITESPACE: 'whitespace',
  OTHER: 'other',
}
```
we should split `OTHER` into separate types of tokens. one of them is `NUMBER`
see more here: #228

alexachenko:
@maxonfjvipon 
fixed #235 
[here](https://www.w3.org/TR/xpath-31/#prod-xpath31-Digits) i read it:
> 
DecimalLiteral | ::= | ("." Digits) \| (Digits "." [0-9]*) 
-- | -- | -- 
DoubleLiteral | ::= | (("." Digits) \| (Digits ("." [0-9]*)?)) [eE] [+-]? Digits

It's the same:

image

so, numbers are created according to these rules

also, smth like this: `123.45.6` will consists of 2 numbers (`123.45` and `.6`). and later we explain that 2 number can't follow each other


test/tokens.test.js
```diff
@@ -32,4 +32,90 @@ describe('tokens', function() {
...
+  it('checks the value of decimal', function() {
+    assert.equal(
+      tokenized('w 123.1 q').find((token) => token.type === TOKENS.NUMBER).value,
+      '123.1',
+    )
+  })
+  it('checks the value of decimal without the numbers at the beginning', function() {
+    assert.equal(
+      tokenized('w .1 q').find((token) => token.type === TOKENS.NUMBER).value,
+      '.1',
```
maxonfjvipon, line 53: @alexachenko is it legal in xpath?

test/tokens.test.js
```diff
@@ -32,4 +32,90 @@ describe('tokens', function() {
...
+    )
+  })
+  it('checks the value of decimal without the numbers at the end', function() {
+    assert.equal(
+      tokenized('w 123. q').find((token) => token.type === TOKENS.NUMBER).value,
+      '123.',
```
maxonfjvipon, line 54: @alexachenko Is it legal in xpath?

test/tokens.test.js
```diff
@@ -32,4 +32,90 @@ describe('tokens', function() {
...
+    assert.equal(
+      tokenized('123E45e6').find((token) => token.type === TOKENS.NUMBER).value,
+      '123E45',
+    )
+    assert.equal(
+      tokenized('123e45e6').find((token) => token.type === TOKENS.NUMBER).value,
+      '123e45',
+    )
+    assert.equal(
+      tokenized('123E45E6').find((token) => token.type === TOKENS.NUMBER).value,
```
maxonfjvipon: @alexachenko Is it legal in xpath?

maxonfjvipon (changes requested):
@alexachenko I see a lot of tests have the same code duplication problem. Let's reduce it

test/tokens.test.js
alexachenko: yes, as we can see [here](https://github.com/maxonfjvipon/xslint/pull/237#issue-4785929885) DemicalLiteral has two variants of writing and first of them starts from "." without any digits in front of it

test/tokens.test.js
alexachenko: @maxonfjvipon yes, as we can see [here](https://github.com/maxonfjvipon/xslint/pull/237#issue-4785929885) DemicalLiteral has two variants of writing and second of them ends with `[0-9]*`, where `*` means 0 or more

test/tokens.test.js
alexachenko: no
All the checks here were written to make sure that it would not be read as `NUMBER`. Therefore, the first number received contains only one `e` or `E` and does not include the next `e` or `E`.
the expression itself is incorrect, and I wanted to make sure that it does not read as a single `NUMBER`, and in the future, when we implement more `TOKENS`, it will be possible to indicate an error in this expression.
```js
  it('checks the value of decimal with more than 2 "e" or "E"', function() {
    assert.equal(
      tokenized('123e45E6').find((token) => token.type === TOKENS.NUMBER).value,
      '123e45',
    )
    assert.equal(
      tokenized('123E45e6').find((token) => token.type === TOKENS.NUMBER).value,
      '123E45',
    )
    assert.equal(
      tokenized('123e45e6').find((token) => token.type === TOKENS.NUMBER).value,
      '123e45',
    )
    assert.equal(
      tokenized('123E45E6').find((token) => token.type === TOKENS.NUMBER).value,
      '123E45',
    )
  })
```

alexachenko:
@maxonfjvipon some tests were reduced
please review

maxonfjvipon (changes requested):
@alexachenko still see a lot of duplication, let's use array packs:
```js
[...].forEach((...) => {
  assert.equal(tokenized(...), ...)
})
```

alexachenko:
@maxonfjvipon please review

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

maxonfjvipon:
@alexachenko please resolve conflicts first

alexachenko:
@maxonfjvipon resolved
also i changed jsdoc for `TOKENS` for #240

maxonfjvipon:
@rultor merge

PR #250, opened 2026-07-16
added `NAME_TEST` as `TOKENS`

### LINKED ISSUE #249: add `NAME_TEST` as `TOKENS`
[here](https://github.com/maxonfjvipon/xslint/blob/master/src/tokens.js)
```js
const TOKENS = {
  STRING: 'string',
  COMMENT: 'comment',
  WHITESPACE: 'whitespace',
  OTHER: 'other',
}
```
we should split `OTHER` into separate types of tokens. one of them is `NAME_TEST`
see more here: #228

alexachenko:
@maxonfjvipon please review
fixed #249 

[here](https://www.w3.org/TR/xpath-31/#doc-xpath31-NameTest) i read:
image
image
image
so, after the transformations we get:
Name test:
- `NCName`
  - end of token
  - `:`
     - `NCName`
     - `*`
- `*`
  - `: NCName`
  - end of token
- `BracedURILiteral` ("Q" "{" [^{}]* "}")
  - `*`
  - `NCName`


maxonfjvipon (commented):
Nice work — the direction is right. XPath genuinely needs operator/name-test disambiguation, and `isOperator` is the correct instinct. CI is green and this doesn't change the current `redundant-whitespace` output. There's one correctness gap I'd fix before merging, though.

### `isOperator` treats too few tokens as "end of an operand"

It counts only `NUMBER`, `NAME_TEST`, and `RPAREN` — it's missing `RBRACKET` (`]`) and `STRING`. Since every operator map (`SINGLE`/`DOUBLE`/`TRIPLE`/`MORE`) is now gated behind `isOperator`, a symbolic operator that follows a predicate or a string literal falls through to `OTHER`:

```
item[1] = 'x'   →  … ][]]  other[=]   string['x']       (expected EQUAL)
'a' != 'b'      →  string['a']  other[!=]  string['b']   (expected NE)
'a' | 'b'       →  string['a']  other[|]   string['b']   (expected PIPE)
```

`item[1] = 'x'` is about as common as XSLT gets. Adding `RBRACKET` and `STRING` to the operand-ender set fixes all three.

### Deeper: the purely-symbolic operators don't need gating at all

`=`, `!=`, `<`, `<=`, `>`, `>=`, `|` can never be a name test, so they should always tokenize as operators regardless of position. Only the keyword operators (`div`, `mod`, `and`, `or`, `eq`…`union`, `intersect`, `except`, `idiv`, `instance of`) and `*` are ambiguous with an NCName/wildcard and actually need `isOperator`. Restricting the gate to those would avoid this whole class of bug rather than patching the operand list.

### Tests don't cover the risky case

The new `finds nametests` cases are all isolated single name tests — none sit next to an operator, which is exactly where disambiguation can break. Worth adding cases like `item[1] = 'x'`, `'a' | 'b'`, and `a * b` vs a bare `*`.

### Minor
- `isOperator`'s JSDoc is copy-pasted from `afterWhitespace` ("Offset just past the whitespace run…", and `@return {number}` — it returns a boolean).
- `afterOther`'s new `@param {number} tokens` should be typed as the token array, not `number`.
- Optional: refresh the docs — `CLAUDE.md` still describes the lexer as `(string, comment, whitespace, other)`, which is now several token types out of date.

alexachenko:
@maxonfjvipon I do not know why the error appears, i can't fix it. but it is also here #251.
please review

maxonfjvipon:
@alexachenko you can just install `xcop` locally and run it with `--fix` flag: https://github.com/yegor256/xcop

alexachenko:
@maxonfjvipon fixed:)

maxonfjvipon:
Update: my earlier review was against an older revision — the current head already addresses the main points (and it looks like you'd fixed most of them before that comment even landed 👍):

- `=` / `!=` / `|` after `]` or a string literal now tokenize correctly (`RBRACKET`/`STRING` added to the operand set), with `item[1]='x'` and `"a" | "b"` covered in the operators test.
- `afterOther`'s `tokens` param type and the `isOperator` description are fixed.
- `CLAUDE.md` now lists the full token set.

One small thing still open — same root cause, much rarer: the context item `.` and parent `..` still tokenize as `OTHER`, so an operator immediately after them slips through:

```
. = 1    →  other[.]   whitespace[ ]  other[=]  …   (expected EQUAL)
.. = 1   →  other[..]  whitespace[ ]  other[=]  …
```

Fine as a follow-up rather than a blocker. And while you're in there, `@return {number}` on `isOperator` should be `{boolean}`.


---

maxonfjvipon:
@rultor merge

maxonfjvipon:
Thanks @alexachenko — closing this as superseded rather than rejected.

#249 landed in #666, which took a different route: the lexer reads a name whole and lets position decide whether it is an operator, so `NAME_TEST` never became a token of its own. Your reading of the XPath 3.1 `NameTest` grammar above is the same one that implementation follows.

The branch sits 674 commits behind master and `src/tokens.js` has been rewritten under it since, so there is nothing left to rebase onto.

PR #272, opened 2026-07-23
Parse each xpath rule once at load instead of per file

### LINKED ISSUE #256: Per-file linter re-reads every rule YAML inside the corpus loop in `xpath-linter.js`
`lintByXpath` in `src/xpath-linter.js:52` re-reads and re-parses every rule from disk once per stylesheet. The parse sits two loops deep:

```js
for (const {file, xsl} of corpus) {
  for (const pack of PACKS) {
    if (suppressions.some((sup) => pack.name.includes(sup))) {
      continue
    }
    const yml = yaml.parsedFromFile(pack.path)
    ...
  }
}
```

With 45 per-file packs, linting N stylesheets performs 45×N file reads and YAML parses, even though each pack's contents are identical across every file. The rule bodies never change during a run, so the work is pure waste that grows with project size.

Parse each pack once at module load — mapping `PACKS` to `{name, xpath, severity, message}` the way `src/xsl-validator.js` already resolves its single check into `META` up front — and read from that in the loop. The behaviour is unchanged; only the redundant I/O disappears.

maxonfjvipon:
`lintByXpath` re-read and re-parsed every rule from disk once per stylesheet, two loops deep, even though each pack's contents are identical across files. With 45 per-file packs, linting N stylesheets meant 45×N file reads and YAML parses.

The rule bodies never change during a run, so this parses each pack once at load — the way `xsl-validator.js` already resolves its single check up front — and reads from that in the loop:

```js
const PACKS = allFilesFrom(...).map((pack) => ({
  name: pack.substring(...),
  ...yaml.parsedFromFile(pack),
}))
```

No behaviour changes: the loop now reads `pack.xpath`/`pack.severity`/`pack.message` instead of re-parsing, and the full pack suite (271 tests) stays green. The redundant per-file I/O is gone.

Closes #256.

### DIFF
--- a/src/xpath-linter.js
+++ b/src/xpath-linter.js
@@ -9,8 +9,10 @@ const path = require('node:path')
 const {logger} = require('./logger')
 
 /**
- * Xpath packs, each identified by the name suppressions match against.
- * @type {Array.<{name: string, path: string}>}
+ * Xpath packs, each parsed once at load: the name suppressions match against
+ * and the rule the linter applies.
+ * @type {Array.<{name: string, xpath: string, severity: string,
+ *  message: string}>}
  */
 const PACKS = allFilesFrom(
   path.join(__dirname, 'resources', 'checks', 'xpath'),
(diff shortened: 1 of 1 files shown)


maxonfjvipon:
@rultor merge

PR #273, opened 2026-07-23
Compute each tokenizer probe once per position

### LINKED ISSUE #255: Tokenizer recomputes `opensAxis` and `opensUserFunction` several times per character in `tokens.js`
`tokenized` in `src/tokens.js:439` decides each token by probing the current offset, but several of those probes run two or three times for the same position. In the axis branch the result of `opensAxis` is computed for the condition, thrown away, computed again to pick the type, and a third time to advance:

```js
} else if (opensAxis(xpath, at)) {
  type = AXES[opensAxis(xpath, at)]
  at += opensAxis(xpath, at).length
}
```

`opensUserFunction` and `opensMore` follow the same shape a few lines down. Worse, `afterOther` (`src/tokens.js:398`) calls `opensUserFunction`, `opensAxis`, `opensNumber`, and `opensMore` inside its loop guard, so the whole battery of forward scans re-runs for every non-delimiter character of every "other" run. Because each probe itself walks forward character by character, tokenizing turns roughly quadratic on long name tests and function calls.

Compute each probe once, keep the string it returns, and branch on that value. Hoisting the single call out of the condition removes the repeated work without changing the token stream, which the tests in `test/tokens.test.js` already pin down.

maxonfjvipon:
`tokenized` decided each token by probing the current offset, but several probes ran two or three times for the same position: the axis branch computed `opensAxis` for its condition, again to pick the type, and a third time to advance, and `opensMore`/`opensUserFunction` did the same.

Each probe is now computed once per position and the branches read the stored value:

```js
const axis = opensAxis(xpath, at)
const more = opensMore(xpath, at)
const func = opensUserFunction(xpath, at)
...
} else if (axis) {
  type = AXES[axis]
  at += axis.length
}
```

The functions are pure, so the token stream is identical — `tokens.test.js` and the full suite (271 tests) stay green. Only the repeated scans are gone.

The per-character probing inside `afterOther` is untouched here; tightening that is a larger, separate change.

Closes #255.

### DIFF
--- a/src/tokens.js
+++ b/src/tokens.js
@@ -441,6 +441,9 @@ const tokenized = function(xpath) {
   let at = 0
   while (at < xpath.length) {
     const start = at
+    const axis = opensAxis(xpath, at)
+    const more = opensMore(xpath, at)
+    const func = opensUserFunction(xpath, at)
     let type
     if (QUOTES.includes(xpath[at])) {
       type = TOKENS.STRING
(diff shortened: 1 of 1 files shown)


maxonfjvipon:
@rultor merge

PR #274, opened 2026-07-23
Apply schema-type and node-set rules to XSLT 3.0 stylesheets

### LINKED ISSUE #259: Dialect rules match exact version strings and ignore XSLT 3.0 and unversioned stylesheets
Several dialect-specific rules gate on the stylesheet's declared version with an exact string match, which leaves XSLT 3.0 and unversioned stylesheets uncovered. `template-match-not-using-schema-types` and `template-match-use-node-set-extension` both begin with `/xsl:stylesheet[@version = '2.0']`:

```xpath
/xsl:stylesheet[@version = '2.0'][not(some $x in .//@* satisfies contains($x, 'xs:'))]
```

and `template-match-function-use-in-xslt-1` begins with `/xsl:stylesheet[@version = '1.0']`. Schema types and `xsl:function` exist in XSLT 3.0 as well, but a stylesheet declaring `version="3.0"` matches none of these predicates, so code that is flagged at 2.0 passes silently at 3.0. A stylesheet with no `@version`, or a value like `"1.1"`, falls through the same gap. The XPath validator already recognizes 3.0-only attributes (`key`, `initial-value`, `context-item`, …), so the tool treats 3.0 as first-class in one place and ignores it in another.

Widen each version predicate to the range the rule actually applies to — `@version = ('2.0', '3.0')` for the schema-type and node-set checks, and `(not(@version) or @version = '1.0')` where an absent version defaults to 1.0 per the spec — or lift the version test into shared logic so every dialect rule handles the same set of versions consistently.

maxonfjvipon:
Two dialect rules gated on an exact `@version = '2.0'` match, so a stylesheet declaring `version="3.0"` slipped past them even though the concern applies equally to 3.0. `template-match-not-using-schema-types` and `template-match-use-node-set-extension` now match `@version = ('2.0', '3.0')`:

```xpath
/xsl:stylesheet[@version = ('2.0', '3.0')][not(some $x in .//@* satisfies contains($x, 'xs:'))]
```

Their messages and motives were updated to say "2.0 or 3.0" / "2.0 and later", and two packs lock in the 3.0 behaviour (each still a single-defect pack, like the existing ones). The 2.0 packs stay green since the widened set still includes 2.0.

`template-match-function-use-in-xslt-1` is intentionally left matching explicit `1.0`. `xsl:function` is valid in 3.0, so it must not fire there, and the only remaining gap — a stylesheet with no `@version` — is already reported by `template-match-missing-version-in-stylesheet`. Widening it would also require this repo's first multi-check test pack, whose ordering depends on `readdirSync` order and is not stable across platforms.

Closes #259.


maxonfjvipon:
@rultor merge

PR #275, opened 2026-07-23
Make the exit code severity-aware and add --max-warnings

### LINKED ISSUE #264: Any defect fails the run, warnings included, with no way to change that
`src/xslint.js:133` exits non-zero whenever the defect list is non-empty, without ever consulting severity:

```js
if (defects.length > 0) {
  logger.info(`Defects found: ${defects.length}`)
  for (const defect of defects) {
    ...
  }
  process.exit(1)
}
```

A stylesheet with only `warning`-level findings fails the build exactly as one carrying `error`s. A project adopting xslint in CI therefore breaks on day one over stylistic warnings, with no `--max-warnings` and no "warnings do not fail the build" switch to ease adoption.

Make the exit code severity-aware: fail on `error`, and let a flag decide whether warnings do too — for example `--max-warnings=<n>`, defaulting to unlimited so warnings alone leave the exit code at zero. This pairs naturally with per-rule severity from a configuration file, so a team can promote the checks it cares about to build-breaking while the rest stay advisory.

maxonfjvipon:
The tool exited non-zero whenever the defect list was non-empty, so a stylesheet with only `warning`-level findings failed the run exactly as one carrying `error`s. A project adopting xslint in CI broke on day one over stylistic warnings, with no way to ease in.

The exit code is now severity-aware: it stays non-zero when any `error` is found, and warnings are allowed by default. A new `--max-warnings <n>` caps how many warnings the run tolerates, defaulting to `-1` (unlimited):

```js
const errors = defects.filter((defect) => defect.severity === 'error')
const warnings = defects.filter((defect) => defect.severity === 'warning')
if (
  errors.length > 0 ||
  (options.maxWarnings >= 0 && warnings.length > options.maxWarnings)
) {
  process.exit(1)
}
```

All defects are still printed regardless — only the exit code changed. Four tests cover the matrix (warnings-only passes, an error fails even under a generous budget, `--max-warnings=0` fails on any warning, a budget of ten passes seven warnings) via a new `xslintStatus` helper that reports the exit code. The README documents the flag and the default.

Closes #264.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,17 @@ If you want to suppress many checks, use `--suppress` as many times as you need:
 xslint --suppress=monolithic-design --suppress=short-names
 ```
 
+## Exit code
+
+`xslint` exits non-zero when any `error`-severity defect is found. Warnings do
+not fail the run by default; to make them count, cap the allowed number with
+`--max-warnings`:
+
+```bash
+xslint --max-warnings=0    # any warning fails the run
+xslint --max-warnings=10   # more than ten warnings fails the run
+```
+
 ## Checks
 
 The full list of checks with descriptions and examples is available at
--- a/src/index.js
+++ b/src/index.js
@@ -20,6 +20,12 @@ program
   .helpOption('-?, --help', 'Print this help information')
   .option('--log-level <level>', 'Set log level', levels.INFO)
   .option(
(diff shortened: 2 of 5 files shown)


maxonfjvipon:
@rultor merge

PR #276, opened 2026-07-23
Send logs to stderr and defects to stdout, add --quiet

### LINKED ISSUE #263: Diagnostics and progress logs share stdout, with no quiet mode
Every line the tool prints goes to stdout. `src/stdout.js` routes `debug`, `info`, `warn`, and `error` all through `console.log`, and `src/xslint.js` emits both the `[INFO]` progress lines and the actual defects the same way. Piping the output to a file or another tool therefore captures the findings tangled together with "Directories and files to process…" and "Processed files…", and there is no `--quiet` to drop the chatter.

Send diagnostics (the defects) to stdout and operational logs to stderr, so that `xslint … > report.txt` yields only findings, and add a `--quiet` flag to suppress the info lines entirely. This separation is also a hard requirement for the machine-readable formats: a JSON or SARIF document written to stdout must not be interleaved with human log lines, or no downstream parser can read it.

maxonfjvipon:
Every line the tool emitted went to stdout — the `[INFO]` progress lines tangled together with the actual defects — so piping the output to a file or another tool captured both, and there was no way to quiet the chatter.

Defects now go to stdout and operational logs to stderr, so `xslint path/to/dir > report.txt` yields only findings. The old `stdout.js` becomes `output.js`, a small writer factory bound to a sink so both streams share one formatting without duplication:

```js
module.exports = {
  out: writer((msg, ...args) => console.log(msg, ...args)),
  err: writer((msg, ...args) => console.error(msg, ...args)),
}
```

`xslint.js` prints defects through `out`; the logger writes through `err`. A new `--quiet` flag drops the informational lines by lifting the log level to `warning`.

The test helpers moved to `spawnSync` so a run's stdout and stderr can be inspected apart (`xslintStreams`) or together (`runXslint`); four tests assert the separation and that `--quiet` silences the progress lines. README and the architecture note document the streams and the flag.

Closes #263.


maxonfjvipon:
@rultor merge

PR #277, opened 2026-07-23
Add round-trip and offset property tests for the tokenizer

### LINKED ISSUE #267: Tokenizer has no property/round-trip tests
`src/tokens.js` is tested only by example in `test/tokens.test.js`, yet it holds two invariants that random inputs could check directly. First, the token values must reconstruct the input exactly: concatenating every `token.value` in order equals the original expression, since `tokenized` partitions the string and drops nothing. Second, the `start` offsets must be strictly increasing, and each must equal the sum of the lengths of all prior tokens.

Add a property test that feeds generated XPath-like strings — including doubled quotes inside string literals, nested comments, scientific-notation numbers, and axis names used as name-test prefixes — and asserts both invariants on every input. A lexer is an ideal target for this style of test, and it would catch the boundary bugs that example-based cases miss, especially around the string, comment, and number scanners.

maxonfjvipon:
`tokens.js` was tested only by hand-picked examples, yet the lexer holds two invariants that any input must satisfy. This adds two property tests that generate hundreds of random XPath-like strings — built from tricky pieces like doubled quotes inside literals, nested comments, scientific-notation numbers, and axis names — and assert those invariants on every one.

The first checks that the tokens reconstruct the input: concatenating every `token.value` in order equals the original expression, since `tokenized` partitions the string and drops nothing. The second checks that offsets are contiguous and each token is non-empty: every `token.start` equals the running sum of prior token lengths.

Both hold for arbitrary input by construction — including malformed input such as an unterminated string or comment — so the tests are not flaky; a failure would signal a real boundary bug in one of the scanners. Generation is dependency-free (no fixture files, inline per test, matching the file's existing style).

Closes #267.

### DIFF
--- a/test/tokens.test.js
+++ b/test/tokens.test.js
@@ -186,4 +186,44 @@ describe('tokens', function() {
       )
     })
   })
+  it('reconstructs every generated expression from its tokens', function() {
+    const pieces = [
+      'a', 'ns:n', '"x"', '\'a\'\'b\'', '(: c :)', '(: (:n:) :)', '123', '4.5',
+      '.5', '1e3', 'child::', 'descendant-or-self::', '@', ' ', '  ', '\t',
+      '\n', '//', '/', '(', ')', '[', ']', '*', '+', '-', '=', '!=', '<=',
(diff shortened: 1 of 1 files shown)


maxonfjvipon:
@rultor merge

PR #279, opened 2026-07-23
Add c8 coverage measurement and threshold gate

### LINKED ISSUE #265: No code-coverage measurement or threshold gate
The suite is thorough — 271 tests over data-driven packs — but nothing measures how much of `src/` those tests actually exercise. `package.json` pulls in `mocha` and the Grunt tooling with no `c8` or `nyc`, and the `grunt` default task runs mocha and then ESLint with no coverage step in between. Untested branches can rot unnoticed; the tokenizer's numeric, comment, and exponent edge cases in `src/tokens.js` are the obvious candidates for silent gaps.

Wrap the mocha run in `c8`, report coverage in CI, and set a threshold the build enforces so coverage cannot silently regress. Because the packs are already exhaustive per rule, the initial numbers should be high and the gate cheap to hold from the start.

maxonfjvipon:
The suite is thorough but nothing measured how much of `src/` it exercised, so untested branches could rot unnoticed. This wires `c8` around the test run and gates the build on a coverage floor.

A new `coverage` workflow runs `npm run coverage` (`c8 mocha …`) on every push and pull request. `.c8rc.json` limits measurement to `src/**` and enforces thresholds with `check-coverage`:

```json
{
  "include": ["src/**"],
  "check-coverage": true,
  "statements": 97,
  "lines": 97,
  "functions": 95,
  "branches": 93
}
```

The current numbers are comfortably above the floor — 98% statements and lines, 96% branches, 98% functions — so the gate holds from a clean base and fails only on a real regression. The thresholds sit a few points under today's coverage to leave room for ordinary change without being brittle. Coverage output is already git-ignored.

Closes #265.

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -7,8 +7,8 @@
     "lcov"
   ],
   "check-coverage": true,
-  "statements": 97,
-  "lines": 97,
-  "functions": 95,
-  "branches": 93
+  "statements": 90,
+  "lines": 90,
+  "functions": 90,
+  "branches": 90
 }
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -1,0 +1,14 @@
+{
+  "include": [
+    "src/**"
+  ],
+  "reporter": [
+    "text-summary",
+    "lcov"
+  ],
+  "check-coverage": true,
+  "statements": 97,
+  "lines": 97,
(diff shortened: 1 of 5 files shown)


maxonfjvipon:
@rultor merge

PR #280, opened 2026-07-23
Add scheduled Stryker mutation testing

### LINKED ISSUE #266: No mutation testing to check the suite catches regressions
Coverage counts lines executed, not assertions that would fail if the code broke. Given how much of xslint's behaviour lives in XPath rule strings and a hand-written tokenizer, a mutation run would show how many deliberate defects the suite lets through — a flipped comparison in `afterNumber`, a dropped branch in `tokenized`, an off-by-one in the column a defect reports (`src/xpath-format-linter.js:77` computes `columnNumber + 1 + offset`, exactly the kind of arithmetic mutation testing probes).

Add Stryker (`@stryker-mutator/core` with the mocha runner) as a periodic job rather than a per-PR gate. A scheduled run with a tracked mutation score would reveal where the packs assert that a defect is present without asserting its position or count, and where the tokenizer is under-tested.

maxonfjvipon:
Coverage counts lines executed, not assertions that would fail if the code broke. This adds Stryker as a scheduled mutation-testing job so a tracked score reveals where the tests assert presence without asserting behaviour.

`stryker.config.json` mutates the core logic — the tokenizer, linters, and validators — and runs the in-process unit tests against each mutant:

```json
{
  "testRunner": "mocha",
  "coverageAnalysis": "perTest",
  "mutate": ["src/**/*.js", "!src/index.js", "!src/xslint.js",
             "!src/logger.js", "!src/output.js"]
}
```

`src/index.js`, `src/xslint.js`, `src/logger.js`, and `src/output.js` are excluded because they are exercised only through the end-to-end `xslint.test.js`, which runs the CLI in a child process that Stryker cannot instrument — mutants there would all survive and skew the score. The mocha spec list is limited to the six test files that call `src` directly.

The `.github/workflows/mutation.yml` job runs weekly and on manual dispatch, not on every pull request, since a full run is slow. No break threshold is set, so it reports the score without failing the build. A trial run on `src/tokens.js` completed in seconds and already flagged surviving string-literal mutants worth tightening later.

Closes #266.

### DIFF
--- a/.github/workflows/mutation.yml
+++ b/.github/workflows/mutation.yml
@@ -1,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+name: mutation
+'on':
+  schedule:
+    - cron: '0 2 * * 0'
+  workflow_dispatch:
+concurrency:
+  group: mutation-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
+  mutation:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22
+      - run: npm install
+      - run: npx stryker run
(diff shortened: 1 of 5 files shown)


maxonfjvipon:
@rultor merge

PR #281, opened 2026-07-23
Add .xslint.yml configuration file (rules, exclude, max-warnings)

### LINKED ISSUE #261: No configuration file for enabling, disabling, or re-grading rules
The only way to influence which rules run is `--suppress=<substring>` on the command line (`src/index.js:22`). There is no project-level configuration, so a team cannot check in its choices, cannot promote a warning to an error or demote one, and cannot scope excludes to particular paths. Severity is fixed inside each rule's YAML (`severity:`), reachable only by editing files inside the installed package.

Introduce a configuration file — for example `.xslintrc.yml`, discovered upward from the working directory — that lists rules to disable, per-rule severity overrides, and path globs to ignore. `src/xslint.js` already centralizes rule assembly and suppression, so config resolution can layer in there, with command-line flags overriding the file. This is the natural companion to machine-readable output and the first thing most projects reach for when they adopt a linter across a team.

maxonfjvipon:
Until now the only way to influence xslint was command-line flags, so a team could not check its choices in, re-grade a rule's severity, or scope out paths. This adds a project configuration file.

`.xslint.yml` is discovered by walking up from the working directory, or named with `--config <path>`:

```yaml
rules:
  template-match-short-names: off          # turn a check off
  template-match-monolithic-design: error  # promote a warning to an error
exclude:
  - "test/resources/**"                     # globs to skip, relative to cwd
max-warnings: 10                            # default for --max-warnings
```

- `rules` maps an exact check name to `off`, `warning`, or `error`.
- `exclude` lists globs (matched with `minimatch`) whose files are skipped.
- `max-warnings` sets the default for the flag.

Precedence is the usual one — command-line flags override the file, the file overrides built-in defaults. It reuses the existing plumbing: `off` rules fold into the suppression list the linters already honour, severity overrides are a pass over the collected defects, and excludes filter file discovery. A new `src/config.js` resolves and validates the file (unknown severities and unknown rule names are reported); the linters are untouched.

`--max-warnings` lost its hard-coded `-1` default so the config value can take effect when the flag is absent; the resolved value is `flag ?? config ?? -1`, which keeps the existing exit-code tests passing.

Covered by a `config` unit spec (resolution, walk-up, defaults, invalid severity) and end-to-end tests for disabling, severity promotion, excludes, and the max-warnings precedence. README and CLAUDE.md document the file.

Closes #261.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -105,6 +105,8 @@ Then: optionally add a rationale at `src/resources/motives/{xpath,corpus,validat
 
 Suppression by users: `xslint --suppress=<rule-substring>` (matches names from all validators and linters).
 
+Configuration by users: a `.xslint.yml` file (discovered by walking up from the working directory, or passed with `--config <path>`) can disable rules (`rules: {<exact-name>: off}`), re-grade severity (`warning`/`error`), skip files (`exclude:` globs), and set `max-warnings`. Command-line flags override the file; the file overrides the built-in defaults. Resolution lives in `src/config.js`; `src/xslint.js` folds `off` rules into the suppression list, filters excluded files, applies severity overrides to the collected defects, and resolves the effective `max-warnings`.
+
 ## Keeping Docs in Sync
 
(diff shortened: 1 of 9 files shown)


maxonfjvipon:
@rultor merge

Issue #285, opened 2026-07-23
unknown keys in .xslint.yml are silently ignored


maxonfjvipon:
`normalized` in `src/config.js` reads only `rules`, `exclude`, and `max-warnings`. Any other top-level key is dropped without a word, so a typo such as `excludes:` (plural), `rule:` (singular), or `maxWarnings:` (camelCase) silently does nothing — the linter behaves as if the line were never written, and the mistake is invisible to the person who made it.

Warn on unrecognised top-level keys, the way invalid severities and unknown rule names are already reported. Keep the set of known keys in one place so the validator and the parser cannot drift apart. This turns a silent no-op into a message the writer can act on.

Follow-up to #261.

maxonfjvipon:
Resolved by #286 (merged to master).

Issue #284, opened 2026-07-23
config rules keys must be exact check names, with no glob support


maxonfjvipon:
In `.xslint.yml`, each `rules` key must be the exact name of a check. `src/xslint.js` folds `off` rules into the suppression list by exact name, looks up severity overrides with `config.rules[defect.name]`, and reports any key not in `CHECKS` as non-existent. So there is no way to write `template-match-*: off` to turn off a family of checks, even though the command-line `--suppress` matches on substrings — disabling a related group means listing every rule by hand.

Allow glob (or at least prefix) patterns in `rules` keys: expand each pattern against the known check names when building the disabled set and the severity-override map, and treat a pattern that matches nothing as the "unknown rule" warning already does.

This pairs with #258 — once the misleading `template-match-` prefix is gone, grouping rules under a meaningful pattern becomes far more useful.

Follow-up to #261.

maxonfjvipon:
Resolved by #286 (merged to master).

Issue #283, opened 2026-07-23
config file cannot set log-level or quiet, only max-warnings


maxonfjvipon:
The configuration file was meant to carry a project's defaults for the command-line flags, but only `max-warnings` is honoured. `normalized` in `src/config.js` reads `rules`, `exclude`, and `max-warnings` and nothing else, and `processOptions` in `src/xslint.js` reads `options.logLevel` and `options.quiet` straight from the CLI:

```js
const processOptions = function(options) {
  logger.setLevel(options.quiet ? levels.WARNING : options.logLevel)
}
```

So a team can check in its warning budget but not its log level or a default `--quiet`, which is an odd half-measure for a project config file.

Extend the schema with `log-level` and `quiet`, and resolve each the way `max-warnings` already is — `flag ?? config ?? default`. `processOptions` would then take the resolved values rather than reading the raw CLI options directly.

Follow-up to #261.

maxonfjvipon:
Resolved by #286 (merged to master).

PR #286, opened 2026-07-23
Complete .xslint.yml config: globs, log-level/quiet, exclude base, key validation

### LINKED ISSUE #282: config exclude globs resolve against the cwd, not the config file directory
When xslint finds `.xslint.yml` by walking up from the working directory, its `exclude` globs are matched against each file's path relative to the **working directory**, not the directory the config lives in. `excluded` in `src/xslint.js:98` computes `path.relative(process.cwd(), file)` and hands that to `minimatch`:

```js
const relative = path.relative(process.cwd(), file).split(path.sep).join('/')
return patterns.some((pattern) => minimatch(relative, pattern))
```

So a project-root `.xslint.yml` with `exclude: ["test/resources/**"]` stops matching the moment xslint runs from a subdirectory, even though the intent — ignore the project's test resources — has not changed. Every other config value is project-anchored, since the file is found by walking up, so anchoring `exclude` to the transient cwd is inconsistent and surprising.

Resolve the globs against the directory of the resolved config file instead: `src/config.js` already knows that path, so return it and compute the relative path from there in `excluded`, falling back to the cwd when there is no config file.

Follow-up to #261.

maxonfjvipon:
Rounds out the `.xslint.yml` support shipped in #261, closing the four follow-ups filed against it.

- **#282** — `exclude` globs now resolve against the config file's own directory, not the transient working directory, so they keep matching when xslint is run from a subdirectory. `src/config.js` returns a `base`, and `excluded` in `src/xslint.js` computes the relative path from it (falling back to the cwd when there is no file).
- **#283** — the config can now set `log-level` and `quiet`, not just `max-warnings`. Each resolves as `flag ?? config ?? default`; the `--log-level`/`--quiet` command-line defaults were dropped so the file's value can take effect when the flag is absent.
- **#284** — `rules` keys may be globs. Each key is expanded against the known check names (`template-match-*: off` disables the family); an exact name still matches exactly one check, and a pattern matching nothing is reported as before.
- **#285** — unrecognised top-level keys (a typo like `excludes:`) are reported instead of silently ignored, alongside the existing invalid-severity and unknown-rule warnings.

```yaml
rules:
  "template-match-unused-*": error   # glob
exclude:
  - "test/**"                        # relative to this file
log-level: info
quiet: false
```

Covered by unit specs (base directory, log-level parsing) and end-to-end tests (glob disable, unknown-key warning, config log-level, config quiet, and the reworked exclude test that anchors on the config directory). Coverage holds at 98%. README and CLAUDE.md are updated.

Closes #282, #283, #284, #285.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -105,7 +105,7 @@ Then: optionally add a rationale at `src/resources/motives/{xpath,corpus,validat
 
 Suppression by users: `xslint --suppress=<rule-substring>` (matches names from all validators and linters).
 
-Configuration by users: a `.xslint.yml` file (discovered by walking up from the working directory, or passed with `--config <path>`) can disable rules (`rules: {<exact-name>: off}`), re-grade severity (`warning`/`error`), skip files (`exclude:` globs), and set `max-warnings`. Command-line flags override the file; the file overrides the built-in defaults. Resolution lives in `src/config.js`; `src/xslint.js` folds `off` rules into the suppression list, filters excluded files, applies severity overrides to the collected defects, and resolves the effective `max-warnings`.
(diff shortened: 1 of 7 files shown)


maxonfjvipon:
@rultor merge

PR #287, opened 2026-07-23
Drop the template-match- rule-name prefix and add a conformance test

### LINKED ISSUE #258: Enforce rule-authoring conventions in CI, and drop the `template-match-` prefix
Nearly every rule under `src/resources/checks/xpath/` begins with `template-match-`, including many unrelated to template matching: `template-match-short-names`, `template-match-monolithic-design`, `template-match-unsorted-imports`. The prefix names nothing, and because `--suppress` matches on substrings (`src/xslint.js`), `xslint --suppress=template-match` silences the entire rule set at once — surprising for a user who wanted to mute a single check.

Renaming the existing rules to what they describe — `short-names`, `monolithic-design`, `unsorted-imports` — is only half the fix. Nothing stops the next rule from reintroducing the prefix, shipping without a motive, or landing with no test pack. Add a conformance test that runs under the existing mocha suite, so the `grunt` CI job on every push and pull request enforces it, asserting for each check that:

- its name matches the agreed convention — kebab-case, no meaningless prefix;
- it has at least one matching pack under `test/resources/xpath-packs/`, so no rule ships untested; and
- every pack and motive maps back to a real check, so a rename leaves no orphans behind.

If the project decides a rationale should be mandatory rather than optional, the same test can also require a matching `motives/xpath/<name>.md`.

The rename itself touches each YAML in `checks/xpath/`, its motive under `motives/xpath/`, and the matching pack in `test/resources/xpath-packs/`, plus a README note about the changed suppression strings and a regenerated docs site. With the conformance test in place, a future rule that misnames itself or omits its pack fails CI instead of slipping in.

maxonfjvipon:
The `template-match-` prefix on every rule name was pure noise — it named nothing, and because `--suppress` matches substrings, `--suppress=template-match` silently disabled the whole set. This drops it from all 46 `xpath` and `corpus` checks, and adds a conformance test so it cannot creep back.

**Breaking.** Rule names are public — they appear in `--suppress`, in `.xslint.yml` `rules` keys, and on the docs site — so `template-match-short-names` is now `short-names`, etc. Because `--suppress` is substring-based, `--suppress=short-names` already worked and keeps working; only full old names or the `--suppress=template-match` blanket break. The tool is pre-1.0, so this is a clean break with a CHANGELOG note rather than carried-forever aliases. No code depended on the prefix — names derive from filenames and matching is generic.

Four names were also rationalized, not just stripped:

| old | new |
|-----|-----|
| `are-you-confusing-variable-and-node` | `confusing-variable-and-node` |
| `can-use-abbreviated-axis-specifier` | `unabbreviated-axis` |
| `template-without-match-have-mode-or-priority` | `mode-or-priority-without-match` |
| `variable-or-param-have-select-and-internal-content` | `variable-or-param-with-select-and-content` |

The rename touches each check YAML, its motive, every pack's `pack:` value (and the 3-element `found.positions` cross-references), the pack filenames, and the hardcoded names in the README and `xslint.test.js`.

**Conformance test** (`test/conformance.test.js`) is the gate that keeps it clean. For every check it asserts a kebab-case name with no `template-match-` prefix, a motive (now mandatory), and — for `xpath`/`corpus` rules — at least one test pack; and it checks that no motive or pack is an orphan. A future rule that misnames itself, drops its motive, or ships without a pack fails the build.

305 tests pass, coverage holds at 98%, `grunt docs` regenerates cleanly. README, CLAUDE.md, and CHANGELOG are updated.

Closes #258.


maxonfjvipon:
@rultor merge

Issue #288, opened 2026-07-23
Report unused inline-suppression directives


maxonfjvipon:
Once inline `xslint-disable-*` directives exist (#262), some of them go stale: a directive that suppresses a rule which no longer fires at that spot silently lingers after the code around it changes, and nothing points it out. ESLint's `reportUnusedDisableDirectives` solves exactly this — it flags disable directives that suppressed nothing so they can be deleted.

xslint could do the same. The suppression pass already tests every defect against every directive, so it can mark each directive that actually matches a defect and, once the run is over, report the directives that matched none. A flag could decide whether an unused directive is a warning or a hard error.

Follow-up to #262.

PR #289, opened 2026-07-23
Add inline xslint-disable comment directives

### LINKED ISSUE #262: No way to suppress a rule on a single element or file
Suppression today is all-or-nothing across a run: `--suppress` drops a check everywhere it would ever fire (`src/xslint.js`). A general-purpose linter needs a local escape hatch for the one place a rule is knowingly and deliberately violated — the equivalent of `eslint-disable-next-line`. As it stands, a single justified exception forces a project to silence the rule for its entire codebase, which erodes the value of that rule everywhere else.

Support an XML-comment directive the linter honours, for example `<!-- xslint-disable template-match-short-names -->` on the line before the offending element, or `<!-- xslint-disable-file -->` at the top of a stylesheet. The linters already resolve each defect to a node with a line number, so a filtering pass that reads the comments surrounding that line and drops matching defects would implement this without touching any rule definition.

maxonfjvipon:
Suppression was all-or-nothing: `--suppress` and config `off` silence a rule everywhere, so a single justified exception forced a project to turn the rule off for its whole codebase. This adds a local escape hatch, the equivalent of `eslint-disable-next-line`, as XML-comment directives.

```xml
<!-- xslint-disable-next-line short-names -->
<xsl:variable name="x" select="1"/>

<!-- xslint-disable-file not-using-schema-types -->
```

Three forms — `xslint-disable-next-line`, `xslint-disable-line`, and `xslint-disable-file` — each taking an optional, space-separated list of rule names (none means every rule at that location).

A new `src/directives.js` scans the raw stylesheet text for the directives and their lines; `src/xslint.js` drops any collected defect a directive covers, matching on `(file, line, name)`. Because it filters the final defect list, it covers every kind — xpath, corpus, and format — and composes with `--suppress`/config `off`. A directive that names a rule which does not exist is reported, so typos do not pass silently.

Block ranges (`xslint-disable … xslint-enable`) are intentionally left out of this first cut, and reporting unused directives is filed separately as #288.

Covered by a `directives` unit spec (parsing each form and the suppression predicate) and end-to-end tests (next-line suppresses its target, leaves other defects, disable-file spans the file, unknown rule names warn). 315 tests pass, coverage holds at 98%. README and CLAUDE.md document the directives.

Closes #262.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add inline suppression directives — `xslint-disable-next-line`,
+  `xslint-disable-line`, and `xslint-disable-file` (#262).
 - **Breaking:** drop the `template-match-` prefix from every rule name and
   rationalize a few awkward ones, so `--suppress` strings and config `rules`
   keys change accordingly; add a conformance test that enforces rule naming,
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -107,7 +107,9 @@ Then: add a rationale (required) at `src/resources/motives/{xpath,corpus,validat
 
 Suppression by users: `xslint --suppress=<rule-substring>` (matches names from all validators and linters).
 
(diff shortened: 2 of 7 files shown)


maxonfjvipon:
@rultor merge

PR #291, opened 2026-07-23
Report unused inline-suppression directives

### LINKED ISSUE #288: Report unused inline-suppression directives

maxonfjvipon:
Inline `xslint-disable-*` directives (#262) go stale: after the code around one changes, a directive that no longer silences anything lingers unnoticed. This reports each such directive as unused, the way ESLint's `reportUnusedDisableDirectives` does.

`src/directives.js` gains a `covers(directive, defect)` predicate (which `suppresses` now uses) and an `unused(directives, defects)` finder that returns the directives matching no defect. `src/xslint.js` warns for each: `Unused xslint-disable directive at <file>:<line>`. It reports against the full defect list of the file, so a directive counts as used the moment it suppresses one defect.

The test fixtures follow the resource convention rather than string concatenation: `test/resources/directives/used.xsl` (a directive that suppresses a real `short-names` defect — no warning) and `unused.xsl` (a directive whose target has nothing to silence — warned). Both carry the SPDX header and are xcop-clean. Unit tests cover `unused` directly; the `should test default directory` count moves to 8 for the two new fixtures.

319 tests pass, coverage holds at 98%. README and CLAUDE.md document the behaviour.

Closes #288.


maxonfjvipon:
@rultor merge

PR #292, opened 2026-07-23
Move test stylesheet fixtures to resources; complain about concatenation in CI

### LINKED ISSUE #290: Replace string-concatenated test stylesheets with .xsl or pack resources
Several tests build their stylesheet fixtures by concatenating lines in JavaScript — `[...].join('\n')` written to a temp file — which reads poorly and sits awkwardly next to the project's committed `.xsl` fixtures and the pack `input:` block scalars. `test/xslint.test.js` does this in eight places: the malformed-stylesheet pair, the error-exit fixture (#264), the exclude fixture (#261), and the four inline-suppression fixtures (#262).

Replace each multi-line fixture with a proper resource — a committed `.xsl` under `test/resources/` linted by path, or a pack YAML when a check is under test — following the approach used for the #288 directive fixtures under `test/resources/directives/`. Leave the trivial one-to-five-line snippets inline; the point is the multi-line ones.

Two things to mind, learned from #288: a committed `.xsl` must carry the column-zero `* SPDX` header and be xcop-clean (the `**/*.xsl` workflow checks it), and the `should test default directory` test counts every `.xsl` in the repo, so each new fixture bumps its `Processed files: N`.

Follow-up to #262.

maxonfjvipon:
Several tests built their stylesheet fixtures by concatenating lines in JavaScript — `[...].join('\n')` — which reads poorly next to the project's committed `.xsl` fixtures and pack `input:` blocks. This moves them to proper resources and adds a CI nudge so new ones do not creep in.

The well-formed multi-line inline-suppression fixtures become committed resources under `test/resources/directives/` and are linted by path: the next-line test reuses `used.xsl`, and two new fixtures, `targeted.xsl` and `disable-file.xsl`, cover the remaining cases. Both carry the SPDX header and are xcop-clean.

Two kinds of fixture stay inline, now as template literals rather than `join` concatenation: malformed stylesheets (the `should lint the parseable stylesheets…` and error-exit tests), which xcop cannot accept as committed files, and trivial one-to-five-line snippets (the exclude and unknown-rule tests). No test now uses `[...].join('\n')`.

A new `fixtures` CI job greps the test files for the concatenation pattern and, if it finds one, emits a warning annotation — a complaint, not a failure, so the convention is visible without being brittle. CLAUDE.md documents the rule and its two exceptions.

The `should test default directory` count moves to 10 for the two new committed fixtures. 319 tests pass, coverage holds at 98%.

Closes #290.

### DIFF
--- a/.github/workflows/fixtures.yml
+++ b/.github/workflows/fixtures.yml
@@ -1,0 +1,23 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+name: fixtures
+'on':
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+jobs:
+  fixtures:
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Complain about concatenated stylesheet fixtures
+        run: |
+          matches=$(grep -rn "]\.join(" test/ --include='*.test.js' || true)
+          if [ -n "${matches}" ]; then
+            echo "${matches}"
+            echo "::warning::Build test stylesheet fixtures from .xsl resources or the pack input: block, not string concatenation"
+          fi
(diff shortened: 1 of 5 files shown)


maxonfjvipon:
@rultor merge

PR #295, opened 2026-07-24
Move every test stylesheet to a resource; fail CI on inline fixtures

### LINKED ISSUE #294: Use committed resources for every test stylesheet, and fail CI on inline ones
#290 moved the multi-line stylesheet fixtures to `.xsl` resources but left the trivial and malformed ones inline as template literals. Those read no better than the `join` concatenation they replaced. Every stylesheet a test uses — well-formed or malformed, one line or twenty — should live in a committed resource, not in the test source.

The remaining inline fixtures in `test/xslint.test.js` are the parseable-and-malformed pair, the error-exit fixture, the exclude fixture, and the unknown-rule fixture. Move each out:

- Well-formed fixtures become committed `.xsl` files under `test/resources/`, linted by path (and validated by the xcop workflow). The exclude case gets its own committed directory with a `.xslint.yml` and a sheet.
- Malformed fixtures become committed YAML sources (a `content:` block scalar), read and written to a temp file at test time — a committed malformed `.xsl` cannot pass the `**/*.xsl` xcop workflow, so it stays out of that glob.

Then make the `fixtures` CI job a hard failure rather than a warning: if any `.test.js` file contains an inline `<?xml` or `<xsl:`, the build fails. And update CLAUDE.md so the convention reads "no inline stylesheets in tests, ever," with the malformed-as-YAML exception spelled out.

Supersedes the "except the easiest one-to-five-line snippets" carve-out from #290.

maxonfjvipon:
#290 moved the multi-line fixtures to `.xsl` resources but left the trivial and malformed ones inline as template literals, which read no better than the `join` they replaced. This moves every remaining test stylesheet into a committed resource and turns the `fixtures` job into a hard gate.

- The parseable-and-malformed pair and the error-exit fixture now read their content from committed YAML sources (`test/resources/malformed/{invalid-xpath,malformed}.yaml`, a `content:` block scalar) and write it to a temp file. A committed malformed `.xsl` cannot pass the `**/*.xsl` xcop workflow, so those stay out of that glob as YAML.
- The exclude case moves to a committed directory — `test/resources/excluded/` with a `.xslint.yml` and `sheet.xsl` — linted by path.
- The unknown-rule case moves to `test/resources/directives/unknown-rule.xsl`, linted by path.

No test source now contains an inline stylesheet. The `fixtures` CI job, previously a warning about `join`, now fails the build if any `.test.js` file contains an inline `<?xml` or `<xsl:`. CLAUDE.md documents the rule and the malformed-as-YAML exception.

The two new committed `.xsl` are xcop-clean with SPDX headers; the `should test default directory` count moves to 12. 319 tests pass, coverage holds at 98%.

Closes #294.


maxonfjvipon:
@rultor merge

PR #297, opened 2026-07-24
Add --format json and sarif output behind a uniform reporter

### LINKED ISSUE #260: No machine-readable output format (SARIF or JSON)
`xslint` emits only human-oriented text. `src/xslint.js:135` formats each defect through `stdout` as `file(line:pos) message (name)`, and there is no flag to produce anything else. For a general-purpose linter this closes off the surrounding tooling ecosystem: GitHub code scanning ingests SARIF, CI systems annotate diffs from SARIF or JSON, and editors read structured diagnostics — none of which can consume the current output.

Add a `--format` option offering at least `sarif` and `json` alongside the default `text`. The defect objects already carry everything the formats need — `name`, `severity`, `message`, `file`, `line`, `pos` — so a JSON emitter is a direct serialization of the defect list, and a SARIF emitter maps each rule name to a `reportingDescriptor` and each defect to a `result` with a `physicalLocation`. This is the single change that unlocks CI and editor integration, and it is a prerequisite for treating xslint as a general-purpose tool rather than a terminal-only one.

maxonfjvipon:
xslint only spoke human text, which closed off the tooling ecosystem — GitHub code scanning, CI annotations, editors. This adds `--format <text|json|sarif>` (default `text`, so nothing changes unless asked).

- **json** — a flat array of `{rule, severity, message, file, line, column}`.
- **sarif** — a SARIF 2.1.0 log; `runs[0].results[]` carry `ruleId`, `level`, `message`, and a `physicalLocation`, and `tool.driver.rules[]` is derived from the rules present, so the log is self-contained. GitHub code scanning ingests it (README shows the `upload-sarif` snippet).

Both machine formats print one document to stdout — clean to pipe, because #263 already put logs on stderr — and use cwd-relative posix paths, what code scanning expects. The exit code stays severity-aware regardless of format.

Per the request, output goes through a single interface: `src/reporters.js` exposes `reporterOf(format)`, and `src/xslint.js` just calls `reporterOf(options.format)(reported)` without knowing which format it is. The three reporters (`text`, `json`, `sarif`) share the `(defects) => void` shape; `text` is the former inline loop moved behind the same interface. An unknown `--format` fails fast via commander's `choices`.

Tested both ways: a `reporters` unit spec drives each reporter and asserts the JSON object shape, the empty-array case, and the SARIF version, result location, derived rule, and error-level mapping; end-to-end tests run the CLI with each `--format` and parse stdout. No new dependency (SARIF is a plain object). 328 tests pass, coverage holds at 98%.

Closes #260.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add machine-readable output: `--format json` and `--format sarif` (SARIF
+  2.1.0, for GitHub code scanning) alongside the default `text` (#260).
 - Add inline suppression directives — `xslint-disable-next-line`,
   `xslint-disable-line`, and `xslint-disable-file` (#262), and report a
   directive that suppresses nothing as unused (#288).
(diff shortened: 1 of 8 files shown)


maxonfjvipon:
@rultor merge

PR #306, opened 2026-07-24
Validate .xslint.yml value types

### LINKED ISSUE #301: .xslint.yml silently accepts wrongly-typed `max-warnings`, `exclude`, and `quiet`
`normalized` in `src/config.js:59` warns about unknown top-level keys and about a rule graded to an unknown severity, but it never checks the *type* of the known keys. `exclude` that is not an array is silently replaced with `[]`, and `max-warnings`, `quiet`, and `log-level` are passed through untouched whatever their type:

```js
exclude: raw && Array.isArray(raw.exclude) ? raw.exclude : [],
maxWarnings: raw && Object.hasOwn(raw, 'max-warnings') ?
  raw['max-warnings'] : null,
```

A non-numeric `max-warnings` then disables the warning gate without a word: `src/xslint.js:213` guards on `maxWarnings >= 0`, and `"abc" >= 0` is `false`, so the run stops failing on warnings and the user is never told the value was ignored.

Validate each known key's type the way severities and unknown keys are already validated — warn and fall back to the default when `max-warnings` is not a number, `exclude` is not an array of strings, or `quiet` is not a boolean — so a typo in a value is as visible as a typo in a key.

maxonfjvipon:
`normalized` in `src/config.js` warned about unknown keys and unknown severities but never checked the type of a known key's value. A non-numeric `max-warnings` slipped through and disabled the warning gate without a word, because `src/xslint.js` guards on `maxWarnings >= 0` and `"abc" >= 0` is `false`; a numeric `log-level` reached `setLevel`, which calls `.toLowerCase()` on it.

A small `typed` helper now validates each known key, warning and falling back to the default when the value has the wrong type — the same visible treatment a mistyped key already got.

```js
maxWarnings: typed(
  raw, 'max-warnings',
  (val) => typeof val === 'number' && !Number.isNaN(val), 'a number', null,
),
```

Closes #301.


maxonfjvipon:
@rultor merge

PR #307, opened 2026-07-24
Color output only for an interactive terminal

### LINKED ISSUE #302: `text` reporter writes ANSI color codes into redirected, non-TTY output
The default `text` reporter colors every line through `src/output.js`, which wraps messages with `colors/safe`:

```js
const safe = require('colors/safe')
...
error: (msg, ...args) => sink(`${safe.red('[ERROR]')} ${msg}`, ...args),
```

`colors/safe` emits escape sequences unconditionally — it has no terminal check — so `xslint path > report.txt` writes raw `\x1b[31m…\x1b[39m` bytes into the file, and the same happens through any pipe. Redirected diagnostics are meant to be plain text, and the escape codes make the captured output awkward to read, grep, or diff.

Suppress coloring when the destination stream is not a TTY (guard on `process.stdout.isTTY` for defects and `process.stderr.isTTY` for logs), and honor the conventional `NO_COLOR` environment variable, so interactive runs stay colored while captured output stays clean.

maxonfjvipon:
The default `text` reporter colors every line through `colors/safe`, which emits escape sequences with no terminal check. So `xslint path > report.txt` wrote raw `\x1b[31m…\x1b[39m` bytes into the file, and any pipe carried the same noise.

`writer` now takes an explicit `colored` flag and paints the prefix only when it is set. The two module-level writers derive the flag from their stream — colored only when `isTTY` and `NO_COLOR` is unset — so interactive runs stay colored while redirected or piped output is plain.

```js
const colorful = function(stream) {
  return Boolean(stream.isTTY) && !Object.hasOwn(process.env, 'NO_COLOR')
}
```

A small `output` unit test covers both branches without faking a terminal.

Closes #302.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Color output only for an interactive terminal and honor `NO_COLOR`, so
+  redirected or piped output no longer carries raw ANSI escapes (#302).
 - Report and ignore `.xslint.yml` values of the wrong type — a non-numeric
   `max-warnings` no longer silently disables the warning gate (#301).
 - Add machine-readable output: `--format json` and `--format sarif` (SARIF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,10 @@ drop the informational log lines:
 xslint --quiet
 ```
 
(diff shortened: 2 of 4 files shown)


maxonfjvipon:
@rultor merge

PR #308, opened 2026-07-24
Set the log level before reading config and suppressions

### LINKED ISSUE #299: `--log-level` and `--quiet` do not silence warnings emitted before the level is set
`xslint` in `src/xslint.js:129` resolves the configuration, expands the rule patterns (`:136`), and validates the suppressions — each able to call `logger.warn` — before `processOptions` (`:148`) reads `--log-level`/`--quiet` and calls `logger.setLevel`. Every one of those early warnings prints at the default level regardless of what the user asked for.

```js
const config = configFrom(options.config)          // may warn: unknown key, bad severity
...
if (matched.length === 0) {
  logger.warn(`Rule '${pattern}' in configuration does not exist`)
}
const suppressions = [...validatedSuppressions(options.suppress), ...disabled]  // may warn
const maxWarnings = options.maxWarnings ?? config.maxWarnings ?? -1
processOptions(options, config)                     // level is set only here
```

So `xslint sheet.xsl --log-level error --suppress=nonexistent` still prints `[WARNING] Check with substring 'nonexistent' does not exist`, and an unknown rule name in `.xslint.yml` still warns under `--log-level error`.

Resolve `--quiet`/`--log-level` from the command line first and call `setLevel` before reading the config, the rule patterns, and the suppressions. The config file's own `log-level`/`quiet` can only take effect once the file is read, but the command-line flags are known up front and should win from the first line of output.

maxonfjvipon:
`configFrom`, the rule-pattern expansion, and `validatedSuppressions` each call `logger.warn`, but the level was applied only afterwards. So `xslint sheet.xsl --log-level error --suppress=nonexistent` still printed the bad-suppress warning, and an unknown rule name in `.xslint.yml` warned under `--log-level error`.

The command-line level is now set first, so every warning that follows obeys it. Once the config is read, it fills the level in only when the command line said nothing — the command line keeps precedence.

```js
logger.setLevel(leveled(options.quiet, options.logLevel))
const config = configFrom(options.config)
if (options.quiet == null && options.logLevel == null) {
  logger.setLevel(leveled(config.quiet, config.logLevel))
}
```

The old `processOptions` — which only set the level and was named for more — gives way to a small `leveled` helper.

Closes #299.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Apply `--log-level`/`--quiet` before reading the configuration, rule
+  patterns, and suppressions, so a raised level now silences their warnings
+  too (#299).
 - Color output only for an interactive terminal and honor `NO_COLOR`, so
   redirected or piped output no longer carries raw ANSI escapes (#302).
 - Report and ignore `.xslint.yml` values of the wrong type — a non-numeric
--- a/src/xslint.js
+++ b/src/xslint.js
@@ -102,16 +102,14 @@ const excluded = function(file, patterns, base) {
 }
 
 /**
- * Set the log level, taking the command line first, then the configuration,
- * then the default.
- * @param {{logLevel: string|undefined, quiet: boolean|undefined}} options - CLI
- *  options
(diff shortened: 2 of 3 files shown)


maxonfjvipon:
@rultor merge

PR #309, opened 2026-07-24
Use bare require names and enforce it; collapse warn naming

### LINKED ISSUE #304: Inconsistent `node:path` imports and duplicated log-level naming in `src/`
Two small consistency snags in `src/`.

First, the `path` module is imported two ways: `require('node:path')` in seven files (`config.js`, `reporters.js`, `xsl-validator.js`, `xpath-validator.js`, `xpath-linter.js`, `corpus-linter.js`, `xpath-format-linter.js`) and `require('path')` in `helpers.js:7` and `xslint.js:6`.

Second, the logging vocabulary carries one level under two names. `src/logger.js:12` defines both `WARN: 'warn'` and `WARNING: 'warning'` mapped to the same number, and `src/output.js:26` exposes both `warn` and `warning` where one just delegates to the other:

```js
warn: (msg, ...args) => sink(`${safe.yellow('[WARNING]')} ${msg}`, ...args),
warning: (msg, ...args) => write.warn(msg, ...args),
```

And `processOptions` in `src/xslint.js:111` is named for a generality it does not have — it only sets the log level.

None of this is a bug, but it is the kind of drift a linter's own source should not carry. Settle on `node:path` everywhere, collapse the duplicate warning name to one, and rename `processOptions` to say what it does.

maxonfjvipon:
The `path` module was required two ways across `src/` — seven files with the `node:` prefix, two without — and `scripts/generate-docs.js` added two more prefixed requires. This settles on the bare name everywhere and, more to the point, makes the convention stick: a new `no-restricted-syntax` selector fails CI on any prefixed require, so the drift cannot return.

```js
{
  selector:
    "CallExpression[callee.name='require'][arguments.0.value=/^node:/]",
  message: "Do not use the 'node:' prefix in require; use the bare name"
}
```

The same commit collapses the duplicate warning name: `LOG_LEVELS` and the output writer each carried both `warn` and `warning` for one level, and now keep only `warning` (the logger's public `warn` verb is unchanged). `processOptions` was already retired in #299.

`CLAUDE.md` gains the rule and the principle behind it — a fixed code-style convention must be machine-enforced, in the same change, so the mistake cannot recur.

Closes #304.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Use bare module names in `require` (drop the `node:` prefix) and enforce it
+  with an ESLint rule; collapse the duplicate `warn`/`warning` naming in the
+  logger and writer (#304).
 - Apply `--log-level`/`--quiet` before reading the configuration, rule
   patterns, and suppressions, so a raised level now silences their warnings
   too (#299).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,9 @@ npx mocha test/xslint.test.js --timeout 10000   # Run a single test file
 npx mocha test/xslint.test.js --grep "sentence"  # Run tests matching a pattern
 ```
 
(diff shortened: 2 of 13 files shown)


maxonfjvipon:
@rultor merge

PR #311, opened 2026-07-24
Assert behavior, not whole-corpus totals, in the e2e tests

### LINKED ISSUE #303: `xslint.test.js` asserts on absolute file and defect counts that any new fixture breaks
Several end-to-end assertions in `test/xslint.test.js` pin exact totals over the whole corpus:

```js
// :103
assert.ok(stdout.includes('Processed files: 14'))
// :31-32
'Processed files: 1',
'Defects found: 7',
```

`Processed files: 14` counts every committed `.xsl` under the default scan, and `Defects found: 7` counts every defect a single fixture happens to trigger. Adding one fixture, or adding or re-grading one rule, breaks a test that has nothing to do with the change — a maintenance tax paid on unrelated work, and a habit that trains contributors to bump a magic number rather than reason about behavior.

Assert on the property each test is about: that a known fixture is in the processed set, that a specific expected defect is present or absent. A dedicated fixture directory the default-scan test owns, rather than the shared corpus, keeps its count stable while the rest of the suite grows.

maxonfjvipon:
The end-to-end tests pinned totals over the whole corpus and rule set: the default-scan test asserted `Processed files: 14`, and two tests asserted `Defects found: 7`/`4`. Adding one fixture, or adding or re-grading one rule, broke a test that had nothing to do with the change.

Those totals are dropped. The per-defect assertions already present — each checking a specific `(line:col) message (rule)` — cover the behavior far better than a count, and the default-scan test now asserts only that it announces `.` and processes a positive number of files:

```js
assert.ok(/Processed files: [1-9]\d*/.test(stdout))
```

Verified that adding a stray `.xsl` no longer breaks the default-scan test. The `CLAUDE.md` note describing the old coupling is updated.

Closes #303.


maxonfjvipon:
@rultor merge

PR #312, opened 2026-07-24
Treat every reported XML problem as malformed and quiet the parser

### LINKED ISSUE #298: `xsl-validator` accepts not-well-formed stylesheets that `@xmldom/xmldom` only warns about
`validate` in `src/xsl-validator.js:46` treats a stylesheet as well-formed whenever `xml.parsedFromString` (from `src/helpers.js:58`) does not throw. But `@xmldom/xmldom` 0.9 splits parse problems into two classes: fatal errors (tag mismatch, unclosed tag, missing root) throw, while recoverable errors — an undefined entity, a stray `<` — are handed to the parser's error handler and a document is still returned.

```js
const parser = new DOMParser()
...
parsed = parser.parseFromString(str, 'text/xml')
```

No `onError` callback is passed to the `DOMParser`, so two things follow. A genuinely not-well-formed stylesheet slips into the corpus unreported: `<out>&undefined; text</out>` is linted as if valid and never raises `malformed-stylesheet`. And xmldom's default handler prints its own diagnostics straight to the console, bypassing `src/logger.js` and surviving `--quiet`:

```
[xmldom error]    entity not found:&undefined; @#[line:4,col:...]
[xmldom fatalError]    Opening and ending tag mismatch: "foo" != "bar" @#[...]
```

Pass an `onError` handler to the `DOMParser` that collects the diagnostics, and treat any reported problem — not only the ones that throw — as a parse failure, so the corpus holds only well-formed stylesheets and the noise flows through the logger. A malformed fixture with an undefined entity in element text locks the regression down.

maxonfjvipon:
`@xmldom/xmldom` splits parse problems in two: it throws on a fatal error (a tag mismatch, an unclosed tag) but only reports a recoverable one (an undefined entity, a stray `<`) to its error handler and returns a document anyway. The parser was built with no handler, so it relied on the throw alone — a not-well-formed stylesheet slipped into the corpus unreported and was linted as valid, while xmldom's default handler spilled `[xmldom error]` lines onto the console, past the logger and past `--quiet`.

`xmlFromString` now parses with an `onError` handler that collects any non-warning problem and fails when one is present, so both classes reach `xsl-validator` as a `malformed-stylesheet` defect and the parser's diagnostics stay off the console.

```js
const parsed = new DOMParser({
  onError: (level, message) => {
    if (level !== 'warning') {
      problems.push(message.trim())
    }
  },
}).parseFromString(str, 'text/xml')
if (problems.length > 0) {
  throw new Error(`Couldn't parse XML:\n${str}\n\nCause: ${problems[0]}`)
}
```

Two `disable-output-escaping` fixture packs carried a raw `<` as `xsl:text` content — not well-formed XML that xmldom's recovery had masked. They are escaped to `&lt;`, the same character, now well-formed. New validator tests cover an undefined entity and the absence of console noise.

Closes #298.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Report a stylesheet as malformed for any well-formedness problem the XML
+  parser reports, not only the fatal ones — an undefined entity or a stray
+  `<` no longer slips through — and route the parser's own diagnostics through
+  the logger instead of leaking them to the console (#298).
 - Assert behavior rather than whole-corpus totals in the end-to-end tests, so
   adding a fixture or a rule no longer breaks an unrelated test (#303).
 - Use bare module names in `require` (drop the `node:` prefix) and enforce it
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -9,12 +9,6 @@ const {DOMParser} = require('@xmldom/xmldom')
 const yaml = require('yaml')
 
 /**
- * XML Parser
- * @type {DOMParser}
- */
-const parser = new DOMParser()
-
-/**
  * Get all the files recursively from given directory
  * @param {string} dir - Directory path
  * @return {Array.<string>} - Array of file in given directory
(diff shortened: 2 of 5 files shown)


maxonfjvipon:
@rultor merge

PR #313, opened 2026-07-24
Reuse one DOMParser instead of building it per parse

maxonfjvipon:
A small follow-up to #298. That change built a fresh `DOMParser` on every call and collected the parser's problems in a mutable array before deciding whether to throw. Neither is needed: the parser can be built once and its `onError` can raise directly on any non-warning problem, so the throw propagates out of `parseFromString` on its own.

```js
const parser = new DOMParser({
  onError: (level, message) => {
    if (level !== 'warning') {
      throw new Error(message.trim())
    }
  },
})
```

One reused instance now replaces the per-call construction and the collection array. Behavior is unchanged — the validator tests added in #298 (an undefined entity is reported, no diagnostics leak to the console) still pass, and a reused parser stays correct across interleaved well-formed and malformed inputs.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -9,6 +9,22 @@ const {DOMParser} = require('@xmldom/xmldom')
 const yaml = require('yaml')
 
 /**
+ * XML parser, built once and reused. Its error handler raises on any
+ * well-formedness problem the parser reports — not only the fatal ones it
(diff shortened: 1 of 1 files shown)


maxonfjvipon:
@rultor merge

PR #314, opened 2026-07-24
Load the CLI entry as an ES module to drop the commander warning

### LINKED ISSUE #300: Test harness masks the `commander` ESM `ExperimentalWarning` that end users see
`commander` is loaded with `require('commander')` in `src/index.js:7`, but the installed `commander` (15.x) ships as ESM only — `"type": "module"` with no CommonJS export. Loading an ES module through `require()` makes Node print an `ExperimentalWarning` on stderr across the supported range (`engines: ">=20"`).

```js
const {program, Option} = require('commander')
```

The test helper hides this rather than removing it:

```js
// test/helpers.js:42
env: {...process.env, NODE_NO_WARNINGS: '1'},
```

So the suite runs clean while a user invoking the CLI directly sees a warning the tests never surface — the harness observes something different from what it ships.

Load commander in a way that does not trip the CJS-to-ESM bridge (pin a version that still ships a CommonJS entry, or move the entry point to ESM), and drop `NODE_NO_WARNINGS` from `test/helpers.js` so the tests observe exactly what users observe.

maxonfjvipon:
`commander` (15.x) ships as an ES module with no CommonJS entry, so `require('commander')` from the CommonJS CLI made Node print an `ExperimentalWarning` on every run. The test harness hid it with `NODE_NO_WARNINGS=1`, so the suite stayed quiet while a user saw the warning the tests never did.

The entry becomes an ES module, `src/index.mjs`, importing commander natively — no require-of-ESM, no warning:

```js
import {program, Option} from 'commander'
import version from './version.js'
import xslint from './xslint.js'
```

`NODE_NO_WARNINGS` is dropped from the test helper, so a spawned run is exactly a user's run. `bin`, `main`, the Stryker mutate list, and the Grunt/ESLint globs now point at the `.mjs` entry, and a new ESLint block lints it as a module so it stays covered — verified that a planted error in `index.mjs` now fails `grunt eslint`.

Closes #300.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Load the CLI entry as an ES module (`src/index.mjs`) so `commander` — which
+  is ESM-only — no longer triggers Node's `ExperimentalWarning` on every run;
+  the test harness no longer silences warnings, so it sees what users see
+  (#300).
 - Report a stylesheet as malformed for any well-formedness problem the XML
   parser reports, not only the fatal ones — an undefined entity or a stray
   `<` no longer slips through — and route the parser's own diagnostics through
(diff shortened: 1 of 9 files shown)


maxonfjvipon:
@rultor merge

PR #315, opened 2026-07-24
Ban a redundant return variable with a local ESLint rule

### LINKED ISSUE #310: Add a code-style check that prohibits a redundant return variable
The project's style avoids unnecessary local variables — a variable that exists only to be handed straight back should be inlined. #304 removed one such variable (`write` in `src/output.js`, assigned and then immediately returned), but nothing stops the next one from appearing.

```js
const write = { ... }
return write        // redundant — the object literal can be returned directly
```

Per the convention now recorded in `CLAUDE.md` — a fixed code-style rule must be machine-enforced so it cannot recur — this pattern deserves its own check. The target is precise and has few false positives: a `const`/`let` whose sole purpose is to be returned by the very next statement (`const x = <expr>; return x`), which should be written `return <expr>`.

A dependency-free way to add it is a small local ESLint rule wired into `eslint.config.mjs` (flat config supports inline `plugins`), reported as an error so CI fails on a redundant return variable. A single `no-restricted-syntax` selector cannot express "the next statement returns this binding," so a tiny custom rule — or an established plugin rule such as sonarjs's `prefer-immediate-return` — is the right tool.

The codebase has no current instances, so this is prevention: the check locks in the convention for new code.

maxonfjvipon:
A variable that exists only to be handed back by the very next statement — `const x = expr; return x` — should be inlined to `return expr`. #304 removed one such variable by hand; this makes the convention stick.

A single `no-restricted-syntax` selector cannot express it, because esquery cannot compare the declaration's name with the identifier the following `return` uses. So the check is a small project-local ESLint rule, wired into `eslint.config.mjs` with no new dependency:

```js
ReturnStatement(node) {
  const prev = block.body[block.body.indexOf(node) - 1];
  if (prev?.type === "VariableDeclaration" &&
      prev.declarations.length === 1 &&
      prev.declarations[0].id.name === node.argument.name) {
    context.report({ node: prev, messageId: "redundant" });
  }
}
```

No current file violates it — an AST scan across `src`, `test`, and `scripts` found zero instances — so this is prevention, not cleanup. `CLAUDE.md` gains the rule alongside the other enforced conventions.

Closes #310.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add a project-local ESLint rule that bans a redundant return variable
+  (`const x = expr; return x`), so the convention is enforced on new code
+  (#310).
 - Load the CLI entry as an ES module (`src/index.mjs`) so `commander` — which
   is ESM-only — no longer triggers Node's `ExperimentalWarning` on every run;
   the test harness no longer silences warnings, so it sees what users see
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -19,7 +19,7 @@ npx mocha test/xslint.test.js --timeout 10000   # Run a single test file
 npx mocha test/xslint.test.js --grep "sentence"  # Run tests matching a pattern
 ```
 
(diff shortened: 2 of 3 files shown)


maxonfjvipon:
@rultor merge

PR #316, opened 2026-07-24
Reach 100% coverage and raise the gate to 100

### LINKED ISSUE #305: Reach 100% statement and branch coverage
Coverage sits just under the ceiling. A handful of branches are trusted by eye rather than exercised by a test:

```
logger.js  | 87.2  | uncovered 61-63, 68-75
index.js   | 91.83 | uncovered 46-49
helpers.js | 93.54 | uncovered 44-45, 47-48, 78-79
tokens.js  | 99.18 | uncovered 378-382
xslint.js  | 99.08 | uncovered 136-137
```

These are `logger.setLevel`'s invalid-level branch and its error path, the top-level `catch` in `src/index.js`, the existence and directory guards in `fromFile` (`src/helpers.js`), a scientific-notation edge in number scanning (`src/tokens.js`), and the unknown-rule branch in `src/xslint.js`.

Add tests for each so the `.c8rc.json` threshold can be raised from 90 to 100 and every branch is checked rather than assumed. Some of these — the malformed-config and unknown-rule branches — overlap the behavior other tickets in this batch already add tests for, so this one is best finished last.

maxonfjvipon:
Coverage was in the high 90s with the gate at 90%. This closes the remaining gaps and raises the `.c8rc.json` thresholds to 100 for statements, branches, functions, and lines.

The untested paths were exercised with focused tests:

- `helpers` — the `fromFile` guards (a missing file, a directory) and a YAML string that does not parse.
- `logger` — the debug, error, and unknown-level paths; its `error` method dropped a dead level guard (error is the highest level, so the check was always true).
- `tokens` — the two malformed-exponent number edges (`1e`, `1e+`).
- `xslint` — the unknown-rule-in-config warning.
- `output` — `colorful` for a non-terminal, and a terminal with and without `NO_COLOR`.
- `xpath` — both `namespaceResolver` branches (a standard `xsl:` prefix and a non-standard one); fontoxpath resolves `xs`/`fn`/`math` natively, so only `xsl:` reaches the standard-prefix branch.

The CLI bootstrap `src/index.mjs` is excluded from coverage, as it already is from mutation testing — its only untested line is the top-level `catch`, which commander does not reach without `exitOverride`.

Closes #305.

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -2,13 +2,16 @@
   "include": [
     "src/**"
   ],
+  "exclude": [
+    "src/index.mjs"
+  ],
   "reporter": [
     "text-summary",
     "lcov"
   ],
   "check-coverage": true,
-  "statements": 90,
-  "lines": 90,
-  "functions": 90,
-  "branches": 90
+  "statements": 100,
+  "lines": 100,
+  "functions": 100,
+  "branches": 100
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Reach 100% statement, branch, function, and line coverage and raise the
+  gate from 90% to 100% (the CLI bootstrap `src/index.mjs` is excluded, as it
+  already is from mutation testing) (#305).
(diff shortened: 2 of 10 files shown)


maxonfjvipon:
@rultor merge

PR #321, opened 2026-07-24
Send the CLI error stack to stderr, not stdout

### LINKED ISSUE #318: A CLI error writes its stack trace to stdout instead of stderr in `src/index.mjs`
The top-level error handler in `src/index.mjs:43` splits one error across two streams, and sends half of it to the wrong one:

```js
} catch (error) {
  console.error(error.message)
  console.debug(error.stack)
  process.exit(1)
}
```

`console.error` writes to stderr, but `console.debug` is an alias for `console.log` and writes to **stdout**. So a failure prints its message to stderr and its stack trace to stdout — the stream that must stay clean, since it carries the defects and, under `--format json`/`sarif`, a single machine-readable document. A parse error would corrupt that document with a stack trace.

The catch is rarely reached (commander exits rather than throws for ordinary CLI errors, which is why `src/index.mjs` is excluded from coverage), but the stream is simply wrong. Send the stack to stderr with `console.error(error.stack)` so all diagnostics stay off stdout.

maxonfjvipon:
`console.debug` is an alias for `console.log` and writes to stdout, so the top-level catch in `src/index.mjs` sent a fatal error message to stderr but its stack trace to stdout — the stream that carries defects and, under `--format json`/`sarif`, a single machine-readable document.

```js
catch (error) {
  console.error(error.message)
  console.error(error.stack)   // was console.debug -> stdout
  process.exit(1)
}
```

A regression test reaches the catch through `--config <directory>` (an unreadable config makes `configFrom` throw) and asserts stdout stays empty.

Closes #318.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Send a fatal CLI error's stack trace to stderr instead of stdout, so stdout
+  stays clean for defects and machine-readable output (#318).
+
 - Reach 100% statement, branch, function, and line coverage and raise the
   gate from 90% to 100% (the CLI bootstrap `src/index.mjs` is excluded, as it
   already is from mutation testing) (#305).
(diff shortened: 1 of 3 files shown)


maxonfjvipon:
@rultor merge

PR #323, opened 2026-07-24
Parse corpus checks once at module load

### LINKED ISSUE #317: `corpus-linter` re-parses every check's YAML on each run instead of once at load
`lintByCorpus` in `src/corpus-linter.js:46` re-reads and re-parses each check's YAML from disk every time it runs. The `CHECKS` array built at module load carries only `{name, path}`, and the parse happens inside the loop:

```js
for (const check of CHECKS) {
  if (suppressions.some((sup) => check.name.includes(sup))) {
    continue
  }
  const yml = yaml.parsedFromFile(check.path)
  ...
}
```

Every other check-owning module parses its YAML once, at module load: `src/xsl-validator.js:20`, `src/xpath-validator.js:21`, and `src/xpath-format-linter.js:21` resolve a `META` up front, and `src/xpath-linter.js:23` maps its packs to fully-parsed objects with `...yaml.parsedFromFile(pack)` — the shape #256 introduced. `corpus-linter` is the only one that re-parses per run.

The rule bodies never change during a run, so the read is pure waste, and the inconsistency invites the next contributor to copy the wrong pattern. Fold the parse into the `.map` at load so `CHECKS` carries `{name, declaration, usage, severity, message}`, and read from that in the loop, exactly as `xpath-linter` does. The behaviour is unchanged; only the redundant I/O disappears.

maxonfjvipon:
`lintByCorpus` re-read and re-parsed each check""s YAML from disk on every run, while every other check-owning module (`xsl-validator`, `xpath-validator`, `xpath-linter`, `xpath-format-linter`) parses once at module load.

The parse now folds into the `CHECKS` map at load, so each check carries its declaration/usage selectors and defect metadata, and the loop reads from that:

```js
const CHECKS = allFilesFrom(...).map((check) => ({
  name: check.substring(...),
  ...yaml.parsedFromFile(check),
}))
```

Behavior is unchanged — the corpus-pack tests still pass — only the redundant per-run I/O disappears.

Closes #317.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Parse the corpus checks once at module load instead of on every run, matching
+  the other linters and validators (#317).
+
 - Send a fatal CLI error's stack trace to stderr instead of stdout, so stdout
   stays clean for defects and machine-readable output (#318).
 
--- a/src/corpus-linter.js
+++ b/src/corpus-linter.js
@@ -9,8 +9,10 @@ const path = require('path')
 const {logger} = require('./logger')
 
 /**
- * Corpus checks, each identified by the name suppressions match against.
- * @type {Array.<{name: string, path: string}>}
+ * Corpus checks, each parsed once at load: the name suppressions match against,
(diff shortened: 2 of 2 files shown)


maxonfjvipon:
@rultor merge

PR #324, opened 2026-07-24
Name out-of-tree files by absolute path in json/sarif

### LINKED ISSUE #320: SARIF and JSON paths escape the working directory for out-of-tree files
`located` in `src/reporters.js:22` turns a defect's absolute file path into a working-directory-relative one for the `json` and `sarif` formats:

```js
const located = function(file) {
  return path.relative(process.cwd(), file).split(path.sep).join('/')
}
```

When the linted file sits outside the working directory, `path.relative` climbs out with `..` segments, producing a URI like `../../../var/folders/.../sheet.xsl`. In the normal invocation — `xslint .` from the repository root — every path stays repo-relative and this never bites. But a run over an out-of-tree target emits escaping paths, and SARIF is meant for GitHub code scanning, which maps `artifactLocation` URIs against the repository checkout and cannot resolve a path that climbs out of it.

Emit the absolute posix path when the relative one would escape the working directory (when `path.relative` returns a `..`-leading or absolute path), so an out-of-tree file is named unambiguously rather than by a confusing relative climb. Paths inside the working directory are unchanged. A note in the README's machine-output section should say SARIF is best generated from the repository root.

maxonfjvipon:
`located` turned a defect's path into a working-directory-relative one for `json`/`sarif`. For a file outside the working directory, `path.relative` climbed out with `..`, producing `../../../var/.../sheet.xsl` — a URI GitHub code scanning maps against the checkout and cannot resolve.

It now falls back to the absolute posix path when the relative one would escape:

```js
const relative = path.relative(process.cwd(), file)
const escapes = relative.startsWith("..") || path.isAbsolute(relative)
return (escapes ? file : relative).split(path.sep).join("/")
```

In-tree paths (the normal `xslint .` run) are unchanged. The README notes SARIF is best generated from the repository root.

Closes #320.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Name an out-of-tree file by its absolute path in `json`/`sarif` output rather
+  than a `..`-climbing relative path that GitHub code scanning cannot map
+  (#320).
+
 - Parse the corpus checks once at module load instead of on every run, matching
   the other linters and validators (#317).
 
--- a/README.md
+++ b/README.md
@@ -162,7 +162,9 @@ pull requests:
 ```
 
 The `|| true` keeps a findings exit code from failing the step before the
-upload; the alerts still surface in code scanning.
+upload; the alerts still surface in code scanning. Run it from the repository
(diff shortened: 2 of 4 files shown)


maxonfjvipon:
@rultor merge

PR #325, opened 2026-07-24
Ban the node: prefix in ESM imports too

### LINKED ISSUE #319: The `node:` require ban does not cover ESM `import`, and `eslint.config.mjs` uses the prefix
The rule that bans the `node:` prefix (added in #304) matches only `require`:

```
CallExpression[callee.name='require'][arguments.0.value=/^node:/]
```

ESM `import` is invisible to it. That leaves two gaps now that the project has ES modules. `eslint.config.mjs:7-8` uses the prefix the rest of the codebase forbids:

```js
import path from "node:path";
import { fileURLToPath } from "node:url";
```

And a future `import fs from 'node:fs'` in `src/index.mjs` — now an ES module and actually linted — would slip past the rule. So the convention ("bare module specifiers, no `node:` prefix") is only half-enforced, and the one ESM config file violates it in spirit.

Extend the rule with a second selector for import declarations, `ImportDeclaration[source.value=/^node:/]`, and switch `eslint.config.mjs` to bare specifiers (`import path from "path"`, `import { fileURLToPath } from "url"`), so the convention holds for every module system. `eslint.config.mjs` sits in the `ignores` list and cannot self-lint, so its own imports must be fixed by hand, but `src/index.mjs` and any future ESM will then be covered.

maxonfjvipon:
The rule added in #304 banned the `node:` prefix only in `require`. Now that the project has ES modules, an `import` could still smuggle it in — and `eslint.config.mjs` itself did.

A second `no-restricted-syntax` selector covers import declarations:

```
ImportDeclaration[source.value=/^node:/]
```

Verified it fails `grunt eslint` on a planted `import fs from "node:fs"` in `src/index.mjs`. `eslint.config.mjs` now uses bare `"path"`/`"url"`, so the one ESM file no longer contradicts the convention it enforces.

Closes #319.


maxonfjvipon:
@rultor merge

PR #326, opened 2026-07-24
Count every src file toward the coverage gate

### LINKED ISSUE #322: Coverage gate ignores files no test loads (`c8` without `all`)
The coverage gate is `check-coverage: true` with every threshold at 100 in `.c8rc.json`, run by the `coverage` job as `npm run coverage`. But `c8` only measures files that were actually *loaded* during the run, because `.c8rc.json` does not set `all`. A source file that no test ever `require`s is silently omitted from the report rather than counted as `0%`, so the 100% threshold is met vacuously and CI stays green.

Proof: running the suite over a single test file exits `0` today, even though almost nothing is covered:

```
c8 mocha test/tokens.test.js   # exit 0 — other files not counted
```

With `"all": true` the same run fails, as it should:

```
ERROR: Coverage for lines (26.64%) does not meet global threshold (100%)
```

So a new module shipped without a test would not fail the build — the gate has a hole. Add `"all": true` to `.c8rc.json` so `c8` instruments every file matched by `include` (`src/**`, minus the excluded `src/index.mjs`) whether or not a test loads it. The full suite still reports 100% and exits `0`, since every module is already exercised; only the escape hatch closes.

maxonfjvipon:
`c8` measured only files that a test actually loaded, because `.c8rc.json` did not set `all`. A source file no test `require`s was omitted from the report rather than counted as `0%`, so the 100% gate was met vacuously and a new untested module would not fail CI.

`"all": true` makes `c8` instrument every file matched by `include` (`src/**`, minus the excluded `src/index.mjs`). Verified: the full suite still reports 100% and exits `0`, while a planted untested file now drops coverage below the threshold and fails the run.

Closes #322.

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -5,6 +5,7 @@
   "exclude": [
     "src/index.mjs"
   ],
+  "all": true,
   "reporter": [
     "text-summary",
     "lcov"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Count every `src` file toward the 100% coverage gate (`c8` `all`), so a module
+  shipped without a test now fails CI instead of being silently omitted (#322).
+
 - Extend the `node:`-prefix ban from `require` to ESM `import` as well, and use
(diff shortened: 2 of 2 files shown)


maxonfjvipon:
@rultor merge

PR #340, opened 2026-07-24
Add a --format github reporter for CI annotations

### LINKED ISSUE #333: Add a `--format github` reporter for inline CI annotations
`--format` selects the output through `reporterOf` in `src/reporters.js` and is validated in `src/index.mjs` against `['text', 'json', 'sarif']`. None of these turn findings into inline annotations when xslint runs inside a GitHub Action (`maxonfjvipon/xslint-action`): `text` is for humans, and `json`/`sarif` need a separate upload step to reach the Security tab.

GitHub Actions renders a step's stdout lines of the form `::warning file=...,line=...,col=...::message` (and `::error ...`) as annotations on the run and inline on the pull-request diff. A `github` reporter would emit exactly that — one workflow command per defect, severity mapped to `warning`/`error`, the path made workspace-relative by the existing `located` helper, the message carrying its text and rule name on a single line (xslint messages are already one sentence).

Add `github` to the `REPORTERS` map in `src/reporters.js` and to the `--format` choices in `src/index.mjs`. The action can then run `xslint --format github` and findings appear inline with no SARIF-upload step — the lowest-friction path to code-review feedback, which is the main reason to adopt a linter in CI.

maxonfjvipon:
Adds a `github` output format that prints GitHub Actions workflow commands, so a run inside a GitHub Action turns each defect into an inline annotation on the pull-request diff — no SARIF-upload step, which is the lowest-friction way to surface findings on a review.

```
::warning file=path/to/sheet.xsl,line=16,col=3,title=short-names::A variable... has a single-character name.
```

`github` joins `text`/`json`/`sarif` in the `reporterOf` map and the `--format` choices; severity maps to `::warning`/`::error`, the path reuses the existing `located` helper, and the rule name becomes the annotation title. Values are escaped per the workflow-command rules. Covered by reporter unit tests and an end-to-end `--format github` test; suite stays at 100% coverage.

This unblocks maxonfjvipon/xslint-action#22 (the action can pass `--format github` for inline annotations).

Closes #333.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add a `--format github` reporter that prints GitHub Actions workflow
+  commands, so findings render as inline annotations on the pull-request diff
+  with no SARIF-upload step (#333).
+
 - Evaluate suppression once per run in the per-file XPath linter instead of
   re-checking it for every file, matching the other linters (#331).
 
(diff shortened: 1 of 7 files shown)


maxonfjvipon:
@rultor merge

PR #350, opened 2026-07-24
Read the changelog for release notes; drop the master push

### LINKED ISSUE #349: release.yml fails on Rultor's existing release and cannot push to protected master
The `0.0.8` release published to npm (OIDC + provenance worked), but the `release.yml` run concluded failure at the post-publish steps, because the workflow was built against two assumptions this repo does not hold:

- **Rultor already creates the GitHub Release** for the tag (every version is `author=rultor`), so `release.yml`'s `gh release create` failed with `422 - Release.tag_name already exists`.
- **`master` is protected** (`required_pull_request_reviews: true`), so the workflow's `git push ... master` to promote the changelog would be rejected regardless.

Adopt the pre-tag-promotion model (approach A):

- `release.yml` only *reads* the changelog: it extracts the notes for the released version and sets them on the GitHub Release with `gh release edit ... || gh release create ...` — no duplicate error, and Rultor's generic notes get replaced by the real changelog entries.
- Drop the `git push ... master` step entirely.
- The `## Unreleased` -> `## X.Y.Z` promotion happens in a normal pull request before tagging, so it goes through the protected-branch flow instead of a CI push.

Also promote the changelog for `0.0.8` (it shipped but was never recorded) and prepare `0.0.9`.

maxonfjvipon:
`0.0.8` published fine (OIDC + provenance), but `release.yml` concluded failure after publishing, because it was built against assumptions this repo does not hold:

- **Rultor already creates the GitHub Release** (every version is `author=rultor`), so `gh release create` returned `422 - Release.tag_name already exists`.
- **`master` requires pull requests**, so the changelog `git push ... master` could not land.

Adopting the pre-tag-promotion model (approach A):

- `release.yml` now *reads* the changelog and sets the notes with `gh release edit "$VERSION" --notes-file` (falling back to `create`) — no duplicate error, and Rultor's generic notes are replaced with the real changelog entries.
- The `git push ... master` step is gone.
- `scripts/changelog-release.js` becomes `scripts/changelog-notes.js`, a read-only extractor for a version's section.
- The `## Unreleased` -> `## X.Y.Z` promotion happens here, in the pre-tag PR: this one promotes the shipped-but-unrecorded `0.0.8` and prepares `0.0.9`.

Next: merge, then `@rultor release, tag=0.0.9` to confirm the workflow goes green end to end.

Closes #349.


maxonfjvipon:
@rultor merge

PR #356, opened 2026-07-25
Enforce a 100-character line length for Markdown

### LINKED ISSUE #355: Enforce a 100-character line length for Markdown
Markdownlint currently disables `line-length` entirely. Set it to 100 for prose, exempting code blocks and tables (which have inherently long lines), so Markdown stays readable and the rule is enforced on new docs.

maxonfjvipon:
Markdownlint had `line-length` disabled entirely. This turns it on at 100 characters, so Markdown prose stays readable and the limit is enforced on new docs.

- `line_length: 100`, with `code_blocks` and `tables` exempt (their lines can't wrap).
- `CLAUDE.md` is exempt via an inline `markdownlint-disable MD013` — it is dense reference prose written one paragraph per line, with several 200–1200-character lines that would be impractical to wrap.
- One motive line (105 chars) is rewrapped.

README badges are already exempt (MD013 does not fire on unbreakable long URLs), and the other auto-fixable rules are handled by the workflow's `fix: true`. Verified against the CI markdownlint version (0.17.2): zero unfixable issues remain.

Closes #355.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,3 +1,7 @@
 {
-  "line-length": false
+  "line-length": {
+    "line_length": 100,
+    "code_blocks": false,
+    "tables": false
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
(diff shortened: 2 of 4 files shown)


maxonfjvipon:
@rultor merge

PR #358, opened 2026-07-25
Auto-bump the xslint-action version in the README

### LINKED ISSUE #357: Auto-bump the xslint-action version referenced in the README
The README references `maxonfjvipon/xslint-action@X.Y.Z` in the CI example. When a new xslint-action release is cut, that pin goes stale. Add a scheduled workflow (mirroring `up.yml`/the action's `up-xslint.yml`) that reads the latest xslint-action release, updates the reference, and opens a PR.

maxonfjvipon:
The README's CI example pins `maxonfjvipon/xslint-action@X.Y.Z`, which goes stale when the action publishes a new release.

`up-xslint-action.yml` runs daily, reads the latest `xslint-action` release with `gh release view`, and — if the README's pin is behind — updates it and opens a PR (via `peter-evans/create-pull-request`, as `up.yml` already does). Unlike `up.yml`, it queries the *other* repository, because the action's tags don't trigger an event here.

```yaml
sed -E -i "s|xslint-action@[0-9.]+|xslint-action@${{ steps.version.outputs.latest }}|g" README.md
```

Its `sed` targets `xslint-action@`, which is disjoint from `up.yml`'s `xslint@` (that pattern never matches `xslint-action@`), so the two auto-bump workflows do not interfere. Verified the `gh` query returns `0.0.5` and the sed bumps a stale reference.

Closes #357.

### DIFF
--- a/.github/workflows/up-xslint-action.yml
+++ b/.github/workflows/up-xslint-action.yml
@@ -1,0 +1,40 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+# yamllint disable rule:line-length
+name: up-xslint-action
+'on':
+  schedule:
+    - cron: '0 6 * * *'
+concurrency:
+  group: up-xslint-action-${{ github.ref }}
+  cancel-in-progress: true
+jobs:
+  up:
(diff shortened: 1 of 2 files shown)


maxonfjvipon:
@rultor merge

Issue #334, opened 2026-07-24
Add a `--fix` mode that rewrites the mechanically-fixable defects


maxonfjvipon:
Autofix is what turns a linter from "ran it once" into "it lives in my pre-commit hook" — the biggest retention driver for ESLint, Prettier, and Stylelint. This issue tracks `--fix` as a **program**, not a single PR: an engine plus a catalog of fixers landed in waves.

## Two tiers, not one

The mistake is to ask "is the fix 100% safe?" and put everything else in "never". Mature linters split fixes in two:

- **Auto-fix** (`--fix`, applied silently) — deterministic *and* semantics-preserving.
- **Suggestion** (offered, applied with a stronger flag or in an editor) — deterministic but *opinionated*, a *removal*, *one of several* valid fixes, or carrying a semantic caveat a human should okay.

With that split, roughly **half of the ~47 checks become fixable in some form**.

### Auto-fix — safe, semantics-preserving

- `redundant-whitespace` — collapse a doubled space, trim a leading/trailing one. **(done, phase 1)**
- `unabbreviated-axis` — `child::x`→`x`, `attribute::x`→`@x`, `parent::node()`→`..` (exact XPath equivalences).
- `use-node-set-extension` — strip the `exsl:node-set(…)` wrapper (guarded to XSLT ≥ 2.0, where it is a no-op).
- `redundant-namespace-declarations` — delete the `xmlns:` the check proved unused.
- `blank-nested-if` — merge nested `xsl:if` into one with `and` (parenthesise the operands for precedence).
- `use-single-option-for-choose` — a one-`when` `xsl:choose` → `xsl:if`.

### Suggestion — deterministic but opinionated / a removal / one of several / caveated

- `incorrect-use-of-boolean-constants` — `'true'`→`true()` (parser-gated so it never touches a real string literal).
- `unused-variable`, `unused-function`, `unused-function-template-parameter`, `unused-named-template` — remove (may be intentional).
- `duplicate-param-name`, `duplicate-with-param-name` — drop the later duplicate.
- `empty-content-in-instructions`, `empty-variable` — remove the empty element.
- `using-disable-output-escaping` — remove the attribute (changes escaping).
- `setting-value-of-variable-incorrectly` — nested `xsl:value-of` → `@select` (**caveat:** changes the bound value's type from a text fragment to the selected value).
- `not-creating-element-correctly` — `xsl:element` with a static name → a literal element (**caveat:** namespace context).
[...]

PR #359, opened 2026-07-25
Add a --fix mode that rewrites redundant whitespace

maxonfjvipon:
Autofix is the strongest retention lever a linter has, and xslint could only report. This adds `--fix`, starting with the one check whose correction is unambiguous — `redundant-whitespace`.

The engine is deliberately check-agnostic. A defect may carry a fix, and that shape is the whole contract:

```js
fix: {
  line: expression.lineNumber,
  col: pos,
  value: value,          // the exact text to replace
  replacement: replacement,  // '' to trim, ' ' to collapse
}
```

`src/fixer.js` maps each fix to a source offset and applies it only when the span still holds that exact `value`, so a shifted offset — an entity ahead of the run, an already-edited file — is skipped rather than corrupting the source. A file's fixes are applied from the end backwards so earlier offsets stay valid. `src/xslint.js` writes the changed files (unless `--fix-dry-run`) and drops the applied defects, so the exit code reflects only what remains.

Because fixability lives entirely in the `fix` shape, the next fixers — axis abbreviation, unused-namespace removal — attach a fix and need no engine change. This is phase 1 of the plan on #334.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add a `--fix` mode (`--fix-dry-run` to preview) that rewrites the
+  mechanically-fixable defects in place. It covers `redundant-whitespace`
+  today: a check-agnostic engine (`src/fixer.js`) applies the `fix` a defect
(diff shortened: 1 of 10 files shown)


maxonfjvipon:
@rultor merge

PR #360, opened 2026-07-25
Make unabbreviated-axis fixable via a token-aware check

maxonfjvipon:
Phase 2 of the `--fix` program (#334): make `unabbreviated-axis` auto-fixable.

The old check was a coarse declarative rule — `contains(@*, "child::")`. It reported once per element, would match `child::` inside a string literal, and flagged `parent::x` even though only `parent::node()` has an abbreviation. None of that can drive a precise fix.

This replaces it with a token-aware linter that scans every XPath and pattern attribute across the corpus, reports one defect per *real* verbose axis, and attaches the fix that abbreviates it:

```js
child::x         → x
attribute::x     → @x
parent::node()   → ..
parent::n        → left alone (no short form)
```

Because it reads the corpus rather than the validator's expression stream, an axis in a template `match` pattern is caught too — and since each occurrence is its own defect fully resolved by its own fix, it rides the phase-1 fix engine unchanged (no re-lint). The lexer keeps string literals whole, so an axis-looking substring inside one is never rewritten.

Two behavior refinements fall out, both discussed and agreed beforehand: positions now point at each axis rather than the whole element, and the check moves from the `xpath` category to `format`, where axis spelling belongs. End-to-end `--fix` output byte-matches the committed `.fixed` fixture, coverage stays at 100%.


maxonfjvipon:
@rultor merge

PR #361, opened 2026-07-25
Make redundant-namespace-declarations fixable

maxonfjvipon:
Phase 3 of the `--fix` program (#334): make `redundant-namespace-declarations` auto-fixable — the first attribute-level fix.

The old check was a declarative rule that fired once per stylesheet and gave no handle on *which* prefix was dead. This replaces it with a code-based linter that walks the DOM, and reports one defect per prefix declared on the stylesheet but used by no element name, attribute name, or qualified name in an attribute value — each positioned at its declaration with a fix that deletes it.

Deleting an attribute needs its exact source span, including the quote character the DOM abstracts away. Rather than plumb raw source everywhere, the fix reconstructs the deletion text from the declaration and lets the existing verify-before-apply guard vet it:

```js
value: ` ${attribute.name}="${attribute.value}"`,   // e.g. ` xmlns:foo="urn:foo"`
replacement: '',
```

So a standard `xmlns:foo="…"` is deleted, and a single-quoted or oddly spaced one is reported but left untouched rather than mis-edited. Detection is unchanged — verified against the existing fixtures (redundant → 1, used → 0). Each dead declaration is its own defect fully resolved by its own deletion, so it rides the phase-1 engine unchanged.

Two consequences worth flagging, consistent with the precise, per-occurrence direction set in phase 2: the position moves from the stylesheet element to each declaration, and the check moves from the `xpath` category to `format`. Porting detection to code also orphaned the `xslint:in-scope-prefixes` custom XPath function — its only caller was this rule — so it's removed. End-to-end `--fix` byte-matches the committed `.fixed` fixture; coverage stays at 100%.


maxonfjvipon:
@rultor merge

Issue #228, opened 2026-06-24
Grow the XPath front end into a position-preserving CST


maxonfjvipon:
## Why xslint needs its own XPath front end

Formatting checks reason about how an expression is *written* — its whitespace, its
parentheses, its axis spelling — not just what it *means*. That information lives in
the source text and its exact positions, and it is precisely what a normal parser
throws away.

We confirmed this with the engine xslint already depends on. `fontoxpath.parseScript`
returns a W3C XQueryX AST, but it is abstract: `foo(  a)  =  'hi'` comes back as
`<xqx:functionCallExpr>` / `<xqx:equalOp>` with the doubled spaces normalised away and
**no source positions** on any node (no line/column/offset). The only place the
original text survives is a verbatim `<!-- -->` comment — useless for pointing at which
space, at which column, is redundant.

So a formatter cannot ride on the execution engine's parser. It needs its own
**position-preserving (lossless) CST**, not a semantic AST: the dependency we already
have builds the latter and it is insufficient. This is the standard architecture for
formatters and linters — Prettier, Roslyn, and rust-analyzer all keep a lossless
syntax layer separate from the compiler's AST for exactly this reason.

## Where we are now

`src/tokens.js` is no longer the four-kind stream this issue was filed against. Its
`TOKENS` map carries 52 kinds: every operator, both brackets, the numbers, all
thirteen axis names, and — since #249 and #666 — a name lexed whole and greedily, so
`border` is one token rather than `b`/`or`/`der`, and a word is an operator only where
the grammar admits one. Gaps are handled where they belong: around an axis's `::`
(#615), in front of a node test's opening bracket (#615) and the one it closes with
(#639), and inside an axis token that had swallowed one (#642).

**Two checks that step 3 assumed needed the tree shipped without it.**
`unabbreviated-axis` does the axis abbreviations — `child::a` to `a`, `attribute::a`
to `@a`, `self::node()` to `.`, `parent::node()` to `..` — straight off the token
stream (#564, #616), and `predicate-position-literal` normalises `[position() = 1]`
to `[1]` the same way. The stream carries more structural work than this roadmap
credited it with, so what is left for the tree is narrower and sharper than the
original list.

**One argument in the July design note has since closed.** It named escaping as a
reason for a parser over raw source: fixes positioned from the decoded `nodeValue`
while the source holds `&gt;`. `src/fixer.js` now decode-walks the raw text (#518,
#525), so an escaped comparison fixes correctly with no tree involved. The other two
[...]

PR #362, opened 2026-07-25
Make use-node-set-extension fixable

maxonfjvipon:
The last non-structural auto-fix on #334: make `use-node-set-extension` fixable. The `exsl:node-set(…)` extension is redundant in XSLT 2.0+, where a variable is already a node sequence — so the fix unwraps it, keeping the argument:

```
exsl:node-set($x)        → $x
exsl:node-set($x)/title  → $x/title
msxsl:node-set(foo(1))   → foo(1)
```

The wrinkle is that the lexer splits `node-set` at its hyphen (it reads the `-` as a subtraction operator), so token classification can't spot the call. Instead the linter **masks** string and comment spans with the lexer, then finds `prefix:node-set(` and balances its parentheses on the blanked text. That makes two things fall out for free: a nested argument like `node-set(foo(1))` matches correctly, and a `node-set(` sitting inside a string literal is never flagged — verified in a test.

Each call is its own defect fully resolved by unwrapping its single span, so it rides the phase-1 engine unchanged (no re-lint). Consistent with the earlier phases, the check becomes precise and per-occurrence, and moves from the `xpath` category to `format`. End-to-end `--fix` byte-matches the committed `.fixed` fixture; coverage stays at 100%.

With this, **four checks are auto-fixable**. The only remaining pure-auto fixes — `blank-nested-if` and `use-single-option-for-choose` — are structural element rewrites that need the concrete syntax tree from #228; everything else is a suggestion.


maxonfjvipon:
@rultor merge

PR #363, opened 2026-07-25
Add a suggestion tier to --fix

maxonfjvipon:
Phase 2 of the `--fix` roadmap (#334): a two-tier fix model, so the fixable set can grow past the handful of edits safe to apply silently.

A fix now carries an optional `suggestion` flag. A plain fix is *safe* — deterministic and semantics-preserving — and `--fix` applies it. A fix with `suggestion: true` is *opinionated* — it changes behavior, removes code, or is one of several valid corrections — so `--fix` leaves it and only the new `--fix-suggestions` applies it:

```bash
xslint --fix              # safe fixes only
xslint --fix-suggestions  # safe fixes + suggestions
```

The other half of this is that a **declarative** xpath rule can now carry a fix without being rewritten as a code-based linter. `src/fixers.js` maps a check name to a `node => fix` builder, and the per-file linter attaches whatever it returns:

```js
const FIXERS = {
  'using-disable-output-escaping': disableOutputEscaping,
}
```

`using-disable-output-escaping` is the first suggestion — deleting the attribute changes how output is escaped, so you opt in. It shares the `deletion(attribute)` helper (`src/fixes.js`) with the namespace linter. And a run *without* `--fix` now nudges: it reports how many defects `--fix` would fix and how many more `--fix-suggestions` would.

This unlocks the whole removals / opinionated-defaults class (unused-*, `missing-version`, boolean constants, `output-method-xml`, …) as follow-ups, each just registering a builder. The structural ones — deleting an unused variable, merging nested `xsl:if` — still wait on the concrete syntax tree from #228 for exact per-node spans. Coverage stays at 100%.


maxonfjvipon:
@rultor merge

PR #364, opened 2026-07-25
Add three registry suggestions to --fix

maxonfjvipon:
Three more `--fix-suggestions`, each a single entry in the `src/fixers.js` registry — the payoff of the tier from #363. No new linter, no category move: a declarative rule carries a fix.

```js
const FIXERS = {
  'using-disable-output-escaping': disableOutputEscaping,
  'output-method-xml': outputMethodXml,            // method="xml" → "html"
  'missing-version-in-stylesheet': missingVersion, // insert version="1.0"
  'mode-or-priority-without-match': modeOrPriority, // delete the orphan attr
}
```

`mode-or-priority-without-match` shows the one wrinkle: a template can carry *both* `mode` and `priority`, which one deletion can't resolve, so the builder returns null there and the linter leaves the defect fix-less (reported, not fixed) rather than half-fixing it and wrongly dropping it. `output-method-xml` is a value edit, `missing-version` an insertion, the other two attribute deletions sharing `deletion()` with the namespace linter.

Each is verified end-to-end under `--fix-suggestions` (byte-exact `.fixed` fixtures), and the existing tier test still guards that plain `--fix` leaves them. Coverage stays at 100%.

## One deliberately left out: `incorrect-use-of-boolean-constants`

The fourth candidate is a mistake to auto-fix. Its fixture is `test="@ooo = 'true'"`, and the "fix" `@ooo = true()` changes a **string** comparison (is the attribute the text "true"?) into a **boolean** one (does @ooo exist?) — usually not what the author meant. The rule's substring match conflates a legitimate `= 'true'` comparison with the real misuse (a bare `test="true"`), so an auto-fix would be wrong more often than right, even as a suggestion. It needs the #228 parser for context, and arguably a rule-design review, so it's left report-only.


maxonfjvipon:
@rultor merge

Issue #336, opened 2026-07-24
Editor integration: an LSP server and a VS Code extension


maxonfjvipon:
Linters spread through in-editor squiggles — the feedback loop that makes developers keep them. xslint runs only in the terminal and CI today, so a developer never sees a defect until a build fails.

An LSP server that reuses the existing validators and linters would surface xslint diagnostics as the user types, lighting up VS Code and, through LSP, other editors (Neovim, and IntelliJ via its LSP support). The defect shape xslint already produces — `{file, line, pos, severity, message, name}` — maps directly onto an LSP `Diagnostic`, so the server is mostly a transport around the existing engine. A thin VS Code extension wrapping the server gives a Marketplace presence.

This is the largest effort of the integration items and the best long-term retention. Likely a separate repository. Start with a standalone LSP server (stdio) that lints on open/change, then publish the VS Code extension on top of it.

maxonfjvipon:
Done — the xslint-lsp server is released (0.0.8) and the VS Code extension is published on Open VSX.

PR #378, opened 2026-07-26
Expose a programmatic lint API

maxonfjvipon:
Phase A of the editor-integration work (#336): give xslint a programmatic API, so the coming LSP server can lint an in-memory buffer in-process instead of shelling out to the CLI per keystroke.

Today the only entry is the CLI, and `main` points at `src/index.mjs` — which parses `process.argv` and can `process.exit` on load. There's nothing safe to `require`.

This lifts the sources-to-defects core out of the CLI into a pure function:

```js
const {lint, fixed} = require('@maxonfjvipon/xslint')
const defects = lint([{file: 'sheet.xsl', content}], {suppress: ['short-names']})
const {contents} = fixed([{file: 'sheet.xsl', content}], defects)
```

`lint(sources, {suppress, overrides})` validates, runs every linter, applies severity overrides, and drops inline-directive-suppressed defects — no file I/O, no printing, no exit. The command-line `xslint()` is now a thin wrapper: discover and read files, call `lint`, apply `--fix`, report, set the exit code. `main` moves to `src/xslint.js` (which re-exports `lint` and `fixed`); the CLI bin stays `src/index.mjs`.

Behavior is unchanged — the whole existing CLI/config/directive suite still passes, coverage stays at 100%, and requiring the package no longer has side effects. The defect shape (`{name, severity, message, file, line, pos, fix?}`) maps straight onto an LSP `Diagnostic`, which is the next phase.


maxonfjvipon:
@rultor merge

PR #386, opened 2026-07-26
Give the up workflows a PAT so they can open PRs

maxonfjvipon:
After the transfer, the `up` and `up-xslint-action` workflows go red with `Resource not accessible by integration` — the new org defaults `GITHUB_TOKEN` to read-only (`default_workflow_permissions: read`), so `peter-evans/create-pull-request` can't create the branch/PR.

Fix: pass `secrets.DISPATCH_TOKEN` (already an org secret for the release cascade) to the two `create-pull-request` steps. A PAT is exempt from both the read-only default and the org's "Actions cannot create pull requests" policy, so this works without loosening any org-wide Actions setting.


maxonfjvipon:
@rultor merge

Issue #396, opened 2026-07-27
invalid-xpath-expression false-positives on XPath 1.0 numeric coercion


maxonfjvipon:
The XPath validator in `src/xpath-validator.js` decides an expression is valid
by compiling it with fontoxpath through `isValid` in `src/xpath.js`. fontoxpath
is an XPath 3.1 engine and applies static typing during compilation, so it
rejects an arithmetic operator whose operand is statically `xs:string` — even
though the expression parses cleanly:

```
substring-before($spans, ':') - 1    -> reported invalid-xpath-expression
concat($spans, ':') - 1              -> reported invalid-xpath-expression
count(*) - 1                         -> accepted (operand is numeric)
number(substring-before($spans, ':')) - 1  -> accepted (wrapped in number())
```

Under XPath 1.0 — the dialect DocBook and a large share of real stylesheets
target (`version='1.0'`) — the `-` operator converts both operands to numbers,
so `substring-before($spans, ':') - 1` is valid and idiomatic: it reads the
numeric prefix of a string. xslint reports it as a syntax error and skips the
expression from every downstream linter. DocBook's `table.xsl` and
`utility.xsl` trip this three times.

This breaks the validator's stated contract — CLAUDE.md says `isValid` fails
"only genuine syntax errors." A static-type rejection is not a parse error. The
fix is to separate the two: accept an expression that parses even when
fontoxpath's static typer objects, or validate a `version='1.0'` stylesheet
under XPath 1.0 semantics where the coercion is legal.

Issue #395, opened 2026-07-27
malformed-stylesheet false-positives on an internal DTD subset


maxonfjvipon:
The XML validator in `src/xsl-validator.js` rejects any stylesheet that declares
entities in an internal DTD subset, even though the file is well-formed XSLT.
DocBook's core stylesheets open exactly this way:

```xml
<?xml version='1.0'?>
<!DOCTYPE xsl:stylesheet [
<!ENTITY lowercase "'abcdefghijklmnopqrstuvwxyz'">
<!ENTITY uppercase "'ABCDEFGHIJKLMNOPQRSTUVWXYZ'">
]>
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version='1.0'>
```

The `@xmldom/xmldom` parser xslint reads with does not accept the internal
subset here, so `xsl-validator` classifies the whole file as
`malformed-stylesheet` and drops it — none of its templates or expressions are
linted. This pattern is not exotic; DocBook-XSL and TEI use it heavily. Running
xslint over DocBook's 18 core stylesheets (`xsl/common`), three well-formed
files — `common.xsl`, `autoidx-kimber.xsl`, `autoidx-kosek.xsl` — are lost this
way, reported only as malformed at `1:1`.

The fix is to resolve (or tolerate) internal-subset entity declarations before
handing the document to the parser, so a stylesheet that happens to define
`&lowercase;` is linted like any other rather than skipped whole.

PR #397, opened 2026-07-27
Add a 'proven on real code' section

maxonfjvipon:
Runs xslint over the 18 core stylesheets of [DocBook-XSL](https://github.com/docbook/xslt10-stylesheets) — mature, widely-used XSLT 1.0 — and reports the outcome on the README and the generated landing page.

Stripped of cosmetic whitespace nits and two validator false positives (filed as #395 and #396), the run surfaced over 200 substantive findings across 18 different checks: 41 `xsl:choose` blocks with no `xsl:otherwise`, 30 named templates never invoked, 12 unused parameters, variables set via nested `xsl:value-of`, and more. Concrete credibility for the linter, in code that has shipped for two decades.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,15 @@ get inline annotations on your pull requests:
 
 Browse the full [check catalog](https://xslint.github.io/xslint/).
 
+## Proven on real code
+
+Pointed at the 18 core stylesheets of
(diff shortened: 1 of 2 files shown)


maxonfjvipon:
@rultor merge

PR #400, opened 2026-07-27
Do not flag XPath 1.0 numeric coercion as invalid

### LINKED ISSUE #396: invalid-xpath-expression false-positives on XPath 1.0 numeric coercion

maxonfjvipon:
Fixes #396.

The `invalid-xpath-expression` validator compiles every expression with fontoxpath, an XPath 3.1 engine that applies static typing and rejects string arithmetic — `substring-before($spans, ':') - 1` throws `XPTY0004`. That coercion is valid and idiomatic in XPath 1.0 (it reads a numeric prefix), so the validator was reporting a false positive and dropping the expression from every downstream linter. It fired on real code — DocBook's `table.xsl` and `utility.xsl`.

`isValid` now distinguishes a genuine syntax error from a static-type complaint by the shape of the engine's message: a parse failure reads `<position>: <source>`, a static or type error a W3C code like `XPTY0004`. Only the former marks the expression malformed — so unknown prefixes, custom functions, and 1.0 coercion all pass, while `1 +` and friends are still caught.

Reproduced by the new `valid-numeric-coercion` pack (expects zero defects); the existing `invalid-*` packs still assert real syntax errors are reported.


maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor merge

PR #401, opened 2026-07-27
Tolerate entities declared in an internal DTD subset

### LINKED ISSUE #395: malformed-stylesheet false-positives on an internal DTD subset

maxonfjvipon:
Fixes #395.

`@xmldom/xmldom` never expands entities declared in an internal DTD subset, so a reference to one — DocBook's `&lowercase;` / `&uppercase;` — surfaced as an `entity not found` error and `xsl-validator` reported the whole, perfectly well-formed stylesheet as `malformed-stylesheet` and dropped it from the corpus unlinted. DocBook's `common.xsl` was lost this way.

The parser now collects the general-entity names the source declares in its own subset and swallows an `entity not found` error for any of them — a declared-but-unexpanded entity is not a malformed document. A reference to a genuinely undeclared entity is still reported, so the deliberate undefined-entity detection is preserved (both cases are covered by new tests). With this, `common.xsl` parses and lints (190 findings) instead of being dropped.

Out of scope: two DocBook files (`autoidx-kimber.xsl`, `autoidx-kosek.xsl`) declare their entities in an *external* DTD pulled in through a `SYSTEM` parameter entity (`<!ENTITY % common.entities SYSTEM "entities.ent">`). Resolving external entity files is a separate, larger problem and stays for a follow-up.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -9,20 +9,46 @@ const {DOMParser} = require('@xmldom/xmldom')
 const yaml = require('yaml')
 
 /**
- * XML parser, built once and reused. Its error handler raises on any
+ * Names of the general entities the given source declares in an internal DTD
+ * subset. `@xmldom/xmldom` never expands them, so a reference to one surfaces
+ * as an "entity not found" error even though the entity is perfectly well
+ * declared — DocBook and TEI stylesheets rely on exactly this. Knowing the
+ * declared names lets the parser tell that recoverable case apart from a
+ * genuinely undefined entity.
+ * @param {string} str - XML source
+ * @return {Set.<string>} - Declared general entity names
+ */
+const declaredEntities = function(str) {
+  const names = new Set()
+  for (const match of str.matchAll(/<!ENTITY\s+([A-Za-z_][\w.-]*)\s/g)) {
+    names.add(match[1])
+  }
+  return names
+}
+
+/**
+ * XML parser for the given source. Its error handler raises on any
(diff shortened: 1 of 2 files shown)


maxonfjvipon:
@rultor merge

PR #404, opened 2026-07-27
Accept the namespace axis as valid XPath

### LINKED ISSUE #402: invalid-xpath-expression false-positives on the namespace axis
The `invalid-xpath-expression` validator compiles each expression with
fontoxpath, an XPath 3.1 engine. XPath 3.0 dropped the `namespace::` axis that
1.0 and 2.0 define, and fontoxpath does not implement it, so a perfectly valid
expression such as

```xsl
<xsl:for-each select="namespace::*">
```

fails to parse and is reported as `invalid-xpath-expression`. It is common in
XSLT 1.0 stylesheets that copy namespace nodes — it fires on TEI's
`common/verbatim.xsl`. `child::`, `attribute::`, `descendant::` and the other
axes all parse; only `namespace::` breaks.

Unlike the numeric-coercion false positive (#396), this is a genuine parse
failure, not a static-type error carrying a W3C code, so the error-code
discriminator in `isValid` cannot tell it apart from a real syntax error.

Proposed fix: recognise the namespace axis and treat such an expression as
valid — for example, validate a copy with `namespace::` rewritten to a
supported axis, and accept the original when only that rewrite makes it parse.

maxonfjvipon:
Fixes #402.

fontoxpath is an XPath 3.1 engine, and XPath 3.0 dropped the `namespace::` axis that 1.0 and 2.0 define, so `select="namespace::*"` — idiomatic in 1.0 stylesheets that copy namespace nodes — failed to parse and was reported `invalid-xpath-expression`, then dropped from every downstream linter. It fires on TEI's `common/verbatim.xsl`. Unlike the numeric-coercion false positive (#396), this is a genuine parse failure with no W3C error code, so the discriminator cannot distinguish it from a real syntax error.

`isValid` now retries a failing expression with its `namespace::` axis rewritten to `child::` and accepts the original when only that rewrite makes it parse — so `namespace::*` passes while `namespace::` (incomplete) and `1 +` are still caught. The compile step is factored into a `compiles` helper.

New `valid-namespace-axis` pack expects zero defects; the `invalid-*` packs still assert real syntax errors are reported.


maxonfjvipon:
@rultor merge

PR #405, opened 2026-07-27
Resolve declared entities so entity-using stylesheets lint cleanly

### LINKED ISSUE #403: Resolve declared entities so entity-using stylesheets lint without false positives
#395 stopped dropping a stylesheet that declares internal-subset entities, but
two residual false positives remain because the entities themselves are never
resolved:

1. **Internal entity in an expression.** `@xmldom/xmldom` leaves the reference
   unexpanded in the node value, so DocBook's `common.xsl` keeps the literal
   `translate($format,&lowercase;,&uppercase;)` in a `@select`, and the
   expression validator reports it as `invalid-xpath-expression` — `&` is not
   an XPath operator. The file lints, but that one expression is a false
   positive.

2. **External entities.** DocBook's `autoidx-kimber.xsl` and `autoidx-kosek.xsl`
   declare their entities in an external DTD pulled in through a parameter
   entity — `<!ENTITY % common.entities SYSTEM "entities.ent">` — so none of
   `&primary;` and friends resolve, and the whole file is still reported
   `malformed-stylesheet`.

Proposed fix:

- Expand the entities a source declares in its own internal subset into the
  parsed node values (attributes and text). Positions stay correct, because
  `@xmldom/xmldom` sets line and column from the original source at parse time
  and the expansion only rewrites in-memory values. `common.xsl`'s expression
  then reads `translate($format, '…', '…')` and validates and lints normally.
- When the DOCTYPE reaches for an external subset — a `SYSTEM`/`PUBLIC`
  identifier or a parameter entity — tolerate an unresolved entity rather than
  reporting the file malformed: the entity may well be declared in the DTD we
  did not read.
- Skip validating an expression whose value still carries an unexpanded entity
  reference. We cannot parse what we cannot resolve, so it is neither reported
  nor linted.

Together these take the DocBook/TEI/DITA corpus to zero validator false
positives.

maxonfjvipon:
Fixes #403.

After #395 a stylesheet that declares internal-subset entities is no longer dropped, but `@xmldom/xmldom` still leaves the references literal, so two false positives remained: DocBook's `common.xsl` kept `translate($format,&lowercase;,&uppercase;)` in a `@select` and it was reported `invalid-xpath-expression`, and `autoidx-kimber/kosek.xsl` — whose entities come from an external `entities.ent` via a `SYSTEM` parameter entity — were still reported `malformed-stylesheet`.

- `helpers` expands the entities a source declares inline into the parsed attribute and text values. Positions are untouched: `@xmldom/xmldom` fixes line and column from the original source at parse time, and only in-memory values change. `common.xsl`'s expression then reads `translate($format, '…', '…')` and validates and lints.
- When the DOCTYPE reaches for an external subset (a `SYSTEM`/`PUBLIC` id or a parameter entity), an unresolved entity is forgiven rather than reported malformed — the entity may well live in the DTD we did not read.
- The XPath validator skips an expression still carrying an unresolved entity reference: `&` is not an XPath operator, so it cannot be parsed, and reporting it would be a false positive over a resolution gap.

DocBook's `common/` subset now reports **zero** validator false positives (was 6), and its two external-entity files lint instead of being dropped. Covered by new xsl-validator tests (external tolerance, internal expansion) and a `valid-unresolved-entity` pack.


maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor merge

PR #408, opened 2026-07-27
Follow unused functions and variables across the corpus

### LINKED ISSUE #407: unused-function and unused-variable false-positive across xsl:import/include
`unused-function` and `unused-variable` are per-file checks (`checks/xpath/`),
while `unused-named-template` is correctly cross-file (`checks/corpus/`). A
multi-file XSLT project shares functions and top-level variables through
`xsl:import`/`xsl:include`: a library declares them and the importing
stylesheets use them. Because the two per-file checks only look inside the
declaring file, every function in a `_funcs.xsl` and every variable in a
`_specials.xsl` is reported as unused even though it is called elsewhere — a
false positive on exactly the shared-library pattern these files exist for.

Running xslint over [objectionary/eo](https://github.com/objectionary/eo)
surfaces this: `eo:has-data`, declared in `_funcs.xsl` and called in 11 places
across the parser, is flagged unused; so are seventeen top-level variables in
`_specials.xsl` that importing stylesheets reference.

Both checks should follow usage across the corpus, as `unused-named-template`
does. Their use lives inside an XPath expression (`name(`, `$name`) rather than
in an attribute, so the corpus check needs substring matching, excluding the
declaration's own subtree so recursion alone is not use. Variables also need
scope: a local one is used only within its parent, a top-level one anywhere in
a file that imports it.

maxonfjvipon:
Fixes #407.

`unused-function` and `unused-variable` were per-file (`checks/xpath/`), while `unused-named-template` was correctly cross-file. So in a multi-file project every function in a shared `_funcs.xsl` and every top-level variable in a `_specials.xsl` was reported unused even when an importing stylesheet used it — a false positive on exactly the shared-library pattern. It fires all over [objectionary/eo](https://github.com/objectionary/eo).

Both checks move to the corpus linter, which gains a substring **`reference`** mode: a declaration is unused when its name — substituted into the template (`{name}(` for a function, `${name}` for a variable) — occurs in no usage value across the corpus, excluding the declaration's own subtree so recursion alone is not use. A **`scoped`** check (a variable) counts a usage only within the declaration's parent, or — for a top-level one — in any file that imports it; a function is global.

The per-file edge packs (recursion, similar names, a function called from another, out-of-scope variable use) move to corpus packs and still pass. Re-running over eo, `unused-function` drops 19 → 1 and `unused-variable`'s shared-library false positives clear — the survivors are genuinely-dead symbols and vendored `xs3p`, not false positives.


maxonfjvipon:
@rultor merge

PR #410, opened 2026-07-27
Flag boolean constants only in a boolean context

### LINKED ISSUE #409: incorrect-use-of-boolean-constants false-positives on strings used as data
`incorrect-use-of-boolean-constants` matches a bare substring over every
attribute — `//*[@*[contains(., 'true') and not(contains(., 'true()'))]]` — so
it fires on things that are not bugs: string comparisons (`@active = 'true'`,
the correct way to test an XML boolean attribute), literal output
(`<td hidden="true"/>`), string parameter defaults
(`<xsl:param name="coverage" select="'false'"/>`), and `translate($x,'TRUE','true')`
normalisation. Over [objectionary/eo](https://github.com/objectionary/eo) it
reports 82 defects, 79 of them `$x = 'true'` comparisons that are correct and
would break if rewritten; there is not one real occurrence of the bug it aims
at.

The motive compounds this: its "Correct" example rewrites `@active = 'true'` to
`@active = true()`, which is a different and usually wrong expression — `= true()`
coerces the left side to boolean, so in XPath 1.0 it is true whenever `@active`
merely exists, and in 2.0 it errors on any value other than true/false. The
check pushes correct code toward a buggier form.

The genuine bug is narrow: a bare boolean *constant* used in a *boolean
context*. `test="'false'"` is a non-empty string, so the branch is always
taken; `test="'true'"` should be `true()`. Narrow the check to an `xsl:if` /
`xsl:when` whose `@test` is exactly `'true'` or `'false'`, and rewrite the
motive to show `test="'false'"` becoming a real condition. Strings used as data
— comparisons, output, parameter defaults — are then left alone. Boolean
`'true'`/`'false'` operands buried inside larger expressions (`'true' and $x`)
wait for the full parser (#228).

maxonfjvipon:
Fixes #409.

`incorrect-use-of-boolean-constants` matched `'true'`/`'false'` as a substring of *any* attribute — `//*[@*[contains(., 'true') …]]` — so it fired on string comparisons (`@active = 'true'`, the correct way to test an XML boolean attribute), literal output (`<td hidden="true"/>`), and string parameter defaults (`select="'false'"`). Over [objectionary/eo](https://github.com/objectionary/eo) it reported **82 defects, 79 of them correct comparisons** and none a real bug. Its motive compounded this by teaching `@active = true()`, a different and usually wrong expression.

Now the check fires only on an `xsl:if`/`xsl:when` whose `@test` is exactly `'true'` or `'false'` — a bare boolean constant used as a condition, where `test="'false'"` is a non-empty string and so always true. Strings used as data (comparisons, output, defaults) are left alone. The motive is rewritten to show the real fix, and the packs now encode the genuine bug plus the comparisons/output that must *not* be flagged.

### DIFF
--- a/src/resources/checks/xpath/incorrect-use-of-boolean-constants.yaml
+++ b/src/resources/checks/xpath/incorrect-use-of-boolean-constants.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //*[@*[(contains(., 'true') and not(contains(., 'true()'))) or (contains(., 'false') and not(contains(., 'false()')))]]
+xpath: "//xsl:*[local-name() = ('if', 'when')][normalize-space(@test) = (\"'true'\", \"'false'\")]"
 severity: warning
-message: "Boolean constants 'true' and 'false' are non-empty strings, not booleans. Use true() and false() instead."
+message: "The test is the string 'true' or 'false', a non-empty string that is always true, not a boolean. Use true() or false() instead."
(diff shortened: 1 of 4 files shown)


maxonfjvipon:
@rultor merge

Issue #414, opened 2026-07-27
not-using-output false-positives on imported library modules


maxonfjvipon:
`not-using-output` is `/xsl:stylesheet[count(xsl:output)=0]`, flagging any
stylesheet without `xsl:output`. A module imported into a pipeline — one that
supplies functions, variables, or templates to an importing stylesheet that
sets the serialization — does not serialize on its own and needs no `xsl:output`
of its own. On [objectionary/eo](https://github.com/objectionary/eo),
`_funcs.xsl` and `_specials.xsl` (pure libraries) are reported.

Fix: exempt a stylesheet that declares no templates (a pure module). A
transform that does have templates but no `xsl:output` remains a legitimate
finding.

maxonfjvipon:
Fixed in 0.0.12 — not-using-output exempts imported modules that have no templates.

Issue #413, opened 2026-07-27
null-output-from-stylesheet false-positives on empty deletion templates


maxonfjvipon:
`null-output-from-stylesheet` reports "The root template contains only variable
declarations and produces no output", but its xpath matches when
`count(*) = count(xsl:variable)`, which is also true when a template has *zero*
children. An empty template is the standard "override to suppress these nodes"
idiom, not a bug.

On [objectionary/eo](https://github.com/objectionary/eo),
`remove-noise.xsl` — a cleanup sheet that deletes listings and metas — uses
`<xsl:template match="/object/listing"><!-- delete it --></xsl:template>` and is
reported twice.

Fix: require `count(xsl:variable) > 0`, matching the message, so an empty
deletion template is not flagged.

maxonfjvipon:
Fixed in 0.0.12 — null-output-from-stylesheet no longer flags an empty node-suppressing template.

Issue #412, opened 2026-07-27
stylesheet-has-no-templates false-positives on function/variable library modules


maxonfjvipon:
`stylesheet-has-no-templates` is `/xsl:stylesheet[count(xsl:template)=0]`, so it
flags any stylesheet without a template. A module that exists only to provide
functions or variables to the stylesheets that `xsl:import` it legitimately has
none. On [objectionary/eo](https://github.com/objectionary/eo), `_funcs.xsl`
(imported by 17 stylesheets) and `_specials.xsl` (by 4) are reported, though
each is a well-formed library.

Fix: only flag a stylesheet that offers nothing — no templates *and* no
functions *and* no variables — so a function or variable library is left alone.

maxonfjvipon:
Fixed in 0.0.12 — stylesheet-has-no-templates exempts library modules with no templates.

PR #415, opened 2026-07-27
Fix four false positives found linting objectionary/eo

### LINKED ISSUE #411: empty-content-in-instructions false-positives on an intentional empty xsl:otherwise
`empty-content-in-instructions` flags an empty `xsl:for-each`, `xsl:if`,
`xsl:otherwise`, or `xsl:when`. But an empty `<xsl:otherwise/>` is the
idiomatic, explicit "for every other case, produce nothing" — and it directly
contradicts `use-choose-without-otherwise`, which asks every `xsl:choose` to
carry an `xsl:otherwise` that is, more often than not, empty. Enable both and
they fight each other.

Found on [objectionary/eo](https://github.com/objectionary/eo), `_funcs.xsl`:
`<xsl:otherwise/>` with the comment "return empty (should not occur if input is
valid UTF-8)" is reported.

Fix: drop `xsl:otherwise` (and `xsl:when`) from the selector. An empty branch
of a choose is a deliberate no-op; only an empty `xsl:if` or `xsl:for-each` is
genuinely useless.

maxonfjvipon:
Running xslint over [objectionary/eo](https://github.com/objectionary/eo) surfaced four checks that misfire on legitimate patterns. Fixes #411, #412, #413, #414.

| check | false positive | fix |
|-------|----------------|-----|
| `empty-content-in-instructions` | an intentional empty `<xsl:otherwise/>` (which `use-choose-without-otherwise` asks for — the two checks fought) | flag only empty `xsl:if` / `xsl:for-each` |
| `stylesheet-has-no-templates` | function/variable **library modules** (`_funcs.xsl` imported by 17 files, `_specials.xsl` by 4) | exempt a stylesheet that has functions or variables |
| `null-output-from-stylesheet` | empty **"delete these nodes"** templates (`remove-noise.xsl`) | require `count(xsl:variable) > 0`, matching the message |
| `not-using-output` | imported modules that never serialize | flag only a stylesheet that has templates |

Verified on eo: the three module/idiom checks drop to **0** and `not-using-output` drops 4 → 2, the two survivors being genuine (templated stylesheets that really lack `xsl:output`). Each fix keeps its genuine case firing (positive packs unchanged) and adds the false-positive case to the negative packs. 376 tests, 100% coverage.

### DIFF
--- a/src/resources/checks/xpath/empty-content-in-instructions.yaml
+++ b/src/resources/checks/xpath/empty-content-in-instructions.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //*[name()='xsl:for-each' or name()='xsl:if' or name()='xsl:otherwise' or name()='xsl:when'][count(node()) = count(text()) and normalize-space() = ""]
+xpath: //*[name()='xsl:for-each' or name()='xsl:if'][count(node()) = count(text()) and normalize-space() = ""]
 severity: warning
-message: An instruction element such as xsl:for-each, xsl:if, or xsl:when has no content. Add content or remove the empty element.
+message: An instruction element such as xsl:for-each or xsl:if has no content. Add content or remove the empty element.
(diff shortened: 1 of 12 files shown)


maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor release, tag=`0.0.12`

PR #421, opened 2026-07-28
Refresh the real-code proof numbers after the false-positive fixes

maxonfjvipon:
The published proof (2,191 findings) predated #407, #409, and #411-#414, so it still counted the false positives those fixes removed. Re-running xslint over the same DocBook/TEI/DITA subsets now gives **2,019 findings across 23 checks, still zero validator false positives** — ~172 fewer, all of them FPs the fixes eliminated:

| check | before → now |
|-------|:---:|
| empty-content-in-instructions | 76 → 1 |
| incorrect-use-of-boolean-constants | 54 → 0 |
| unused-function | 71 → 40 |
| null-output-from-stylesheet | 8 → 0 |

Crucially, the old copy cited `54 uses of the string 'true' as a boolean` — which were exactly the `incorrect-use-of-boolean-constants` false positives, now 0. Keeping it would misrepresent the tool. Lower number, higher signal.


maxonfjvipon:
@rultor merge

Issue #423, opened 2026-07-28
Audit: three checks are not genuine best practices


maxonfjvipon:
A linter earns its keep by enforcing one right way. A check that fires on
code which is actually fine, or that pushes a change which can break the
stylesheet, does the opposite — it trains people to ignore us. Reviewing all
49 checks against that standard, three fail it, and four motives argue their
case dishonestly.

**Drop `unsorted-imports`.** It wants `xsl:import` elements sorted
alphabetically by `@href`. But import order *is* import precedence in XSLT —
a later import overrides an earlier one on a template conflict. Sorting can
silently change which template wins. We must never advise a reorder that can
alter behavior for the sake of tidiness.

**Rework `too-many-small-templates` → `too-many-templates`.** The current rule
(ten or more templates each with fewer than three children) contradicts
`monolithic-design`, and small match templates are idiomatic. The real
readability concern is a file with too many templates at all, regardless of
size. Fire when a stylesheet declares ten or more `xsl:template`, and rename
the check so the name is honest (a breaking change to `--suppress`/config
keys, like the earlier prefix removal).

**Rework `monolithic-design`.** Firing on any single-template stylesheet is
wrong — a small one-template sheet is perfectly good. The smell is one
*undecomposed* template that should have fanned out with `apply-templates`.
Fire only when the sheet has exactly one template or function and that one
holds more than thirty `xsl:*` elements — below `function-template-complexity`
(fifty) so the two do not double-report.

**Fix four misleading motives.** `missing-id` and `not-using-output` claim
technical necessity (`@id` "required", output "implementation-defined"); both
claims are false — `@id` does nothing on a standalone stylesheet and the
default output method is defined by the spec. The honest rationale is
consistency: either every stylesheet carries it or none does. And
`use-double-slash`/`starts-with-double-slash` blame a "full document scan"; a
`//` match pattern is not evaluated as a scan — the real cost is an over-broad,
under-specific pattern.

Together these form one readability band: one template over thirty elements is
monolithic, any template over fifty is too complex, ten or more templates is
too many, and two to nine is the healthy middle.

maxonfjvipon:
Refined the size/count band after review — splitting by node type instead of
sharing a threshold, so the two size checks can never double-report:

- `too-many-small-templates` → `too-many-templates`: `count(//xsl:template) >= 10`.
- `function-template-complexity` → `function-complexity`: `//xsl:function[count(.//xsl:*) > 50]`
  (functions only — a template no longer counts here).
[...]

PR #424, opened 2026-07-28
Drop the unsorted-imports check

maxonfjvipon:
`unsorted-imports` warned when `xsl:import` elements were not sorted
alphabetically by `@href`. That advice is unsafe: in XSLT, import order is
import precedence — a module imported later overrides one imported earlier when
their templates conflict. Sorting the imports can therefore silently change
which template wins.

A linter has to be trustworthy about the fixes it suggests. A stylistic reorder
that can alter behavior fails that bar, so the check comes out entirely — rule,
motive, and both test packs.

Part of the check audit in #423.


maxonfjvipon:
@rultor merge

PR #425, opened 2026-07-28
Rework the size and count checks into a coherent band

maxonfjvipon:
The three checks that judge how work is split across a stylesheet were pulling
against each other. `too-many-small-templates` wanted fewer templates while
`monolithic-design` wanted more, and `function-template-complexity` graded
templates and functions on one scale even though they decompose in opposite
ways — a big template fans out with `apply-templates`, a big function splits
into smaller functions.

This separates them by node type so they can never double-report the same node,
and gives each an honest name:

- `function-template-complexity` → `function-complexity`: `//xsl:function[count(.//xsl:*) > 50]`.
- `monolithic-design` → `oversized-template`: `//xsl:template[count(.//xsl:*) > 100]`, so a
  small one-template sheet is fine and only an undecomposed hundred-element template fires.
- `too-many-small-templates` → `too-many-templates`: `count(//xsl:template) >= 10`, size-independent.

A function over fifty elements is too complex, a template over a hundred is
oversized, ten or more templates is too many, and two to nine is the healthy
middle. The renames are breaking for `--suppress` and config keys.

Part of the check audit in #423.


maxonfjvipon:
@rultor merge

PR #426, opened 2026-07-28
Rewrite four misleading check motives

maxonfjvipon:
A check is only worth following if its stated reason holds up. Four motives
argue their case with a claim that is not true, so this rewrites them around
the honest rationale.

`missing-id-in-stylesheet` and `not-using-output` both claim technical
necessity. Neither is: `id` does nothing on a standalone stylesheet, and the
default output method is fixed by the spec (`xml`, or `html` for an `<html>`
root), not left implementation-defined. They are consistency rules — either
every stylesheet declares the thing or none does — and now say so.

`use-double-slash` and `starts-with-double-slash` blame a "full document scan".
A match pattern is tested against a node, not walked as a query, so no scan
happens. The real problem is an over-broad, under-specific pattern; a leading
`//` is redundant outright, since an unanchored pattern already matches at any
depth.

Also fixes the copy-pasted "Missing version in stylesheet" title sitting atop
the `missing-id` motive. Part of the check audit in #423.


maxonfjvipon:
@rultor merge

PR #429, opened 2026-07-28
Refresh the real-code proof after the check audit

maxonfjvipon:
Re-ran xslint over the same 70 core stylesheets from DocBook-XSL, TEI, and
DITA-OT with the audited check set. The honest number is now **1,974 findings
across 22 different checks**, down from 2,019 across 23: the audit dropped
`unsorted-imports` and narrowed `function-complexity` to functions only, while
the size/count reshuffle roughly balanced out. The three cited specifics — 106
`xsl:choose` blocks without `xsl:otherwise`, 67 unused named templates, 40
uncalled functions — are unchanged.

Updates the README and the docs landing page, and bumps the stale `0.0.12`
install example to `0.0.13`. Follows the audit in #423.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Refresh the real-code proof after the check audit: 1,974 findings across 22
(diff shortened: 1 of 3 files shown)


maxonfjvipon:
@rultor merge

Issue #430, opened 2026-07-28
Reconsider not-using-schema-types as a precise per-binding check


maxonfjvipon:
The `not-using-schema-types` check was removed in the #423 audit. It fired on a
2.0/3.0 stylesheet that used no `xs:` type *anywhere*, which was wrong on two
counts: the underlying idea (declare schema types) is an optional-by-design
preference rather than a universal best practice, and the rule did not even
measure that idea — "mentions `xs:` at least once" passes a sheet that types one
of fifty variables and is silenced by sprinkling in a single token type.

There may still be a worthwhile check hiding here, but a precise one: flag an
individual `xsl:variable`, `xsl:param`, or `xsl:function` that has no `as` type
in a 2.0/3.0 stylesheet — per binding, not per file. That is honest about what
it asks for and points at the exact place to fix.

Two open questions before reviving it: whether requiring `as` everywhere is a
best practice we want to insist on at all (XSLT makes typing optional, and much
good code is untyped), and how noisy a per-binding rule would be on real
stylesheets. Parking this until we decide.

PR #431, opened 2026-07-28
Drop the not-using-schema-types check

maxonfjvipon:
`not-using-schema-types` fired on any XSLT 2.0/3.0 stylesheet that used no
`xs:` type anywhere in an attribute value. That bar is arbitrary — it passes a
sheet that types one of fifty variables and is silenced by adding a single
token type — and it flags perfectly valid untyped stylesheets, which XSLT
allows by design. The underlying idea (declare schema types) is an optional
preference, not a universal best practice, and the rule did not even measure
that idea.

Removed entirely: rule, motive, and all four packs. A precise per-binding
successor — flag an individual untyped `xsl:variable`/`param`/`function` — is
parked for later in #430. Two end-to-end fixtures used the check as a sample
defect; they now assert the `not-using-output` defect they still produce.

Closes the last item of the audit in #423.


maxonfjvipon:
@rultor merge

PR #432, opened 2026-07-28
Make the double-slash check messages honest

maxonfjvipon:
Follow-up to the motive fixes in #426. Those corrected the *motives* for
`starts-with-double-slash` and `use-double-slash` but left the one-line
*messages* still claiming a "document scan" — which a match pattern never
performs, so the message contradicted its own motive.

The messages now say what is actually true: a leading `//` is redundant because
an unanchored pattern already matches at any depth, and an inner `//` matches
at any depth and is broader than intended. Updates the two end-to-end
expectations that pin the `starts-with-double-slash` message. Part of #423.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


maxonfjvipon:
@rultor merge

Issue #436, opened 2026-07-28
New check: count(X) compared to 0 should be an existence test


maxonfjvipon:
`count($x) &gt; 0` and `count($x) = 0` force the processor to walk and tally the
whole sequence only to ask whether it is non-empty. The intent is existence, and
XPath says it directly: `$x` (in a boolean context), `exists($x)`, or `empty($x)`
/ `not($x)`. The idiomatic form is clearer and lets the engine stop at the first
node instead of counting all of them.

Proposed check `count-compared-to-zero`: flag `count(...)` compared with `0`
(`&gt; 0`, `= 0`, `!= 0`, `&lt;= 0`, `&gt;= 1`) in a `test`/`select`. Fix: drop the
`count()` and use the sequence directly, `exists()`, or `empty()`.

Widespread: DocBook-XSL 130, TEI 20, DITA-OT 6, objectionary/eo 1. Example:

- eo `parse/stars-to-tuples.xsl:51` — `test="count($elems)&gt;0"` → `test="exists($elems)"`

Found while bringing eo to the full check set (#423).

maxonfjvipon:
**Autofix: yes, safe `--fix`.** The rewrites are exact equivalences:
`count(X) > 0` / `!= 0` / `>= 1` → `exists(X)`, and `count(X) = 0` / `< 1` /
`<= 0` → `empty(X)`. It is a single-span replacement of the sub-expression, so
it fits the current fix engine — extract the `count()` argument by balancing
parens over the masked token stream (the same technique `use-node-set-extension`
already uses to unwrap its argument), no full-fidelity parser needed. Ship the
`fix` as a plain safe fix, not a suggestion.

maxonfjvipon:
Implemented in #441 (+ #442) and merged to master; ships in the next release (0.0.15).

PR #441, opened 2026-07-28
Add the count-compared-to-zero check

maxonfjvipon:
Implements #436. Flags `count(X)` compared with `0` to test existence and
rewrites it to the direct form:

```xsl
<xsl:if test="count($items) &gt; 0">   →   <xsl:if test="exists($items)">
<xsl:if test="count($items) = 0">       →   <xsl:if test="empty($items)">
```

Covers `> 0`, `!= 0`, `>= 1` → `exists()` and `= 0`, `<= 0`, `< 1` → `empty()`.
A genuine count (`count(X) > 1`, `count(X) = 5`) is left alone, a custom
`my:count()` is not matched, and a `count(` inside a string literal is masked.

It is a code-based linter (`src/count-linter.js`) modeled on `node-set-linter`:
it masks string and comment spans with the lexer, finds the call, and balances
parentheses on the blanked text, so nested arguments and literals are handled.
The `--fix` is safe (`exists`/`empty` are exact equivalences of the count
comparison); as `fixer.js` verifies the span before applying, a comparison
written with `&gt;`/`&lt;` in the source is skipped rather than mis-edited.

100% covered; README, CLAUDE.md, and the docs site updated.


maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor stop

maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor merge

PR #442, opened 2026-07-28
Add --fix-dry-run and report-drop tests for count-compared-to-zero

maxonfjvipon:
Follow-up to #441: the count fix landed with a single end-to-end test. This
adds the other two the fixable checks all carry — `--fix-dry-run` leaves the
file untouched, and a fixed `count-compared-to-zero` defect is dropped from the
report — so the fix is exercised from more than one angle.

Relates to #436.


maxonfjvipon:
@rultor merge

Issue #437, opened 2026-07-28
New check: string-length(X) compared to 0 should test emptiness directly


maxonfjvipon:
`string-length($x) &gt; 0` and `string-length($x) = 0` compute the full length of
a string to ask whether it is non-empty. The direct form says it better:
`$x != ''` (or `normalize-space($x)` when whitespace should not count as
content, which is usually the intent). Same family as the `count()`-vs-existence
antipattern.

Proposed check `string-length-compared-to-zero`: flag `string-length(...)`
compared with `0`. Fix: `!= ''` / `= ''`, or `normalize-space(...)` when blank
should read as empty.

Widespread: TEI 70, DocBook-XSL 41, DITA-OT 2, objectionary/eo 2. Example:

- eo `atoms-table.xsl:19` — `string-length(@atom) &gt; 0` → `@atom != ''`

Found while bringing eo to the full check set (#423).

maxonfjvipon:
**Autofix: yes.** `string-length(X) > 0` → `X != ''` and `string-length(X) = 0`
→ `X = ''` — single-span, extract `X` by paren-balancing like #436. Safe
`--fix` when `X` is a singleton (`@a`, `$v`), which is the common case. Two
caveats to gate on: if `X` can be a sequence the string-length form and the
`!= ''` form differ, so treat non-singleton operands conservatively (skip or
suggestion); and the `normalize-space(X)` rewrite is a separate
`--fix-suggestions`, since folding whitespace to empty is a behavior change.

maxonfjvipon:
Implemented in #443 and merged to master; ships in the next release (0.0.15).

PR #443, opened 2026-07-28
Add the string-length-compared-to-zero check

maxonfjvipon:
Implements #437, the sibling of `count-compared-to-zero`. Flags
`string-length(X)` compared with `0` to test emptiness and rewrites it to the
direct form:

```xsl
<xsl:if test="string-length(@name) &gt; 0">   →   <xsl:if test="@name != ''">
<xsl:if test="string-length(@name) = 0">       →   <xsl:if test="@name = ''">
```

It covers both operand orders (`0 &lt; string-length(@n)`) and works inside
larger boolean expressions. A genuine length check (`string-length(X) &gt; 1`)
is left alone, and `my:string-length()` / a literal are not matched.

The one wrinkle over the count check: `X != ''` is only precedence-safe when
`X` is a single operand, so the fix is attached only for simple arguments;
a union like `string-length($a | $b)` is still **reported** but left for a
human to rewrite. `normalize-space(X)` (which folds whitespace to empty) is a
different test, so it is never applied automatically.

Along the way the shared lexer helpers `masked`/`closes` move to
`src/expressions.js`, now used by the count, node-set, and string-length
linters (three copies down to one). 100% covered, with detection packs for both
operand orders / embedding / the union no-fix case, and `--fix`, `--fix-dry-run`,
and report-drop fixer tests. README, CLAUDE.md, and the docs site updated.


maxonfjvipon:
@rultor merge

Issue #435, opened 2026-07-28
New check: a select that starts with // scans the whole document


maxonfjvipon:
`starts-with-double-slash` and `use-double-slash` flag `//` only in a `match`
pattern. A `//` at the *start of a `select`* is a different and worse problem:
unlike a match pattern (where a leading `//` is merely redundant), a select like
`//o[@local]` is evaluated as `/descendant-or-self::node()/o`, so it walks the
entire document from the root every time the expression runs — quadratic when
it sits inside a template applied per node. It is also frequently a latent bug:
the author usually meant "descendants of the current node" (`.//o`) or a
specific path, not "every `o` in the document".

Proposed check `select-scans-from-root` (or extend the double-slash family to
`select`): flag a `select` whose XPath begins with `//`. Anchor it to a context
(`.//o`) or a specific path.

Real and widespread — leading-`//` selects: objectionary/eo ~7, DocBook-XSL 49,
TEI 72, DITA-OT 21. Examples in eo:

- `parse/cti-adds-errors.xsl:15` — `select="//o[@base='Φ.cti']"`
- `parse/resolve-local-names.xsl:100` — `select="//o[@local]"`
- `parse/fragile-dispatch.xsl:15` — `select="//o[@method and not(@fragile) and ...]"`

Found while bringing eo to the full check set (#423).

maxonfjvipon:
**Autofix: no — leave fix-less.** The correct rewrite depends on intent —
`.//o` (descendants of the current node) versus a specific path — so there is no
deterministic, semantics-preserving edit. A blind `//` → `.//` would be a
guess that changes scope and can break a case that genuinely wanted the whole
document. Ship the check as a warning with no `fix`.

maxonfjvipon:
Implemented in #444 and merged to master; ships in the next release (0.0.15).

PR #444, opened 2026-07-28
Add the select-starts-with-double-slash check

maxonfjvipon:
Implements #435. A `select` whose XPath begins with `//` is evaluated as
`/descendant-or-self::node()/…`, so it walks the whole document from the root —
and inside a template applied per node, once per node. Unlike a `match` pattern,
where a leading `//` is merely redundant, in a `select` it is a real, repeated
scan, and often a latent bug where `.//x` was meant.

```xsl
<xsl:for-each select="//item">   →   <xsl:for-each select=".//item">
```

Warning only, with no `fix`: the correct anchor (`.//` vs a specific path)
depends on intent, so there is no deterministic edit. An inner `//`
(`items//item`), a `//` inside a string literal, or one reached through a
variable (`$root//item`) is left alone. A plain declarative XPath rule —
`//*[starts-with(normalize-space(@select), '//')]` — so no linter code, just the
rule, its motive, and positive/negative packs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add the `select-starts-with-double-slash` check: a `select` whose XPath begins
+  with `//` scans the whole document from the root every time it runs — a real,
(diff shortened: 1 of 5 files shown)


maxonfjvipon:
@rultor merge

Issue #439, opened 2026-07-28
New check: comparing name()/local-name() to a string should be a node test


maxonfjvipon:
Testing an element's identity by string-comparing its name is slower and more
fragile than a node test:

```xsl
<xsl:if test="name() = 'div'">          →  <xsl:if test="self::div">
<xsl:apply-templates select="*[local-name() = 'label']"/>
                                        →  <xsl:apply-templates select="*:label"/>  (2.0)
```

`name() = 'div'` compares the lexical QName, so it silently depends on the
prefix the source happens to use and breaks under a different but equivalent
namespace binding; `self::div` matches by expanded name. `local-name() = 'x'`
throws the namespace away entirely — the intent is "any namespace", which 2.0
writes as `self::*:x` / `*:x`. A node test also lets the engine match without
building and comparing strings.

Proposed check `name-comparison-instead-of-node-test`: flag `name()` or
`local-name()` compared (`=`/`!=`) with a string literal in a `test`/`select`.
Fix: a node test — `self::name` for `name()`, `self::*:name` (or `*:name`) for
`local-name()`.

Widespread: TEI 334 (184 `name()` + 150 `local-name()`), DocBook-XSL 138,
DITA-OT 20, objectionary/eo ~13. Examples: eo `name()='object'`, TEI
`local-name()='label'`.

Found while auditing eo and the proof-section projects (#423).

maxonfjvipon:
**Autofix: yes, `--fix-suggestions` tier.** `name() = 'x'` → `self::x` and
`local-name() = 'x'` → `self::*:x` (or `*:x`) are single-span rewrites the
current engine can apply. Mark the `fix` `suggestion: true`: it changes matching
from the lexical QName to the expanded name — which is the whole point of the
correction, but it is a behavior change, and `self::x` vs `self::*:x` is a
choice — so `--fix` should leave it and only `--fix-suggestions` apply it.

maxonfjvipon:
Implemented in #445 and merged to master; ships in the next release (0.0.15).

PR #445, opened 2026-07-28
Add the name-compared-to-string check

maxonfjvipon:
Implements #439. Comparing an element's name to a string is prefix-fragile and
slower than a node test:

```xsl
<xsl:if test="name() = 'div'">                        →  <xsl:if test="self::div">
<xsl:apply-templates select="*[local-name() = 'x']"/>  →  ...select="*[self::*:x]"...
```

`name() = 'div'` compares the lexical QName (depends on the source's prefix and
breaks under an equivalent binding); `self::div` matches by expanded name.
`local-name() = 'x'` discards the namespace, which 2.0 writes as the wildcard
`self::*:x`.

It flags both operand orders (`'div' = name()`), works inside larger boolean
expressions, and only considers a call over the current node (`name()` /
`name(.)`) — `name($x)` and custom functions are left alone. The rewrite is a
**`--fix-suggestions`**, since it shifts lexical- to expanded-name matching (the
intended correction, but a behavior change); `!= 'x'` becomes `not(self::x)`,
the `local-name()` → `*:x` form is offered only for 2.0/3.0, and an invalid name
(spaces) is reported without a fix.

Code-based linter sharing `src/expressions.js`. 100% covered; detection packs
for both orders / 1.0 version-gating / invalid name / skip cases, and
`--fix`/`--fix-suggestions`/report-drop fixer tests. README, CLAUDE.md, docs
updated.


maxonfjvipon:
@rultor merge

Issue #440, opened 2026-07-28
New check: translate() case-folding should be lower-case()/upper-case() in 2.0+


maxonfjvipon:
The classic XSLT 1.0 way to change case is a `translate()` against the full
alphabet:

```xsl
translate(@ident, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz')
```

XSLT 2.0 has `lower-case()` and `upper-case()`, which are shorter, Unicode-aware
(the `translate` trick only folds ASCII), and say what they mean. In a 2.0/3.0
stylesheet the `translate` alphabet spell-out is an anachronism.

Proposed check `translate-for-case`: in a stylesheet whose `version` is 2.0 or
3.0, flag a `translate(...)` whose second and third arguments are the ASCII
alphabet in the two cases. Fix: `lower-case(...)` or `upper-case(...)`. Must be
version-gated — in 1.0 the `translate` form is the only option, so it is not a
defect there.

Present in TEI (2.0) 29 times, e.g.
`translate(@ident,'ABCDEFGHIJKLMNOPQRSTUVWXYZ','abcdefghijklmnopqrstuvwxyz')`.
(DocBook-XSL's few are 1.0, correctly not a defect.)

Found while auditing eo and the proof-section projects (#423).

maxonfjvipon:
**Autofix: yes, `--fix-suggestions` tier.** `translate(X,'A..Z','a..z')` →
`lower-case(X)` and the reverse → `upper-case(X)` — single-span; extract the
first argument by paren-balancing. Mark it `suggestion: true`: `lower-case()`
folds all of Unicode whereas the `translate` trick folded only ASCII, so it is a
(usually desirable) behavior change, not a strict equivalence.

maxonfjvipon:
Implemented in #446 and merged to master; ships in the next release (0.0.15).

PR #446, opened 2026-07-28
Add the translate-for-case check

maxonfjvipon:
Implements #440. In an XSLT 2.0/3.0 stylesheet, the 1.0-era way to fold case —

```xsl
translate(@id, 'ABCDEFGHIJKLMNOPQRSTUVWXYZ', 'abcdefghijklmnopqrstuvwxyz')
```

— is flagged and a `--fix-suggestions` rewrites it to `lower-case(@id)` (or
`upper-case(@id)` for the reverse alphabets).

**Version-gated**: it fires only in 2.0/3.0, since in 1.0 the `translate` form
is the only option and not a defect. It is a **suggestion** because
`lower-case()`/`upper-case()` fold all of Unicode while the `translate` trick
only folds ASCII — a usually-desirable behavior change, not an exact
equivalent. Arguments are split on the depth-zero commas of the blanked text, so
`translate(concat(@a, @b), …)` is handled; a `translate` with any other pair of
arguments is left alone.

Code-based linter sharing `src/expressions.js`. 100% covered, with detection
packs (both directions, the 1.0 no-op, non-alphabet/2-arg/embedded), and
`--fix`/`--fix-suggestions`/report-drop fixer tests. README, CLAUDE.md, docs
updated.


maxonfjvipon:
@rultor merge

Issue #438, opened 2026-07-28
New check: static xsl:attribute on a literal element should be a literal attribute


maxonfjvipon:
The attribute analog of `not-creating-element-correctly`. When an
`xsl:attribute` with a static name is a child of a literal result element and
its value is a single `xsl:value-of` or plain text, it can be written as a
literal attribute with an attribute value template:

```xsl
<fo:block>
  <xsl:attribute name="line"><xsl:value-of select="@l"/></xsl:attribute>
  ...
```
becomes
```xsl
<fo:block line="{@l}">
  ...
```

The AVT form is shorter, keeps the attribute next to the element it belongs to,
and is the idiomatic spelling. `xsl:attribute` should be reserved for a computed
name or a value that needs instructions (e.g. `xsl:choose`) an AVT cannot hold.

Proposed check `not-creating-attribute-correctly`: flag `xsl:attribute` with a
static name whose only content is a single `xsl:value-of` or text, when its
parent is a literal result element. Fix: fold it into the parent as an AVT (or a
static attribute when the value is literal).

Widespread: DocBook-XSL 739, TEI 447, objectionary/eo ~17, DITA-OT 9. Example:

- eo `parse/cti-adds-errors.xsl:33` — `<xsl:attribute name="line">…</xsl:attribute>`

Found while bringing eo to the full check set (#423).

maxonfjvipon:
**Autofix: yes, but blocked on the full-fidelity parser (#228).** The fix moves
the attribute onto the parent element's start tag as an AVT
(`<xsl:attribute name="line"><xsl:value-of select="@l"/></xsl:attribute>` on
`<fo:block>` → `<fo:block line="{@l}">`). That is a structural, multi-span edit
— delete the child node and edit the parent's start tag — which the current
single-span engine cannot express. Attach the `fix` (a safe one, since the AVT
is equivalent) once #228 gives exact per-node spans.

maxonfjvipon:
Implemented in #447 (detection; the structural AVT fix waits on #228) and merged to master; ships in the next release (0.0.15).

PR #447, opened 2026-07-28
Add the not-creating-attribute-correctly check

maxonfjvipon:
Implements #438, the attribute analog of `not-creating-element-correctly`. When
an `xsl:attribute` with a static name sits on a literal result element and its
value is simple — a single `xsl:value-of`, plain text, or empty — it can be
written inline:

```xsl
<fo:block>
  <xsl:attribute name="line"><xsl:value-of select="@l"/></xsl:attribute>
  <xsl:attribute name="kind">head</xsl:attribute>
```
→
```xsl
<fo:block line="{@l}" kind="head">
```

**Warning only, no fix**: the inline rewrite moves the attribute onto the
parent's start tag — a structural, multi-span edit that waits on the
full-fidelity parser (#228), as noted there. `xsl:element`/`xsl:copy` parents,
computed (AVT) names, and values needing an `xsl:choose` are left alone, since
`xsl:attribute` earns its place there.

A plain declarative XPath rule (no linter code), with positive/negative packs
and its motive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add the `not-creating-attribute-correctly` check: an `xsl:attribute` with a
+  static name on a literal result element, whose value is simple (a single
+  `xsl:value-of`, text, or empty), can be written inline as a literal attribute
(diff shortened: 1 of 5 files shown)


maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor merge

PR #462, opened 2026-07-29
Make starts-with-double-slash fixable

### LINKED ISSUE #455: Make starts-with-double-slash fixable: drop the redundant leading // (safe --fix)
`starts-with-double-slash` flags an `xsl:template` whose `@match` begins with `//`. In a match pattern the leading `//` is redundant — a pattern is already unanchored, so `//para` and `para` select the same nodes — so removing it is deterministic and semantics-preserving. That makes it a **safe `--fix`**, not a suggestion, and the only safe fix in this batch.

Implement as a declarative fixer in `src/fixers.js` (a `node => fix` builder over the `xsl:template`): a single span at `@match` with `value: 'match="//'`, `replacement: 'match="'`, deleting the two leading slashes. Verify-before-apply skips a single-quoted or reshaped attribute rather than cutting the wrong span.

No blocker. Add a `test/resources/fix/starts-with-double-slash.{xsl,fixed.xsl}` pair and the usual `--fix` / `--fix-dry-run` / report-drop cases in `test/fixer.test.js`.

maxonfjvipon:
`starts-with-double-slash` flags an `xsl:template` whose `@match` begins with `//`. In a match pattern that leading `//` is redundant — a pattern is already unanchored, so `match="//para"` and `match="para"` select the same nodes. This wires up the correction as a **safe `--fix`** (not a suggestion, since it preserves the matched nodes).

It is a declarative fixer in `src/fixers.js` — a single span over `@match` that drops the two leading slashes, verified against the original source before applying:

```js
const startsWithDoubleSlash = function(node) {
  const match = node.getAttributeNode('match')
  return {
    line: match.lineNumber,
    col: match.columnNumber - match.name.length - 1,
    value: `${match.name}="${match.value}"`,
    replacement: `${match.name}="${match.value.slice(2)}"`,
  }
}
```

A committed `.xsl`/`.fixed.xsl` pair plus three `test/fixer.test.js` cases cover it: `--fix` rewrites in place, `--fix-dry-run` leaves the file untouched, and the fixed defect drops from the report. Docs and CHANGELOG updated; coverage stays at 100%.

Closes #455.


maxonfjvipon:
@rultor merge

PR #463, opened 2026-07-29
Make incorrect-use-of-boolean-constants fixable

### LINKED ISSUE #456: Make incorrect-use-of-boolean-constants fixable: 'true'/'false' to true()/false()
`incorrect-use-of-boolean-constants` flags an `xsl:if`/`xsl:when` whose `@test` is the string literal `'true'` or `'false'` — a non-empty string, hence always true, never the boolean the author meant. The fix replaces `'true'` with `true()` and `'false'` with `false()`.

A **`--fix-suggestions`**, not a plain fix: for `'false'` it changes the truth value (always-true becomes false), which is exactly the bug but still a behavior change.

Implement as a declarative fixer over the `@test` attribute value — a single span, like `output-method-xml` — reading whether the literal is `true` or `false` to build the replacement. The rule matches under `normalize-space`, so tolerate surrounding whitespace when locating the literal. No blocker. Tests: a fixture pair plus plain `--fix` leaves it / `--fix-suggestions` applies / report-drop.

maxonfjvipon:
`incorrect-use-of-boolean-constants` flags an `xsl:if`/`xsl:when` whose `@test` is the string literal `'true'` or `'false'` — a non-empty string, hence always true, never the boolean the author meant. This wires up the correction: `'true'` becomes `true()`, `'false'` becomes `false()`.

A **`--fix-suggestions`**, not a plain fix: for `'false'` it changes the truth value (always-true becomes false), which is exactly the bug but still a behavior change. It is a declarative fixer over the `@test` attribute value, a single span verified against the original source.

A committed `.xsl`/`.fixed.xsl` pair (carrying both a `'true'` and a `'false'` test) plus three `test/fixer.test.js` cases cover it: plain `--fix` leaves it untouched, `--fix-suggestions` rewrites both, and the fixed defect drops from the report. Docs and CHANGELOG updated; coverage stays at 100%.

Closes #456.


maxonfjvipon:
@rultor merge

PR #464, opened 2026-07-29
Make select-starts-with-double-slash fixable

### LINKED ISSUE #457: Make select-starts-with-double-slash fixable: anchor // as .//
`select-starts-with-double-slash` flags a `@select` that begins with `//`, rescanning the whole document from the root on every evaluation. The mechanical anchor is `.//` (descendants of the context node): `//x` becomes `.//x`.

A **`--fix-suggestions`** — it changes behavior (absolute to relative) and `.//` is one of several valid anchors, since a specific path is often the better correction. Implement as a declarative fixer inserting `.` before the leading `//` of `@select`, a single span. The rule matches under `normalize-space`, so skip any leading whitespace when locating the `//`.

No blocker. Tests: a fixture pair plus plain `--fix` leaves it / `--fix-suggestions` applies / report-drop.

maxonfjvipon:
`select-starts-with-double-slash` flags a `@select` that begins with `//`, rescanning the whole document from the root on every evaluation. This wires up the mechanical anchor: `select="//title"` becomes `select=".//title"`, scanning the context node's descendants instead.

A **`--fix-suggestions`** — it changes behaviour (absolute to relative) and `.//` is one of several valid anchors, since a specific path is often the better correction. It is a declarative fixer that inserts `.` before the leading `//` of `@select`, a single span; it locates the `//` by index, so a value with leading whitespace (which `normalize-space` still flags) is handled.

A committed `.xsl`/`.fixed.xsl` pair plus three `test/fixer.test.js` cases cover it: plain `--fix` leaves it untouched, `--fix-suggestions` anchors it, and the fixed defect drops from the report. Docs and CHANGELOG updated; coverage stays at 100%.

Closes #457.


maxonfjvipon:
@rultor merge

PR #465, opened 2026-07-29
Make confusing-variable-and-node fixable

### LINKED ISSUE #458: Make confusing-variable-and-node fixable: prepend $ to the bare name
`confusing-variable-and-node` flags an `xsl:apply-templates` whose `@select` is a bare name equal to, or beginning with, a declared variable's name — a forgotten `$`. Because the rule fires only when the name sits at the **start** of `@select` (`@select = $var/@name`, or `starts-with(@select, concat($var/@name, '/'))`), the fix is a clean single-span insert: `select="foo…"` becomes `select="$foo…"`.

A **`--fix-suggestions`** — prepending `$` assumes the author meant the variable rather than a child element named `foo`, a semantic change.

No blocker: no expression parsing is needed, since the name is anchored at the start of the attribute value, so a declarative fixer inserting `$` after the opening quote of `@select` suffices. Tests: a fixture pair plus plain `--fix` leaves it / `--fix-suggestions` applies / report-drop.

maxonfjvipon:
`confusing-variable-and-node` flags an `xsl:apply-templates` whose `@select` is a bare name equal to (or beginning with) a declared variable's name — a forgotten `$`. This wires up the correction: `<xsl:apply-templates select="items"/>` becomes `select="$items"`.

Because the rule fires only when the name sits at the **start** of `@select`, the fix is a clean single-span insert of `$` after the opening quote — no expression parsing needed. A **`--fix-suggestions`**, since prepending `$` assumes the author meant the variable rather than a child element named `items`.

A committed `.xsl`/`.fixed.xsl` pair plus three `test/fixer.test.js` cases cover it: plain `--fix` leaves it untouched, `--fix-suggestions` prepends the `$`, and the fixed defect drops from the report. Docs and CHANGELOG updated; coverage stays at 100%.

Closes #458.


maxonfjvipon:
@rultor merge

PR #466, opened 2026-07-29
Make text-outside-xsl-text fixable

### LINKED ISSUE #459: Make text-outside-xsl-text fixable: wrap the literal text in xsl:text
`text-outside-xsl-text` flags literal, non-whitespace text sitting directly inside an `xsl:` instruction element. The fix wraps it: `TEXT` becomes `<xsl:text>TEXT</xsl:text>`.

A **`--fix-suggestions`**, since `xsl:text` changes whitespace handling. Two wrinkles beyond the attribute-value fixers:

- The fix targets a **text node**, not an attribute — `value: TEXT`, `replacement: <xsl:text>TEXT</xsl:text>` at the text node's source span. `@xmldom/xmldom` exposes a start position on text nodes; verify-before-apply skips text carrying entities, whose raw and parsed forms differ.
- One flagged element can hold **several** non-whitespace text nodes (text around child elements), so the check must emit one fix per text node — disjoint spans — rather than one fix per element.

No hard blocker, but it needs the linter to attach fixes at text-node granularity, a small extension of the current element-oriented model. Tests: a fixture pair (including a multi-text-node element) plus plain `--fix` leaves it / `--fix-suggestions` applies / report-drop.

maxonfjvipon:
`text-outside-xsl-text` flags loose literal text sitting directly inside an `xsl:` instruction. This wires up the wrap: `<xsl:if test="a">hello</xsl:if>` becomes `<xsl:if test="a"><xsl:text>hello</xsl:text></xsl:if>`.

A **`--fix-suggestions`** — a stylistic rewrite that inserts an element. Two points beyond the attribute-value fixers:

- The fix targets a **text node**, not an attribute — it uses the text node's own source span (`@xmldom/xmldom` exposes line/column on text nodes), verified against the original source before applying.
- It is offered **only when the instruction holds a single non-whitespace text node**. Text on both sides of a child element (`x<foo/>y`) would need several wraps that the one-span fixer cannot make, so those stay report-only — the same conditional-fix pattern `mode-or-priority-without-match` uses.

Tests: a committed `.xsl`/`.fixed.xsl` pair plus three `test/fixer.test.js` cases (plain `--fix` leaves it, `--fix-suggestions` wraps, report-drop), and a mixed-content xpath pack that pins the no-fix path. Docs and CHANGELOG updated; coverage stays at 100%.

Closes #459.


maxonfjvipon:
@rultor merge

Issue #468, opened 2026-07-29
Cross-tree consistency checks on the import graph: version, output, mode, key (from #210)


maxonfjvipon:
Spun out of #210. Heuristic consistency checks over a resolved import tree:

- a `version` mismatch between a stylesheet and a module it imports;
- conflicting `xsl:output` (method/encoding) across the tree;
- an `xsl:apply-templates` `mode="m"` with no template in the tree carrying that mode;
- an undefined `xsl:key` (a `key('k', …)` naming no declared key) or an unused one.

Each is a name/attribute check scoped to the resolved tree rather than the global union. Guardrail: assert only when the relevant modules are all present in the corpus — an href resolving outside the linted set means "external, assume fine" — so a subset corpus never false-positives.

Moderate. Depends on the import-graph builder from #210.

## Fix (implement in the same PR)

One of these four is auto-fixable, the rest are not — implement each check with its fix (or explicitly without one) in the same PR:

- **`unused xsl:key`** → **safe `--fix`**: delete the unused `xsl:key` element (its own line, reconstructed from the element the way `redundant-import` deletes a duplicate import — `xsl:key` is a leaf with `name`/`match`/`use` attributes). Semantics-preserving, since a key that no `key(...)` calls is dead.
- **undefined `xsl:key`**, **`mode` with no template**, **conflicting `xsl:output`**, **version mismatch** → **no fix**: each needs content the tool cannot author (a key/template definition) or a judgment call (which output/version wins), so they stay report-only.

Test the `unused xsl:key` fix with a committed `.xsl`/`.fixed.xsl` pair plus `--fix` / `--fix-dry-run` / report-drop cases, as with the other fixable checks.

maxonfjvipon:
The `#210` dependency in this issue is satisfied — #210 is closed and
`src/import-graph.js` ships `importsOf` and `graphOf`, which the corpus checks
already walk. So the import-graph half is no longer a blocker here.

maxonfjvipon:
Held until Phase 4 of #644 lands, not rejected.

A new check written against the current substrate inherits the whole blind-spot cross-product the existing text-matching checks are parked for — it would arrive already owing the same fixes, and would have to be migrated onto the CST from #228 alongside them. Adding one now grows the row rather than the coverage.

Once the eleven text-matching checks are on the tree, a new check starts there and this becomes worth writing.

maxonfjvipon:
[...]

Issue #467, opened 2026-07-29
Detect redundant and self imports on the import graph (from #210)


maxonfjvipon:
Spun out of #210. On the import graph built there, flag two shapes:

- **self-import** — a stylesheet that `xsl:import`s or `xsl:include`s itself (directly or transitively back to itself); an XSLT error.
- **redundant import** — the same resolved href imported/included more than once within one stylesheet's own top-level list.

Precedence subtlety to respect: a *direct* import at a different precedence than a *transitive* one is not truly redundant (it deliberately changes override order), so scope "redundant" to exact-duplicate hrefs in a single stylesheet's own import/include list first, and treat transitive duplication cautiously to avoid false positives.

Moderate. Depends on the import-graph builder from #210.

Issue #452, opened 2026-07-29
Add an unresolved-reference validator for in-corpus symbols that are never imported


maxonfjvipon:
xslint today accepts a stylesheet that a real XSLT processor rejects at compile time. When `main.xsl` calls `my:even(...)` but never imports the module that declares `my:even`, Saxon raises `XPST0017` (unknown function); a `call-template` naming an unimported template is `XTSE0650`. xslint reports nothing. The `xpath-validator` deliberately lets unknown functions and prefixes through — `src/xpath.js:19` notes that "an unknown prefix must not be mistaken for a syntax error" — and the corpus checks pool every declaration and every reference across the whole corpus without ever resolving `xsl:import`/`xsl:include` (`src/corpus-linter.js:161` and `:198`). So the one reference that is a provable error slips through alongside the many that are not.

This surfaced while splitting `unused-function` into `unused-function` and `unreachable-function` (#175): a test fixture had `main.xsl` call a function it never imported, which the linter happily accepted because nothing resolves the import graph.

The proposal is a new validator, `unresolved-reference`, sitting as the natural third validation layer after well-formedness and XPath syntax — well-formed, then parseable, then resolvable. It flags a reference only when both conditions hold: the symbol is declared in another file that is part of the linted corpus, and the referencing file cannot reach that file through its import/include graph.

The narrow framing is what keeps it quiet. If the symbol is declared nowhere in the corpus, the check stays silent — that is the ordinary "unknown function" case (an EXSLT or Saxon extension, a function the host registers, or simply a module the user did not hand to xslint), and guessing there would demand a complete inventory of every built-in and extension namespace. By firing only on the provable in-corpus case — you call `my:even`, which is declared right here in `funcs.xsl`, but `main.xsl` never imports it — the check needs no such inventory and produces near-zero false positives. It is the mirror image of `unused-function`'s pooling: instead of "referenced somewhere, so fine," it is "declared here, but the import is missing."

This depends on #210. Deciding "cannot reach through imports" means resolving `href` attributes, building the transitive import/include tree, and knowing which declarations are visible in which file — exactly the import-graph machinery #210 introduces. It should be built on top of that, never before it: a version that guesses about built-ins instead of resolving the graph would be the false-positive machine the narrow framing exists to avoid.

Add it as `src/resources/checks/validation/unresolved-reference.yaml` with code in a new `src/unresolved-validator.js`, gated on the #210 import-graph resolver, and keep it silent whenever the symbol is absent from the corpus.

[...]

Issue #210, opened 2026-06-18
import-graph builder + circular-import check (first slice of the import-graph epic)


maxonfjvipon:
The precision upgrade ("Semantic B") over the global-union corpus from #106
(PR #207): parse `xsl:import`/`xsl:include` hrefs, resolve them, and build the
dependency graph. This also sharpens the duplicate-name checks (judge within
an import tree, not globally).

### Checks unlocked by the graph

1. **Circular import/include** — A imports B imports A; an XSLT error.
2. **Redundant import** — the same href imported/included more than once in a
   tree; **self-import**.
3. **`xsl:apply-imports` / `xsl:next-match` with nothing to override** in the
   tree.
4. **Cross-tree consistency** (heuristic): conflicting `xsl:output`
   (method/encoding), `version` mismatch between a stylesheet and what it
   imports, `apply-templates mode="m"` with no template carrying that mode,
   and undefined/unused `xsl:key`.

Larger than the symbol-table and filesystem issues; depends on building the
graph infrastructure first.

maxonfjvipon:
## Re-scoping to a first slice: the import-graph builder + `circular-import`

This issue bundled a foundation with ~7 checks spanning trivial to hard. Splitting it: **#210 now covers just the graph builder and the one check that is both high-value and false-positive-proof.** The rest is spun out:

- #467 — redundant / self-import (moderate)
- #468 — cross-tree consistency: version, output, mode, key (moderate)
- #469 — apply-imports / precedence conflicts (hard, needs #228)

The graph also unlocks #452 (`unresolved-reference`) and later lets the `unused-*` / duplicate-name corpus checks sharpen from global-union to within-tree.

### 1. `src/import-graph.js` — the foundation

`graphOf(corpus)` builds an adjacency over the corpus files:
- for each `{file, xsl}`, take the top-level `xsl:import`/`xsl:include` children, read `@href`, resolve it against `path.dirname(file)` (`path.resolve` + normalize), and match against the set of corpus file paths;
- return, per file, its resolved dependency edges (tagged import|include) and its unresolved hrefs.
- **Guardrail:** an href resolving to no corpus file is "external, assume fine" — recorded, never a defect. This keeps a subset corpus (linting a subdirectory whose imports point outside it) from false-positiving. No new I/O — the corpus already holds every file with its path.

### 2. `circular-import` — `src/import-linter.js`

[...]

PR #470, opened 2026-07-29
Add circular-import check and the import-graph foundation

maxonfjvipon:
The first slice of the import-graph epic (#210, re-scoped): the graph builder plus the one check that is both high-value and false-positive-proof.

**`src/import-graph.js`** — `graphOf(corpus)` resolves every top-level `xsl:import`/`xsl:include` `@href` against the importing file's own directory and matches it to the corpus, returning the dependency edges (each carrying its declaring element). No file is read — the corpus already holds every parsed stylesheet with its path. An href that resolves outside the corpus is external and yields no edge, so a stylesheet importing a library it was not handed alongside is never mistaken for a dependency.

**`circular-import`** (`src/import-linter.js`) — DFS the graph; an edge sits on a cycle exactly when its target can reach back to its source. Each such `xsl:import`/`xsl:include` is reported at its element, severity **error** (the XSLT static error `XTSE0210`), covering both a mutual `A ↔ B` cycle and a self-import.

Why this check first: a cycle needs *every* edge in it to resolve within the corpus, so an external/unresolved href can never be part of one — `circular-import` **cannot false-positive on a partial corpus** (linting a subdirectory whose imports point outside it). It also exercises the whole graph builder, de-risking the follow-ups (#452 unresolved-reference, #467 redundant/self-import, #468 cross-tree consistency) and a later within-tree sharpening of the `unused-*` checks.

Four packs pin it: a mutual cycle (both imports flagged), a self-import, an acyclic include (not flagged), and an external import outside the corpus (not flagged — the guardrail). New `src/import-graph.js` + `src/import-linter.js` follow the code-based-check layout; docs, README, and CHANGELOG updated; coverage stays at 100%.

Part of #210.


maxonfjvipon:
@rultor merge

PR #471, opened 2026-07-29
Add redundant-import check

### LINKED ISSUE #467: Detect redundant and self imports on the import graph (from #210)

maxonfjvipon:
The next slice on the import graph (#210): flag the same module `xsl:import`ed or `xsl:include`d more than once within one stylesheet, **and fix it**. Closes #467.

**`redundant-import`** (a warning) reports the second and later import of the same resolved target in one stylesheet's own list. Hrefs resolve against the importing file's directory, so `lib/util.xsl` and `./lib/util.xsl` count as one, and the target need not be a corpus file — importing the same external library twice is redundant too.

**Safe `--fix`** — it deletes the duplicate import's whole line, reconstructed from the element (indentation, the self-closing tag with its single `href`, and the trailing newline), so the fixer verifies before applying and safe-skips an oddly formatted, single-quoted, or non-self-closing import. Deleting a duplicate is semantics-preserving (the module stays imported by the first reference), so it is a safe fix, not a suggestion.

Scope note: **self-import and cross-file cycles are already caught by `circular-import`** (a self-loop is a cycle), so #467 is deliberately just the duplicate-within-one-file case — no double-reporting.

`src/import-graph.js` grows an `importsOf(corpus)` that lists every import with its resolved target (external included); `graphOf` is refactored to build on it (in-corpus edges only, for `circular-import`). `src/import-linter.js` now owns both checks.

Packs pin the check (a duplicate import flagged, two distinct imports not flagged, plus a suppression test); a committed `.xsl`/`.fixed.xsl` pair and three `test/fixer.test.js` cases pin the fix (`--fix` deletes, `--fix-dry-run` leaves, report-drop). The existing circular packs are unaffected. Docs, README, and CHANGELOG updated; coverage stays at 100%.


maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor stop

maxonfjvipon:
@rultor merge

PR #472, opened 2026-07-29
Test that --fix removes four of five duplicate imports

maxonfjvipon:
A hardening test for the `redundant-import` `--fix` (shipped in #471): a stylesheet that imports the same module **five** times should keep exactly one after `--fix`, with the other four removed.

It confirms the multi-duplicate case works — the four deletions are disjoint whole-line removals applied end-to-start against the original offsets, so they never overlap or corrupt the source, and the result is clean (no blank lines, no mis-indentation):

```
5 × <xsl:import href="common.xsl"/>   →   1 × <xsl:import href="common.xsl"/>
```

A committed `.xsl`/`.fixed.xsl` pair plus one `test/fixer.test.js` case. No production change — coverage stays at 100%.

### DIFF
--- a/test/fixer.test.js
+++ b/test/fixer.test.js
@@ -113,6 +113,14 @@ describe('fixer', function() {
       fixture('redundant-import.fixed.xsl'),
     )
   })
+  it('should delete four of five duplicate imports with --fix', function() {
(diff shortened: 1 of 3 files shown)


maxonfjvipon:
@rultor merge

PR #474, opened 2026-07-29
Make test files table-driven to cut duplication

### LINKED ISSUE #473: Reduce test duplication with table-driven cases
Many tests differ only in a couple of fixture names or arguments, yet each is written out in full. `test/fixer.test.js` is the clearest case — the safe-fix, dry-run, report-drop, and suggestion cases repeat the same five-line body per check:

```js
it('should delete a redundant import with --fix', function() {
  const file = scratch(fixture('redundant-import.xsl'))
  runXslint(['--fix', file])
  assert.equal(fs.readFileSync(file, 'utf-8'), fixture('redundant-import.fixed.xsl'))
})
it('should delete four of five duplicate imports with --fix', function() {
  const file = scratch(fixture('redundant-import-many.xsl'))
  runXslint(['--fix', file])
  assert.equal(fs.readFileSync(file, 'utf-8'), fixture('redundant-import-many.fixed.xsl'))
})
```

These should collapse into a table plus one generator, so adding a case is one object, not a new block:

```js
[
  {name: 'should delete a redundant import with --fix', before: 'redundant-import.xsl', after: 'redundant-import.fixed.xsl'},
  {name: 'should delete four of five duplicate imports with --fix', before: 'redundant-import-many.xsl', after: 'redundant-import-many.fixed.xsl'},
].forEach(({name, before, after}) => {
  it(name, function() {
    const file = scratch(fixture(before))
    runXslint(['--fix', file])
    assert.equal(fs.readFileSync(file, 'utf-8'), fixture(after))
  })
})
```

Sweep the whole test suite by the same principle: where several `it(...)` blocks share a body and differ only in data, replace them with a data array and one `it` generator, grouped by the shape of the body (safe `--fix`, `--fix-dry-run`, report-drop, `--fix-suggestions`, etc.). The pack-driven linter tests already do this. Genuinely unique tests (the `fixed(...)` unit cases, the "announce how many" cases, one-off assertions) stay as they are — do not force-fit them.

Coverage and behaviour must not change; this is a pure readability refactor.

maxonfjvipon:
Closes #473. Where several `it` blocks shared a body and differed only in data, they now come from a data array plus one `it` generator — so adding a case is a new row, not a new block.

- **`fixer.test.js`** — the bulk: three tables (`APPLIED`, `UNCHANGED`, `DROPPED`) drive the per-check `--fix`/`--fix-dry-run`/report-drop/`--fix-suggestions` cases from `{name, flag, before/after/sheet/check}` rows. The `fixed(...)` unit cases, the "announce" cases, and the parent-axis one-off stay as they are.
- **`config.test.js`** — a `{name, content, field, expected}` table for the "write `.xslint.yml`, resolve a field" cases; the no-file/explicit-file/walk-up ones stay unique.
- **`xsl-validator.test.js`** — `KEPT`, `REPORTED`, and `EXPAND` tables; the console-leak, mixed-corpus, and suppression cases stay unique.
- **`directives.test.js`** — `PARSED`, `SUPPRESSED`, `UNUSED` tables.
- **`logger.test.js`** — a table for the level cases; the unknown-level warning stays unique.
- **`tokens.test.js`** — `NUMBERS` and `SCANS` tables, a shared `PIECES`/`generated()` for the two round-trip properties, and one accidental duplicate test removed.

The pack-driven linter tests were already table-driven and are untouched. Genuinely distinct tests are left alone. Behaviour and coverage are unchanged (100%); the only test-count change is the removed duplicate. CLAUDE.md now records the convention so new tests join an existing table where one fits.


maxonfjvipon:
@rultor merge

PR #475, opened 2026-07-29
Add when-or-otherwise-outside-choose check

### LINKED ISSUE #368: Add a when-or-otherwise-outside-choose check
A new lint check: `xsl:when` and `xsl:otherwise` are valid only as children of `xsl:choose`. One with any other parent is invalid XSLT that XML well-formedness will not catch.

Incorrect:
```xsl
<xsl:template match="/">
  <xsl:when test="x"/>
</xsl:template>
```

A declarative XPath rule, **report-only** (no unambiguous fix — wrapping in `xsl:choose` or converting to `xsl:if` is a judgement call): `//xsl:when[not(parent::xsl:choose)] | //xsl:otherwise[not(parent::xsl:choose)]`. Follows the Adding-a-New-Rule pattern.

maxonfjvipon:
Closes #368. `xsl:when` and `xsl:otherwise` mean something only inside an `xsl:choose`; with any other parent they are invalid XSLT that XML well-formedness never catches (the element is well-formed on its own).

A declarative XPath rule, severity **error**:

```
//xsl:when[not(parent::xsl:choose)] | //xsl:otherwise[not(parent::xsl:choose)]
```

**Report-only** — the right correction (wrap the stray branch in an `xsl:choose`, or rewrite an `xsl:when` as an `xsl:if`) is a judgement call, not one mechanical edit; the motive says so.

Also, on the way: CLAUDE.md now states explicitly that **a fixable check must ship and test its fix in the same change** (fixture pair + `fixer.test.js` rows), never detection-then-defer, with structural-only fixes deferred to #228. And I corrected a stale line that claimed every `src/fixers.js` fix is a suggestion — `starts-with-double-slash` is a safe fix.

One YAML + one pack + one motive, auto-discovered; docs regenerated (58 checks); coverage stays at 100%.


maxonfjvipon:
@rultor merge

PR #477, opened 2026-07-29
Add empty-choose check

### LINKED ISSUE #374: Add an empty-choose check
A new lint check: an `xsl:choose` with no `xsl:when` branch is degenerate — it is either dead structure or an `xsl:choose` that should be removed (its `xsl:otherwise`, if any, always runs).

Incorrect:
```xsl
<xsl:choose>
  <xsl:otherwise>...</xsl:otherwise>
</xsl:choose>
```

A declarative XPath rule, **report-only** (unwrapping the sole `xsl:otherwise` is structural): `//xsl:choose[not(xsl:when)]`. Follows the Adding-a-New-Rule pattern.

maxonfjvipon:
Closes #374. An `xsl:choose` with no `xsl:when` is degenerate — XSLT requires at least one, and a `choose` holding only an `xsl:otherwise` always runs its single branch. Declarative XPath rule, severity **error**: `//xsl:choose[not(xsl:when)]`. Report-only (add a `when`, or drop the `choose` — a judgement call).

Note: a no-`when` choose already trips `use-single-option-for-choose` (`count(*)=1`) on the same node, so the two co-fire; both packs record it via the harness's length-3 positions. That overlap points at a small follow-up — `use-single-option-for-choose` mis-fires on a lone `xsl:otherwise` (its "use xsl:if" advice makes no sense with no condition), so it could be tightened to `count(xsl:when)=1`, which would make the two checks disjoint. Out of scope here.

One YAML + pack + motive; docs regenerated (60 checks); 452 tests; coverage 100%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add the `empty-choose` check: an `xsl:choose` with no `xsl:when` is
+  degenerate — XSLT requires at least one, and a `choose` holding only an
+  `xsl:otherwise` is a wrapper that always runs. Report-only, since the fix
+  (add a `when`, or drop the `choose`) is a judgement call (#374).
+
 - Add the `otherwise-not-last` check: an `xsl:otherwise` followed by another
   branch is invalid — it is the default and must come last in its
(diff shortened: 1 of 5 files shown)


maxonfjvipon:
@rultor merge

PR #478, opened 2026-07-29
Add sort-not-first check

### LINKED ISSUE #372: Add a sort-not-first check
A new lint check: `xsl:sort` must be the first child of its `xsl:for-each` or `xsl:apply-templates`. One that follows other content is invalid and silently ignored by some processors.

Incorrect:
```xsl
<xsl:for-each select="item">
  <xsl:value-of select="."/>
  <xsl:sort select="@n"/>
</xsl:for-each>
```

A declarative XPath rule, **report-only** (the fix would move the `xsl:sort` to the front — a structural reorder that needs the #228 parser groundwork): `//xsl:sort[preceding-sibling::*[not(self::xsl:sort)]]`. Follows the Adding-a-New-Rule pattern.

maxonfjvipon:
Closes #372. An `xsl:sort` that follows other content in its `xsl:for-each`/`xsl:apply-templates` is invalid and silently ignored by some processors. Declarative XPath rule, severity error: `//xsl:sort[preceding-sibling::*[not(self::xsl:sort)]]`. Report-only, since moving it to the front is structural (#228). One YAML + pack + motive; 454 tests, coverage 100%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add the `sort-not-first` check: an `xsl:sort` that follows other content in
+  its `xsl:for-each` or `xsl:apply-templates` is invalid and silently ignored
+  by some processors. Report-only, since moving it to the front is structural
+  (#372).
+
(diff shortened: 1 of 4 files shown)


maxonfjvipon:
@rultor merge

PR #479, opened 2026-07-29
Add param-after-content check

### LINKED ISSUE #367: Add a param-after-content check
A new lint check: `xsl:param` must be declared before any other content in an `xsl:template` or `xsl:function`. A param that follows real content is invalid XSLT that XML well-formedness will not catch.

Incorrect:
```xsl
<xsl:template name="t">
  <xsl:variable name="v" select="1"/>
  <xsl:param name="p"/>
</xsl:template>
```

A declarative XPath rule is enough, and this one is **report-only**: the fix would be to move the param ahead of the content, a structural reorder that needs the full-fidelity parser groundwork (#228), so no fix for now. Detection, roughly: `//(xsl:template | xsl:function)/xsl:param[preceding-sibling::*[not(self::xsl:param) and not(self::xsl:context-item)]]` (the XSLT 3.0 `xsl:context-item` is also allowed before params). Follows the Adding-a-New-Rule pattern.

maxonfjvipon:
Closes #367. An `xsl:param` that follows real content in its `xsl:template` or `xsl:function` is invalid XSLT that XML well-formedness never catches — only other `xsl:param` siblings and an `xsl:context-item` may precede it. Declarative XPath rule, severity **error**. Report-only, since moving the param to the front is structural (#228).

Two existing packs (`template-parameter-uses-not-in-their-scope`, `with-param-use-with-another-prefix`) had a param declared after content only as incidental scaffolding; I reordered them so the param comes first, leaving their own checks (and positions) intact — cleaner fixtures, no co-fire.

One YAML + pack + motive; docs regenerated (62 checks); 456 tests; coverage 100%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Add the `param-after-content` check: an `xsl:param` that follows real
+  content in its `xsl:template` or `xsl:function` is invalid XSLT that XML
+  well-formedness never catches (only other params and an `xsl:context-item`
+  may precede it). Report-only, since moving the param up is structural
+  (#367).
+
 - Add the `sort-not-first` check: an `xsl:sort` that follows other content in
   its `xsl:for-each` or `xsl:apply-templates` is invalid and silently ignored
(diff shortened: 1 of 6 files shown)


maxonfjvipon:
@rultor merge

PR #481, opened 2026-07-29
Make the degenerate-choose checks disjoint

### LINKED ISSUE #480: Tighten use-single-option-for-choose so it stops firing on a no-when choose
`use-single-option-for-choose` matches `//xsl:choose[count(*) = 1]` — any one-child `xsl:choose`. That includes a `choose` holding only an `xsl:otherwise` (no `xsl:when`), where its advice — "replace with `xsl:if`" — makes no sense, since there is no condition. That same no-`when` choose is now flagged, correctly, by `empty-choose` (#374), so the two overlap and give contradictory guidance on one node.

Tighten `use-single-option-for-choose` to the case it actually means — a single `xsl:when` and no `xsl:otherwise`, which collapses cleanly to an `xsl:if`:

```
//xsl:choose[count(xsl:when) = 1 and not(xsl:otherwise)]
```

This is a pure narrowing: it keeps firing on `<choose><when>…</when></choose>` (the real single-option case) and stops firing on `<choose><otherwise>…</otherwise></choose>` (left to `empty-choose`). The two checks become disjoint, each giving correct advice. Update the pack (its current input is a lone-otherwise choose, which no longer matches) and the motive.

maxonfjvipon:
Closes #480. Resolves the overlap between the degenerate-`xsl:choose` checks by making them a **disjoint partition**, each firing once with the right advice:

| Check | Fires on | Advice |
|-------|----------|--------|
| `empty-choose` | no `xsl:when` | add a `when`, or drop the `choose` |
| `use-single-option-for-choose` | exactly one `xsl:when`, no `xsl:otherwise` | use `xsl:if` |
| `use-choose-without-otherwise` | two or more `xsl:when`, no `xsl:otherwise` | add an `xsl:otherwise` |

Before, `use-single-option-for-choose` matched `count(*) = 1` — any one-child choose, **including a lone `xsl:otherwise`**, where "use `xsl:if`" is nonsense (no condition) and it double-reported with `empty-choose` (#374). Tightening it to `count(xsl:when) = 1 and not(xsl:otherwise)` fixed that but left it a subset of `use-choose-without-otherwise` (which matched *any* no-otherwise choose), so I also tightened that to `>= 2` whens. Now every degenerate choose is caught by exactly one check.

Rules, packs, and motives updated for both; a lone-`when` choose is now correctly routed to `use-single-option` (better advice) instead of "add otherwise". 456 tests, coverage 100%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,14 @@ publication date only; detailed notes begin with the Unreleased section.
 
 ## Unreleased
 
+- Make the degenerate-`xsl:choose` checks disjoint, so each fires once with
+  the right advice: `empty-choose` on no `xsl:when`, `use-single-option-for-
+  choose` on exactly one `xsl:when` and no `xsl:otherwise` (use `xsl:if`), and
+  `use-choose-without-otherwise` on two or more `xsl:when` and no
+  `xsl:otherwise` (add one). Previously `use-single-option-for-choose` fired
+  on any one-child `choose` — including a lone `xsl:otherwise`, where its
+  advice was nonsense and it overlapped with `empty-choose` (#480).
+
 - Add the `param-after-content` check: an `xsl:param` that follows real
   content in its `xsl:template` or `xsl:function` is invalid XSLT that XML
   well-formedness never catches (only other params and an `xsl:context-item`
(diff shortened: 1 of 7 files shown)


maxonfjvipon:
@rultor merge

PR #482, opened 2026-07-29
Add fixable redundant-double-negation check

### LINKED ISSUE #366: Add a fixable redundant-double-negation check
A new lint check for `not(not(x))`, a redundant double negation — equivalent to `boolean(x)`, and to `x` in a boolean context.

Incorrect:
```xsl
<xsl:if test="not(not(@enabled))">
```
Correct:
```xsl
<xsl:if test="boolean(@enabled)">
```

Implement it as a **fixable** code-based check (not a declarative XPath rule): mask string and comment spans, find a `not(` immediately followed by another `not(`, balance both parentheses, and rewrite `not(not(x))` to `boolean(x)`. That rewrite is always equivalent, so the fix can be a safe (auto) one rather than a suggestion. Part of #334.

Follows the Adding-a-New-Rule pattern.

maxonfjvipon:
Closes #366. `not(not(x))` is a redundant double negation — exactly `boolean(x)` — reported in any `@test`/`@select`, **with a safe `--fix`** that rewrites it to `boolean(x)` (always equivalent, so `--fix`, not a suggestion).

A code-based linter alongside the count/node-set/string-length ones: it masks string and comment spans and balances parentheses via `src/expressions.js`, matching an outer unprefixed `not(...)` whose only content is an inner `not(...)`. Left alone: an outer `not(...)` with more than a lone inner `not(` (`not(not(x) and y)`), a custom `my:not(...)`, and (defensively) an unbalanced call.

Ships with its fix per the CLAUDE.md rule: a `test/resources/fix/redundant-double-negation.{xsl,fixed.xsl}` pair and three `test/fixer.test.js` rows (`--fix` rewrites, `--fix-dry-run` leaves it, report-drop), plus packs pinning the positives and every skip branch. Docs regenerated (63 checks); 461 tests; coverage 100%.


maxonfjvipon:
@rultor merge

PR #483, opened 2026-07-29
Add redundant-boolean-call check, keep double-negation fix coherent

maxonfjvipon:
A whole `@test` wrapped in `boolean()` is redundant — a test already coerces its value to a boolean, so `test="boolean(@x)"` and `test="@x"` behave identically. This adds a `redundant-boolean-call` linter (`src/redundant-boolean-call-linter.js`) that flags it and strips the wrapper with a safe `--fix`. A `boolean()` that is only part of a larger expression (`a = boolean(b)`) is left alone, since there the coercion can change what the comparison means.

That surfaced an incoherence with the already-merged `redundant-double-negation` check, which rewrote `not(not(x))` to `boolean(x)` everywhere. In a whole `@test` that left a `boolean(x)` the new check would immediately re-flag. So the double-negation fix is now context-aware:

```js
const bare = attribute.nodeName === 'test' &&
  attribute.nodeValue.trim() === value
// ...
replacement: bare ? argument : `boolean(${argument})`,
```

A whole-`@test` `not(not(x))` collapses straight to `x`; everywhere else it still becomes `boolean(x)`. The two checks now agree on the canonical form.

The packs exercise more than the textbook case: the construct buried in a predicate and an `and`/`or` operand, three or more occurrences in one expression, and the similar-looking negatives that must not fire. `CLAUDE.md` gains a note to test those hard cases going forward. Docs, README, and the architecture notes are updated in the same change.


maxonfjvipon:
@rultor merge

PR #484, opened 2026-07-29
Add a fixable predicate-position-literal check

maxonfjvipon:
A predicate that holds a single number is, by definition, a test on the context position: `item[1]` means `item[position() = 1]`, and `item[last()]` means `item[position() = last()]`. Spelling the `position() =` out adds nothing but length.

The new `predicate-position-linter` (`src/predicate-position-linter.js`) flags the long form in any XPath or pattern attribute — in either operand order — and shortens it with a safe `--fix`:

```
item[position() = 1]                 -> item[1]
row[1 = position()]                  -> row[1]
cell[position() = last()]            -> cell[last()]
```

It tokenizes, matches brackets through a stack so a nested predicate is judged on its own contents (and a `[` inside a string literal is never seen), then reduces the significant tokens to a short signature. A predicate that does more is left alone — `[position() = 1 and @x]`, `[position() > 1]`, and the value comparison `[position() eq 1]` all fail to match, since there the `position()` is load-bearing.

The pack exercises all four operand forms, a nested predicate where only the inner one fires (`wrap[kid[position() = 1]]`), three occurrences in one expression, the look-alike negatives, and an unbalanced bracket. Docs, README, and the architecture notes are updated in the same change.


maxonfjvipon:
@rultor merge

PR #512, opened 2026-07-29
Extract a shared defect() builder and metaOf() loader across the code-based linters

### LINKED ISSUE #500: Extract a shared defect() builder and metaOf() loader across the code-based linters
The twelve code-based format linters (`count`, `string-length`, `node-set`, `name`, `translate`, `redundant-double-negation`, `redundant-boolean-call`, `predicate-position`, `xpath-axis`, `namespace`, `result-namespace`, `xpath-format`) open with an identical preamble —

```js
const CHECK = "<name>"
const META = yaml.parsedFromFile(path.join(__dirname, "resources", "checks", "format", `${CHECK}.yaml`))
const names = [CHECK]
```

— the identical suppression guard `if (!suppressions.some(sup => CHECK.includes(sup)))`, and an identical ~13-line defect-push block (`pos = attribute.columnNumber + 1 + offset`, the `defects.push({name, severity, message, file, line, pos, fix})` shape). `import-linter.js` already has the extractable helpers the other eleven lack (`meta(check)`, `defect(check, file, node)`, `suppressed(check, suppressions)`).

Extracting a shared `defect()`/`metaOf()` into a helper module (they already all import `expressions.js`/`helpers.js`) collapses most of the boilerplate **without** any linter importing another, preserving the stated topology. CLAUDE.md's own rule (every convention machine-enforced) argues for it: today the copies can silently drift. Found during a code-quality audit.

maxonfjvipon:
Closes #500.

The twelve code-based format linters each opened with an identical metadata preamble, repeated the suppression guard `if (!suppressions.some(sup => CHECK.includes(sup)))`, and carried a ~13-line defect-push block that differed only in the finder and the fix. This introduces `src/checks.js`:

- `metaOf(check)` — loads the check's `checks/format/*.yaml`.
- `suppressed(check, suppressions)` — the shared suppression guard.
- `defect(check, meta, file, node, offset, fix)` — assembles the standard defect, computing `pos = node.columnNumber + 1 + offset` and anchoring the fix (`{value, replacement, suggestion?}`) at that same line/column.

The ten attribute/expression-offset linters (count, string-length, name, translate, node-set, xpath-format, double-negation, boolean-call, predicate-position, axis) use all three. The DOM linters (`namespace`, `result-namespace`) and `import-linter` keep their own node positioning — their column formula differs — but share `metaOf`/`suppressed`. `import-linter` already had local `meta`/`defect`/`suppressed`; those are now the shared ones.

No behavior change: the defect object shape and key order are preserved, so the full suite, 100% coverage on all four metrics, and ESLint stay green. Net −161 lines. `CLAUDE.md` documents `src/checks.js` as the building block a new linter reuses instead of re-deriving the boilerplate.


maxonfjvipon:
@rultor merge

PR #513, opened 2026-07-29
Unify the count and string-length comparison scan

### LINKED ISSUE #501: count-linter and string-length-linter are near-total clones; unify the count-comparison finder
`count-linter.js` and `string-length-linter.js` are structurally the same file: byte-identical `TAIL`/`HEAD`/`FLIP` regex tables, and their `collapses()`/`empty()` semantic tables are the same operator/zero branch tree differing only in the returned token (`exists`/`empty` vs `false`/`true`). They could share a single `count-comparison` finder parameterized by a small result map, removing a whole duplicated file's worth of logic that must currently be kept in sync by hand. Found during a code-quality audit.

maxonfjvipon:
Closes #501.

`count-linter.js` and `string-length-linter.js` were structurally the same file: byte-identical `TAIL`/`HEAD`/`FLIP` tables and the same "find `f(x)` compared with `0`/`1` in either operand order" scan, differing only in the function name and how a matched comparison becomes a replacement.

This extracts `src/comparisons.js`:

```js
comparedToZero(expression, name, decide)
```

which runs the whole scan (blank literals, balance parentheses, match the forward `f(x) op N` and reversed `N op f(x)` forms) and hands each comparison to a per-check `decide(operator, zero, argument, blanked)` classifier. `count` maps an existence test to `exists(...)`/`empty(...)`; `string-length` maps an emptiness test to `X = ''`/`X != ''`, reporting without a fix when the argument is not a simple operand.

No behavior change — the full suite, 100% coverage on all four metrics, and ESLint stay green. Net −102 lines across the two linters. `CLAUDE.md` documents `src/comparisons.js`.


maxonfjvipon:
@rultor merge

PR #514, opened 2026-07-29
Scan the full XPath/pattern attribute set in count/string-length/name/translate/double-negation

### LINKED ISSUE #502: count/string-length/name/translate/double-negation under-scan: they miss @use, @value, @group-by, and pattern attributes
These five linters scan only `//@test | //@select`, while `xpath-axis-linter` and `predicate-position-linter` correctly scan the full 17-attribute set (adding `use`, `value`, `key`, `group-by`, `group-adjacent`, and the pattern attributes `match`/`count`/`from`/…). So a `count(x) > 0`, `string-length(x) = 0`, `name() = 'x'`, alphabet `translate()`, or `not(not(x))` sitting in an `xsl:key/@use`, an `xsl:sort/@select` is caught, but in `@value` (xsl:number), `@group-by`, or a `match`/`from` **pattern** is silently missed. There is no comment explaining the divergence, so it looks accidental. Suggest reusing the shared `ATTRIBUTES`/`SELECTOR` constant (see the dedupe issue). Found during a code-quality audit.

maxonfjvipon:
Closes #502.

Five linters — `count-compared-to-zero`, `string-length-compared-to-zero`, `name-compared-to-string`, `translate-for-case`, `redundant-double-negation` — scanned only `//@test | //@select`, while `unabbreviated-axis` and `predicate-position-literal` already scanned the full 17-attribute XPath/pattern set. So a `count(x) > 0` in an `xsl:key/@use`, a `name() = 'x'` in a `@group-by`, or a `not(not(@a))` in a `match` pattern went silently unreported — false negatives.

The five now scan the same set. That set was duplicated verbatim in the axis and predicate linters; it is now the single source `src/attributes.js` (`ATTRIBUTES`/`SELECTOR`), imported by all seven. `node-set` (`@select` only, where the extension lives) and `redundant-boolean-call` (a whole `@test`) keep their deliberately narrow, documented scope.

Each of the five packs gains a `match`-pattern occurrence proving the wider scan (e.g. `match="section[count(para) > 0]"` → `exists(para)`). Full suite, 100% coverage on all four metrics, ESLint, and xcop stay green. `CLAUDE.md` documents `src/attributes.js` and the scan-scope rule.


maxonfjvipon:
@rultor merge

PR #515, opened 2026-07-29
Derive the CHECKS name list from the linters registry

### LINKED ISSUE #503: CHECKS list in xslint.js is hand-maintained and can desync from the linters
`src/xslint.js` builds `CHECKS` (used for `--suppress` validation and config globbing) by hand-spreading each linter's `names` export across two lines. Every new linter must be added in three places (import, `LINTERS`, and the `CHECKS` spread) and the last is easy to forget — a missing entry silently breaks suppression/config for that check with no test catching it. Since each linter already exports `names`, `CHECKS` should be derived by flat-mapping `names` over `LINTERS`/`EXPRESSION_LINTERS` (plus the validators). Found during a code-quality audit.

maxonfjvipon:
Closes #503.

`CHECKS` — the list `--suppress` and `.xslint.yml` globs match against — was a hand-maintained spread of every linter's `names`, parallel to the `LINTERS` array of functions. Adding a linter meant editing both, and a forgotten `CHECKS` entry silently broke suppression/config for that check with no test catching it.

`LINTERS` and `EXPRESSION_LINTERS` are now arrays of `{run, checks}` entries, pairing each linter with its own `names` export:

```js
const CHECKS = [
  ...VALIDATOR_CHECKS,
  ...LINTERS.flatMap((stage) => stage.checks),
  ...EXPRESSION_LINTERS.flatMap((stage) => stage.checks),
]
```

A linter and its suppression names can no longer drift apart — the list is computed, not restated. No behavior change: the full suite, 100% coverage, and ESLint stay green. `CLAUDE.md` reflects the derived wiring.


maxonfjvipon:
@rultor merge

PR #517, opened 2026-07-29
xcop-check the inline pack-fixture XSL in CI

### LINKED ISSUE #505: Inline pack-fixture XSL is never checked by xcop in CI
`test/xcop.test.js` gates every assertion on `cmdAvailable('xcop')`, but no test-running job installs Ruby/xcop (`grunt.yml`, `coverage.yml`, `daily.yml` set up Node only). The dedicated `xcop.yml` runs xcop over committed `**/*.xsl`/`**/*.xml`, but pack fixtures store their XSL **inline** in YAML (`input:`), so that formatting is neither seen by `xcop.yml` nor exercised by `xcop.test.js` in CI — it is only ever checked on a developer machine that happens to have xcop. CLAUDE.md overstates this ("validated by the xcop workflow"). Fix: install xcop in the test job (or extract pack XSL to temp files the `xcop.yml` glob covers). Related to #170 (motive XSL). Found during a CI audit.

maxonfjvipon:
Closes #505.

`test/xcop.test.js` extracts every pack's inline `input:` XSL and runs xcop over it, but it gates on `cmdAvailable('xcop')` and no test-running job installed xcop — so pack-fixture formatting was only ever checked on a developer machine that happened to have it (`grunt.yml`/`coverage.yml` set up Node only; `xcop.yml` set up xcop but not Node). CLAUDE.md overstated the coverage.

`xcop.yml` now also sets up Node and runs `npx mocha test/xcop.test.js`, alongside the existing `xcop` CLI pass over committed `.xsl`. Since that job already installs xcop, `cmdAvailable` returns true there and the 183 fixture assertions actually run.

Enabling the check surfaced two latent problems it had been silently hiding:

- **`stylesheet-has-no-templates`** carried an incidental unused `xmlns:xs` that xcop canonicalises away — dropped it (the defect stays at `[2,1]`).
- **`redundant-namespace-declarations`** *must* carry an unused namespace — the very thing the check flags — so its fixture can never be canonical XML. Excluded via a new `UNFORMATTED` list, mirroring how the `fix/` fixtures are excluded from the `xcop.yml` CLI pass.

Full suite, 100% coverage, and ESLint stay green. CLAUDE.md now describes the CI wiring and the exclusion.


maxonfjvipon:
@rultor merge

Issue #518, opened 2026-07-30
count/string-length --fix silently skips comparisons that use < or > (entity-encoded in XML)


maxonfjvipon:
`count-linter`/`string-length-linter` build `fix.value` from the *decoded* attribute value (`attribute.nodeValue`), e.g. `count(x) > 0`. But the source is XML, where `>`/`<` are written `&gt;`/`&lt;`, and `src/fixer.js:56` verifies the fix by matching that decoded value against the *raw* file content:

```js
const matches = content.substr(start, defect.fix.value.length) === defect.fix.value
```

So for `test="count(x) &gt; 0"` the span holds `count(x) &gt; 0` while `fix.value` is `count(x) > 0` — they never match, the fix is skipped with a "source no longer matches" warning, and `--fix` leaves the code unchanged. Verified end-to-end: `count(x) &gt; 0` stays put while `count(y) = 0` becomes `empty(y)`.

This silently breaks the fix for every `>`/`>=`/`<`/`<=` comparison — half of the count/string-length forms. `=`/`!=` work because they carry no entity.

The shared `defect()` builder in `src/checks.js` is the single place these offset-family fixes are assembled, so encoding the XML-significant characters (`&`, `<`, `>`) of `fix.value` (and `replacement`) there makes the value match the source for the standard entities and is a no-op for every other check (whose fix spans contain none). Found while adding fix-location assertions to the pack harnesses (#506).

PR #519, opened 2026-07-30
Assert the fix replacement in the code-based linter pack harnesses

### LINKED ISSUE #506: Pack harnesses assert defect positions but never the fix; fixers are covered by one fixture each
Every code-based-linter pack harness asserts only `defects.length` and `defect.line`/`defect.pos` — none asserts `fix.value`/`fix.replacement`. Fix generation is exercised solely by `test/fixer.test.js`, one `<name>.xsl`/`<name>.fixed.xsl` pair per check. So detection is stress-tested (buried, multi-occurrence, negatives) but the span-computation/replacement logic is validated by a single fixture per fixer, and a bug in offsets for an unusual shape can slip through. Fix: assert `fix` in the pack harnesses (or add fixer-focused unit tests). Found during a testing audit.

maxonfjvipon:
Closes #506.

The code-based linter pack harnesses asserted `defects.length` and each defect's `line`/`pos`, but never the `fix` — so fix generation was validated only by the single `<name>.xsl`/`.fixed.xsl` pair per check in `test/fixer.test.js`. A fix producing the wrong replacement for an unusual shape (one not in that single fixture) could slip through.

Each code-based linter pack now carries a `found.fixes` array aligned 1:1 with `positions` — the expected `fix.replacement` for each occurrence, or `null` for a report-only defect:

```yaml
  positions: [ [ 4, 36 ], [ 5, 36 ], ..., [ 11, 32 ] ]
  fixes: [ "exists(x)", "empty(x)", ..., "exists(para)" ]
```

and every harness asserts it alongside the positions. Because the packs already pin many occurrences (buried, reversed, nested, three-in-one), this now checks the fix *text* is right at each one, not just its span.

While writing this, the assertion surfaced a real fixer bug — `count`/`string-length` `--fix` silently no-ops on `<`/`>` comparisons because the decoded fix value never matches the entity-encoded (or raw-`>`) source. That is a deeper fixer/coordinate issue, filed separately as **#518**; this PR scopes to the harness assertion (entity-independent, comparing the emitted replacement text). Full suite, 100% coverage, ESLint, and the CI xcop pass stay green.


maxonfjvipon:
@rultor merge

PR #520, opened 2026-07-30
Require a test for every format and validation check in conformance

### LINKED ISSUE #507: conformance.test.js requires a test pack only for xpath and corpus checks, not format/validation
`test/conformance.test.js` sets `PACKED = {xpath: 'xpath-packs', corpus: 'corpus-packs'}`, so a new `checks/format/<name>.yaml` (or validation) with a motive but no test pack and no linter test passes conformance. The only backstop is the 100% coverage gate forcing *some* test to touch the code — not that the check has a pack. Fix: extend the pack requirement to format/validation (mapping each to its `*-packs` dir / linter test). Found during a testing audit.

maxonfjvipon:
Closes #507.

`conformance.test.js` enforced "every rule has a test pack" only for `PACKED = {xpath, corpus}`, so a new `format` or `validation` check could ship with a motive but no test — the only backstop was the 100% coverage gate forcing *some* test to touch the code, not that the check itself is exercised.

Two new assertions close that:

- **format** — every format check must have a pack whose `pack:` field equals its name, matched *anywhere* under `test/resources` (a format check's packs live in the scattered per-linter dirs, e.g. `count-compared-to-zero` in `count-packs/`, not a single `format-packs/`).
- **validation** — every validation check must be named in some `*.test.js` (`malformed-stylesheet` and `invalid-xpath-expression` are tested by bespoke harnesses with descriptive pack names / inline sources, not a name-matching pack).

All 14 format checks and both validators already satisfy this, so the build stays green; the guard bites only when a future check ships testless. Full suite, 100% coverage, and ESLint pass. CLAUDE.md updated.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -155,7 +155,7 @@ The `--fix` behavior of every fixable check is covered end-to-end in `test/fixer
 
 ## Adding a New Rule
 
-Rule names are kebab-case with no `template-match-` (or other noise) prefix. Every rule needs a motive and at least one test pack. `test/conformance.test.js` enforces all three across `xpath` and `corpus` (naming and motives are enforced for `validation` and `format` too), so a rule that misnames itself, drops its motive, or ships without a pack fails the build.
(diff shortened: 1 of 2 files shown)


maxonfjvipon:
@rultor merge

PR #521, opened 2026-07-30
Machine-enforce the xcop pack-dir and local-rule conventions

### LINKED ISSUE #508: Several CLAUDE.md conventions are not machine-enforced despite the project's own rule
CLAUDE.md:28 states every code-style convention must be machine-enforced, yet these are convention-only:

- "Table-driven tests" — no check that shared-body `it` blocks are expressed as a data array + generator.
- "Test the hard cases, not just the textbook one" — no check for a minimum `found.amount` or a buried/multi-occurrence position.
- "A new pack dir must be registered in `test/xcop.test.js`'s PACKS list" — unenforced (CLAUDE.md admits "or it goes unchecked").
- The custom ESLint rule `local/no-redundant-return-variable` has no unit test of its own.

Some are hard to mechanize, but they are exactly the drift the project's doctrine says should fail CI. Found during a CI/testing audit.

maxonfjvipon:
Closes #508.

Two conventions CLAUDE.md's own "every convention must be machine-enforced" rule left unenforced:

**Pack-dir registration.** `test/xcop.test.js` kept a hand-maintained `PACKS` list of `*-packs` directories, and CLAUDE.md admitted "a new pack directory must be added … or it goes unchecked." It now auto-discovers every `test/resources/*-packs` directory:

```js
const PACKS = fs.readdirSync(RESOURCES)
  .filter((entry) => entry.endsWith('-packs'))
  .flatMap((dir) => allFilesFrom(path.resolve(RESOURCES, dir)))
```

The old list matched the dirs on disk exactly, so the same 183 fixtures are checked — now future-proof.

**The local ESLint rule had no test.** `no-redundant-return-variable` moved out of `eslint.config.mjs` into `eslint-local-rules.js` (imported back into the config unchanged) and is now unit-tested with ESLint's `RuleTester` in `test/eslint-local-rules.test.js` — valid cases (direct return, used-in-between, different name, multi-declarator) and invalid ones. Verified the rule still fires on the codebase after the move.

The two inherently-subjective conventions ("table-driven tests", "test the hard cases") remain author-guidance, as #508 itself noted. Full suite, 100% coverage, ESLint, and the CI xcop pass stay green.


maxonfjvipon:
@rultor merge

Issue #511, opened 2026-07-29
Gruntfile default runs mocha before eslint, masking lint regressions when a test fails


maxonfjvipon:
`Gruntfile.js:30` registers the default task as `[mochacli, eslint]`; grunt stops at the first failing task, so whenever a test fails the ESLint step never runs and a co-occurring lint regression is hidden until the test is fixed. Since lint is fast and independent, running it first (or as a separate job) surfaces both at once. Found during a CI audit.

PR #522, opened 2026-07-30
Run ESLint once in a separate job, not per matrix cell

### LINKED ISSUE #509: ESLint runs redundantly across the 6-cell OS/node matrix
`grunt.yml` runs the whole `grunt` default (`[mochacli, eslint]`) on every matrix cell (`[windows-2022, ubuntu-24.04, macos-15] × node [20,22]`), but ESLint is platform- and node-independent, so five of the six runs are wasted (and it runs again in `daily.yml`/`release.yml`). Suggest a single lint job (ubuntu/one node) separate from the test matrix. Found during a CI audit.

maxonfjvipon:
Closes #509.

The `grunt` workflow's `build` matrix ran the whole default task — `mochacli` then `eslint` — on every cell of `[windows, ubuntu, macos] × [node 20, 22]`, so ESLint (platform- and node-independent) ran six times over for one result.

The matrix now runs `grunt mochacli` (tests only), and a single `lint` job runs `grunt eslint` once on ubuntu/node 22. Same coverage, five fewer redundant lint runs, and — as a side effect — a lint failure and a test failure now surface as separate independent checks rather than one masking the other (part of #511's concern too).


maxonfjvipon:
@rultor merge

PR #523, opened 2026-07-30
Drop the near-no-op shellcheck workflow

### LINKED ISSUE #510: shellcheck.yml is a near-no-op and duplicates actionlint
`shellcheck.yml` runs `ludeeus/action-shellcheck@master` on a repo with no `*.sh` files — all shell is inline in workflow `run:` steps, which `actionlint.yml` already shellchecks. The job adds CI time for no coverage (and pins the action at `@master`, unlike the rest of the repo). Suggest dropping it or pointing it at real scripts if any are added. Found during a CI audit.

maxonfjvipon:
Closes #510.

`shellcheck.yml` ran `ludeeus/action-shellcheck@master` over a repository with **no `*.sh` files** — all shell lives inline in workflow `run:` steps, which the `actionlint` workflow already shellchecks. So it checked nothing, added CI time, and pinned its action at `@master` unlike the rest of the repo.

Removed. `master` has no required status checks (verified), so dropping the `shellcheck` context blocks nothing. If real scripts land later, actionlint still covers workflow shell and a dedicated job can come back.


maxonfjvipon:
@rultor merge

Issue #525, opened 2026-07-30
count/string-length --fix still skips a comparison shifted by an entity earlier in the same attribute value


maxonfjvipon:
Follow-up to #518. That fix made `fixer.js` match entity-flexibly, so `count(x) > 0` now fixes whether the source writes the operator as `>` or `&gt;`. One narrower case remains.

A fix's column is `attribute.columnNumber + 1 + offset`, where `offset` is an offset into the **decoded** attribute value but `columnNumber` is a **raw** source column. When an entity appears *before* the matched comparison in the same attribute value — e.g. `test="@a &lt; 5 and count(x) &gt; 0"` — the raw column of `count` is further right than `columnNumber + 1 + offset` (the `&lt;` expands to four raw characters, one decoded), so the sticky match lands too early, fails, and the fix is **safely skipped** (verified: the source is left unchanged, never corrupted).

Fully resolving it needs the fixer to reconcile the two coordinate systems — e.g. the fix carrying the raw attribute-value start and the decoded offset separately, so the fixer can decode-walk from the value start to the match. Low frequency (a comparison preceded by another `<`/`>`/`&` in the same expression) and safe today, so filed rather than rushed.

PR #526, opened 2026-07-30
Match fixes entity-flexibly so count/string-length --fix works on &gt;/&lt;

### LINKED ISSUE #518: count/string-length --fix silently skips comparisons that use < or > (entity-encoded in XML)

maxonfjvipon:
Closes #518.

`fixer.js` verified a fix by comparing `fix.value` — sliced from the **decoded** attribute text, e.g. `count(x) > 0` — against the **raw** XML source, where a `>`/`<` operator is usually written `&gt;`/`&lt;`. The two never matched, so `src/fixer.js:56`'s exact comparison failed and every `count`/`string-length` `--fix` over a `<`/`>` comparison was silently skipped with a "source no longer matches" warning. (Only `=`/`!=` worked, since they carry no entity — and the fix fixtures happened to use a raw `>`, which is also legal, so the bug never surfaced in tests.)

The verify step now builds a sticky regex where `&`, `<`, `>` each match **both** their literal and entity forms, and replaces the raw span it actually matched:

```js
value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
     .replace(/[&<>]/g, (char) => ENTITIES[char])   // > -> (?:&gt;|>)
```

So a decoded `count(x) > 0` now fixes both `count(x) > 0` and `count(x) &gt; 0`, and it is backward-compatible — a value with literal markup (the `redundant-import` line, `<xsl:import…/>`) still matches via the literal alternative. A genuine non-match is still skipped, never corrupting.

A new `count-entity-comparison` fixture (`count(item) &gt; 0`, `string-length(@name) &lt;= 0`) locks in the entity case. One narrower edge remains — a comparison shifted by an entity *earlier* in the same attribute value (`@a &lt; 5 and count(x) &gt; 0`) mixes a raw column with a decoded offset and is left a **safe skip**; filed as #525. Full suite, 100% coverage, ESLint, and the CI xcop pass stay green.


maxonfjvipon:
@rultor merge

PR #527, opened 2026-07-30
Decode-walk the source so an entity-shifted count/string-length fix applies

### LINKED ISSUE #525: count/string-length --fix still skips a comparison shifted by an entity earlier in the same attribute value

maxonfjvipon:
Closes #525.

#518 made `fixer.js` match entity-flexibly, but it still anchored the match at `fix.col`, which is a **raw** column plus a **decoded** offset. When an entity appears *earlier* in the same attribute value — `test="@a &lt; 5 and count(x) &gt; 0"` — the `&lt;` expands to four raw characters (one decoded), so the raw position of `count` is further right than `fix.col`, the anchored match landed too early, and the fix was silently skipped.

The fix now carries its decoded `offset`, and the fixer **decode-walks** the raw source:

- from the node's raw start (`col - offset`) it skips `offset` decoded characters, reading each named entity (`&lt;`, `&gt;`, `&amp;`, …) as the single character it stands for — so a shifting entity ahead of the match is counted correctly;
- then it matches `value`, decoding as it goes, so `count(x) > 0` fixes whether the source wrote `>` or `&gt;`.

This subsumes #518 and resolves #525: `@a &lt; 5 and count(x) &gt; 0` → `@a &lt; 5 and exists(x)`, with the text before the comparison preserved. A span that no longer decodes to `value` (an already-edited file) is still skipped, never corrupting. A numeric or unknown entity yields a non-match (safe skip) — the named forms xcop produces are what occur in practice.

New `count-shifted-comparison` fixture locks in the entity-before case; unit tests cover the run-off-the-end guard. Full suite, 100% coverage, ESLint, and the CI xcop pass stay green.


maxonfjvipon:
@rultor merge

PR #528, opened 2026-07-30
Offer the count exists()/empty() fix only on XSLT 2.0/3.0

### LINKED ISSUE #485: count-compared-to-zero --fix emits XPath 2.0 exists()/empty() on 1.0 stylesheets
`count-linter.js` rewrites `count(x) > 0` → `exists(x)` and `count(x) = 0` → `empty(x)`, but `fn:exists`/`fn:empty` are XPath 2.0 functions — they do not exist in XPath 1.0. The fix carries no `suggestion: true` flag and no version guard, so plain `xslint --fix` applies it even on an `version="1.0"` stylesheet, silently rewriting working code into a sheet that calls undefined functions.

This is inconsistent with the linter's own siblings. `node-set-linter.js`, `translate-linter.js`, and `name-linter.js` all guard their 2.0-only rewrites on `const MODERN = ['2.0', '3.0']` (e.g. `node-set-linter.js:36,83`). `count-linter.js` has no such guard (`src/count-linter.js:143-169`), and its neighbour `string-length-linter.js` deliberately chose the 1.0-safe `X != ''` / `X = ''` rewrite instead.

Reproduced: on a `version="1.0"` stylesheet, `count(a) > 0` yields a defect whose fix replacement is `exists(a)` with `suggestion` undefined — so `--fix` rewrites it in place.

The detection is fine on every version — using `count()` for existence is a smell in 1.0 too — so the narrow fix is to gate only the *fix*: attach the `exists()`/`empty()` replacement only when `@version` is `2.0`/`3.0`, and on 1.0 either report without a fix or offer a 1.0-safe rewrite (`count(x) > 0` in a boolean context is `boolean(x)`, `count(x) = 0` is `not(x)`). The message should likewise not prescribe `exists`/`empty` to a 1.0 sheet.

maxonfjvipon:
Closes #485.

`count-compared-to-zero` rewrote `count(x) > 0` → `exists(x)` (and `= 0` → `empty(x)`) with no version guard, but `fn:exists`/`fn:empty` are XPath 2.0 functions. On a `version="1.0"` stylesheet, plain `xslint --fix` rewrote working code into calls to undefined functions.

The `exists()`/`empty()` fix is now attached only when the stylesheet is `2.0`/`3.0` — matching the `MODERN` gate its siblings (`node-set`, `translate`, `name`) already use. **Detection is unchanged on every version**: `count(x) > 0` is a wasteful full walk in 1.0 too, so the smell is still reported there — just without a fix, since the direct 1.0 form (the node-set in a boolean context, or `not()`) depends on where the test sits.

A new `count-in-xslt-1-0` pack pins the report-only behaviour (`found.fixes: [null, null]`); the 2.0 packs still get their `exists()`/`empty()` fixes. The message and motive no longer prescribe `exists()`/`empty()` unconditionally. Full suite, 100% coverage, ESLint, and xcop stay green.


maxonfjvipon:
@rultor merge

PR #529, opened 2026-07-30
Make the string-length emptiness fix a suggestion, not a safe fix

### LINKED ISSUE #488: string-length-compared-to-zero --fix is unsafe for an absent attribute or empty node-set
`string-length-linter.js` rewrites `string-length(X) = 0` → `X = ''` as a plain (non-suggestion) `--fix`. That is wrong when `X` is an absent attribute or empty node-set:

- `string-length(@x) = 0` with `@x` absent → `string-length('')` is `0` → **true**.
- `@x = ''` with `@x` absent → an empty node-set compared to a string → **false**.

So `xslint --fix` silently flips the meaning of the test. The `> 0` → `X != ''` direction is safe (both false for an absent node); only the emptiness (`= 0`) direction is affected. Fix: drop the fix for the `= 0` case or mark it `suggestion: true`. Found during a full best-practices audit.

maxonfjvipon:
Closes #488.

`string-length-compared-to-zero` rewrote `string-length(X) op 0` → `X op ''` as a plain (safe) `--fix`, but `X op ''` is **not a general equivalent** of `string-length(X) op 0`:

- **Absent node** — `string-length(@x) = 0` is *true* when `@x` is missing (the empty string has length 0), but `@x = ''` is *false* (an empty node-set matches nothing). So `= 0` → `= ''` flips the result.
- **Multi-node set** — `string-length(X)` reads the first node's value; `X != ''` is true if *any* node is non-empty.

For the common single, present node they agree, but a safe `--fix` must preserve semantics for *all* inputs, so this one shouldn't have been auto-applied. It is now a **suggestion**: `--fix-suggestions` applies it, plain `--fix` leaves it — matching the tier used for `name-compared-to-string`, `translate-for-case`, etc.

Detection is unchanged (the audit confirmed no false positives). The `count-entity-comparison` fixture drops its incidental string-length line (the fixer's entity handling is check-agnostic and covered by the count fixtures), so `--fix` there still exercises the count entity path. Motive, README (moved to the suggestions list), and CLAUDE.md explain the caveat. Full suite, 100% coverage, ESLint, and xcop stay green.


maxonfjvipon:
@rultor merge

Issue #530, opened 2026-07-30
Make updating a check's motive mandatory when its behavior changes


maxonfjvipon:
The motive (`src/resources/motives/{kind}/<name>.md`) is where an end user learns *why* a check exists, what it flags, and what its fix does — the check's most user-facing documentation. Yet CLAUDE.md's "Keeping Docs in Sync" rule listed README, CLAUDE.md, and the docs site but did not explicitly require re-reading the motive when a check's behavior changes.

A check-behavior change (what it flags, severity, fix or fix tier, version scope, the constructs it leaves alone) that leaves the motive describing the old behavior misleads exactly the person the check is for. Add an explicit, mandatory rule: whenever a check's behavior changes, re-read and, if needed, update its motive in the same change.

PR #534, opened 2026-07-30
Fire function-use-in-xslt-1 on any non-2.0/3.0 stylesheet

### LINKED ISSUE #532: function-use-in-xslt-1 misses xsl:function in an unversioned (or version=1.1) stylesheet
`function-use-in-xslt-1` keys on `/xsl:stylesheet[@version="1.0"]//xsl:function`. `xsl:function` is a 2.0+ instruction, so declaring it under 1.0 is the error the check names — but the exact `@version="1.0"` predicate leaves two real violations unflagged:

- **Unversioned stylesheet** — `<xsl:stylesheet xmlns:xsl="…"><xsl:function name="my:f"/></xsl:stylesheet>` has no `@version`, so `xsl:function` (2.0+) slips through. Verified: not flagged. (`missing-version-in-stylesheet` flags the absent `@version` separately, but the misplaced `xsl:function` itself is not caught.)
- **`version="1.1"`** — a draft version with no `xsl:function`; a declared-1.1 sheet using it is not flagged.

(The `xsl:transform` root case — `<xsl:transform version="1.0">…<xsl:function>` — is the same blind spot tracked systemically in #498, since the rule hard-codes the `/xsl:stylesheet` root.)

Suggested framing: fire on `xsl:function` when the stylesheet is not a 2.0/3.0 sheet (`@version` absent or not in `2.0`/`3.0`), against either root, rather than matching the exact string `"1.0"`. Found during an XSLT-version-sensitivity audit of the checks.

maxonfjvipon:
Closes #532.

`function-use-in-xslt-1` keyed on `/xsl:stylesheet[@version="1.0"]//xsl:function`. `xsl:function` is a 2.0+ instruction, so declaring one where the stylesheet isn't 2.0/3.0 is the error — but the exact `@version="1.0"` + `/xsl:stylesheet` predicate missed real violations: an **unversioned** stylesheet, a `version="1.1"` one, and any `xsl:transform`-rooted sheet.

The selector is now:

```
//xsl:function[not(/*/@version = '2.0' or /*/@version = '3.0')]
```

`/*` matches either root (`xsl:stylesheet` or `xsl:transform`), and the predicate fires whenever the version isn't `2.0`/`3.0` — covering `1.0`, `1.1`, and unversioned — while leaving a genuine 2.0/3.0 function alone. Message and motive are generalised from "version 1.0" to "not 2.0/3.0" (per the #530 motive rule).

New packs cover the `xsl:transform` root, the unversioned stylesheet (whose unavoidable `missing-version` co-fire is expressed as a 3-element position), and a `2.0` negative that must stay silent. Full suite, 100% coverage, ESLint, and xcop stay green.

### DIFF
--- a/src/resources/checks/xpath/function-use-in-xslt-1.yaml
+++ b/src/resources/checks/xpath/function-use-in-xslt-1.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: /xsl:stylesheet[@version="1.0"]//xsl:function
+xpath: "//xsl:function[not(/*/@version = '2.0' or /*/@version = '3.0')]"
 severity: error
-message: "'xsl:function' is declared in XSLT version 1.0. Use a named template or upgrade the stylesheet version to 2.0 instead."
+message: "'xsl:function' requires XSLT 2.0 or later, but this stylesheet is not version 2.0 or 3.0. Use a named template, or set the version to 2.0."
--- a/src/resources/motives/xpath/function-use-in-xslt-1.md
+++ b/src/resources/motives/xpath/function-use-in-xslt-1.md
@@ -1,35 +1,41 @@
-# Function use in XSLT version 1.0
+# Function use in a pre-2.0 stylesheet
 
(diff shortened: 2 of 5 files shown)


maxonfjvipon:
@rultor merge

PR #536, opened 2026-07-30
Require checking version-dependence when a check is added or modified

### LINKED ISSUE #535: Make checking version-dependence mandatory when a check is added or modified
Several checks are XSLT-version-sensitive: their detection or their fix is only valid for certain versions (`exists()`/`empty()`, `lower-case()`, `xsl:function`, the `node()` sequence model, the deprecated `namespace::` axis). When such a check lacks a version guard it either false-positives on the wrong dialect or its `--fix` emits code the stylesheet's version cannot run (see #485).

CLAUDE.md should make it a mandatory authoring step: when a check is added or modified, decide whether it is version-dependent, and if so make the version test part of the check — gate the detection or the fix on the root's `@version` (handling `xsl:transform` and unversioned sheets), as `count`/`node-set`/`translate`/`name` already do with `MODERN = ['2.0','3.0']`.

maxonfjvipon:
Closes #535.

Adds a mandatory authoring rule to CLAUDE.md's "Adding a New Rule" section: when a check is added or modified, decide whether it is XSLT-version-dependent (its detection or fix valid only for certain versions — 2.0+ constructs, or the deprecated `namespace::` axis), and if so make the version test part of the check. Gate on the root's version via `documentElement`/`MODERN = ['2.0','3.0']` (handling `xsl:transform` and unversioned sheets); a declarative rule keys on `/*/@version`, not `/xsl:stylesheet[@version="1.0"]`. Never emit a fix the stylesheet's version can't run — withhold it or emit the version-appropriate form. A version-sensitive check with no guard is a bug.

Doc-only. (This is the rule that #485/#532 retrofitted onto `count` and `function-use-in-xslt-1`.)


maxonfjvipon:
@rultor merge

PR #538, opened 2026-07-30
Give count-compared-to-zero a version-appropriate fix on every version

### LINKED ISSUE #537: count-compared-to-zero: guide 1.0 users and give 1.0 its own safe fix (boolean/not/x)
After #485, `count-compared-to-zero` fixes to `exists()`/`empty()` on 2.0/3.0 and is report-only on 1.0. Two gaps remain, raised on re-examination:

1. **The motive doesn't tell a 1.0 user what to do** — it only shows the 2.0+ `exists()`/`empty()` form and hand-waves that "the direct 1.0 form depends on where the test sits."
2. **Is it even a smell in 1.0?** Yes: `count(x) > 0` forces a full count of `x` when a node-set in a boolean context short-circuits at the first node. And 1.0 *does* have a clean direct form — `boolean(x)` (or a bare `x` in an `xsl:if`/`xsl:when` `@test`, which already coerces) for the existence forms, and `not(x)` for the emptiness forms. All are valid in 1.0 (and 2.0/3.0).

So the check should (a) explain the smell and the 1.0 alternative in the motive, and (b) offer a **safe** 1.0 fix — context-aware like `redundant-double-negation`: existence → bare `x` in a whole `@test`, `boolean(x)` elsewhere; emptiness → `not(x)`. The fix becomes available on every version (2.0/3.0 → `exists`/`empty`; 1.0, 1.1, unversioned → `boolean`/`not`/`x`, which are valid everywhere), each version-appropriate.

maxonfjvipon:
Closes #537.

Re-examining the count check after #485: it reported the count-for-existence smell on 1.0 but offered **no fix**, and its motive only showed the 2.0+ `exists()`/`empty()` form — leaving a 1.0 user without guidance. But it *is* a genuine smell in 1.0 (`count(x) > 0` walks the whole sequence where a node-set in a boolean context short-circuits at the first node), and 1.0 has a clean direct form.

So the fix is now **version-appropriate and safe on every version**:

| | existence (`> 0`, `!= 0`, `>= 1`, reversed) | emptiness (`= 0`, `<= 0`, `< 1`) |
|---|---|---|
| 2.0 / 3.0 | `exists(x)` | `empty(x)` |
| 1.0 / unversioned | bare `x` in a whole `@test`, else `boolean(x)` | `not(x)` |

The 1.0 forms are context-aware like `redundant-double-negation` (bare `x` in a whole `@test`, `boolean(x)` in a `@select`), so the fix never leaves a `boolean(x)` that `redundant-boolean-call` would re-flag. They're valid in every version, so an unversioned stylesheet gets them too. `comparisons.js` now merges the classifier's fields through, so the linter builds the version-appropriate replacement.

The **motive** now explains the smell and gives the 1.0 alternative explicitly (the point you raised). A `count-in-xslt-1-0` fix fixture proves `--fix` produces `item` / `boolean(node())` / `not(@x)` end-to-end, and the 1.0 pack asserts the fixes. Full suite, 100% coverage, ESLint, and xcop stay green.


maxonfjvipon:
@rultor merge

PR #539, opened 2026-07-30
Convert using-namespace-axis to a code-based linter

### LINKED ISSUE #497: using-namespace-axis: string-literal false positive, XSLT 3.0 miss, and narrow attribute scope
`using-namespace-axis` selector: `//*[(@match | @select)[contains(., 'namespace::')][/xsl:stylesheet/@version = '2.0']]`. Three problems:

1. `contains(., 'namespace::')` is a raw string test, so a literal string `'namespace::'` in a `@select` false-positives (verified).
2. Gated to `@version = '2.0'` only, so the axis in a 3.0 stylesheet (still deprecated) is missed (verified false negative).
3. Only `@match`/`@select` are scanned, so the axis in `@use`, `@test`, etc. is missed.

Should tokenize (to skip string literals) and treat 2.0/3.0 alike. Found during a best-practices audit.

maxonfjvipon:
The `using-namespace-axis` check was a declarative XPath rule with three problems (#497): it tested `contains(., 'namespace::')`, so a literal string `'namespace::'` in a `@select` false-positived; it was gated to `@version = '2.0'`, so the axis in a 3.0 stylesheet (still deprecated) was missed; and it scanned `@match`/`@select` only, so the axis in `@use`, `@test`, and the rest slipped through.

This replaces it with a code-based linter that tokenizes every XPath and pattern attribute and reports each `namespace::` axis on an XSLT 2.0 or 3.0 stylesheet. Because the lexer keeps string literals whole and isolates the axis from lookalike calls, all three problems go away at once:

```js
const axes = function(expression) {
  return tokenized(expression)
    .filter((token) => token.type === TOKENS.NAMESPACE)
    .map((token) => token.start)
}
```

In XSLT 1.0, where `namespace::` is the standard way to read namespace nodes, the check stays silent. Rewriting the axis to `in-scope-prefixes()`/`namespace-uri-for-prefix()` is a structural change, so the check remains report-only.

The check moves from the `xpath` kind to `format` since its logic now lives in code; the YAML, motive, and test packs move with it, and a new harness plus 2.0 (hard cases and negative neighbours), 3.0, and 1.0 packs cover it. Closes #497.


maxonfjvipon:
@rultor merge

Issue #499, opened 2026-07-29
Reconsider the opinionated checks: severities, thresholds, and idiom bans


maxonfjvipon:
A full best-practices audit (following #423) found ~20 checks that are defensible but opinionated — worth a deliberate decision on severity/threshold/scope rather than leaving them as silent policy:

**Severity**
- `not-using-output` fires as **error** on entirely valid, ubiquitous code (the default output method is spec-defined); its own motive concedes it is a consistency choice. Warning at most.
- `missing-version-in-stylesheet` is a warning, but a conformant processor rejects a version-less stylesheet — arguably error.

**Thresholds (all arbitrary)**
- `too-many-templates` at 10 is low; idiomatic push-style modules routinely have 20+ small templates.
- `function-complexity` (>50 `xsl:*`), `oversized-template` (>100).

**Idiom bans / nuance**
- `name-compared-to-string` flags the standard namespace-agnostic `local-name()='x'` idiom (the only clean option in 1.0).
- `redundant-whitespace` penalizes deliberately multi-line-formatted complex 2.0/3.0 XPath (`if/then/else`).
- `short-names` flags idiomatic `$i`/`$n`/`$x` accumulators and math params.
- `text-outside-xsl-text` flags valid bare text in `xsl:if`/`when`/`otherwise` — including the repo's own "Correct" examples in other motives.
- `use-choose-without-otherwise` — emitting nothing for unmatched cases is a valid, common pattern.
- `select-starts-with-double-slash` — a leading `//` in a top-level variable is correct and idiomatic (context is the root).
- `starts-with-double-slash` motive says the fix is "exactly the same," but `match="//x"` has default priority 0.5 vs `match="x"` at 0, so dropping `//` can change template-conflict resolution — the motive is misleading and the safe-fix claim is suspect.
- `empty-variable` ignores `@as`, so a deliberate empty sequence `<xsl:variable name="e" as="node()*"/>` in 2.0/3.0 is flagged.
- `missing-id-in-stylesheet` — `@id` is inert on a standalone stylesheet; consistency-only, near-nobody sets it. Reconsider whether it earns a warning.

Each needs a keep / soften / re-severity / drop decision. Found during a best-practices audit.

maxonfjvipon:
Spot-checked every bullet that states a fact rather than a preference. The
thresholds are as quoted — `count(//xsl:template) >= 10`, `count(.//xsl:*) > 50`,
[...]

PR #540, opened 2026-07-30
Stop empty-variable flagging a typed empty sequence

maxonfjvipon:
`empty-variable` flags `//xsl:variable[not(@select) and not(node())]` — a variable with neither a `@select` nor content, which binds the empty string. But a variable that declares a type is a different case: `<xsl:variable name="acc" as="node()*"/>` in XSLT 2.0/3.0 binds the empty *sequence* of that type on purpose (a deliberate empty accumulator), and the check was reporting it as an accidental empty string.

Exclude `not(@as)`:

```yaml
xpath: //xsl:variable[not(@select) and not(node()) and not(@as)]
```

The `@as` attribute exists only in 2.0/3.0, so its presence is itself the signal that the emptiness is intentional — no separate version gate is needed. The motive gains the typed-empty-sequence case, and the pack a negative neighbour that must stay silent.

This addresses the `empty-variable`/`@as` item of #499; the rest of that tracker's opinionated-check decisions remain open.

### DIFF
--- a/src/resources/checks/xpath/empty-variable.yaml
+++ b/src/resources/checks/xpath/empty-variable.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: //xsl:variable[not(@select) and not(node())]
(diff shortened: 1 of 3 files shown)


maxonfjvipon:
@rultor merge

Issue #498, opened 2026-07-29
Read a cross-file reference off the tokens rather than the raw attribute text


maxonfjvipon:
A full audit found three cross-cutting patterns that make many checks silently miss valid targets. Two of the three have since landed under their own tickets, and each is machine-enforced now, so what is left here is the third.

1. **Lexical QName compares — landed.** Rules written as `name()='xsl:variable'` compare the raw prefix, so any stylesheet that binds the XSLT namespace to a non-`xsl` prefix is missed, and one binding `xsl:` to something else draws the check against a literal result element. Both directions were live in the tree: TEI's `rdf/make-acdc.xsl` writes its XSLT as `XSL:`, so eight checks read none of it. Every selector spells the node test now, whose prefix XPath binds for us, and `test/conformance.test.js` refuses a `name()` compare in a selector of any kind (#784).
2. **`/xsl:stylesheet`-only root — landed.** Root-level rules matched `/xsl:stylesheet` and skipped the equally-valid `xsl:transform` root entirely. Every one of them matches `(/xsl:stylesheet | /xsl:transform)` now, and `test/conformance.test.js` enforces the pair, a descendant root test included.
3. **Substring reference matching in corpus checks.** `unused-function` and `unreachable-function` looked for `name(` and `unused-variable` for `$name` as substrings of the raw attribute text, so what stood between the mark and the name was invisible. Four blind spots, two inventing a defect against working code and two withholding one: a gap in front of the bracket a call opens (`my:spaced (1)`) and a 3.0 named function reference, which carries no bracket at all (`my:pick#1`), are references the scan missed, so the function was reported dead; a mark inside a string literal or a comment (`concat('$quoted', 'x')`, `1 (: $commented :)`) is a name no processor evaluates, so a genuinely unused declaration went unreported. Needs token-aware reference detection.

`unused-named-template` was named in the third group and does not belong to it: it carries no `reference` key at all, matching `//xsl:call-template/@name` by exact identity rather than by substring, so no reading of the reference text was ever its fault. Its real gap is the shadow `_name`, whose value is an attribute value template, which wants the static evaluator #851 holds — and that gap can only ever leave a report unmade, never invent one. Its `nursery:` mark names #851 for that reason.

Grouped because they share a root cause (crude string/prefix matching where a namespace-, root- or token-aware selector is needed). Found during a best-practices audit.

maxonfjvipon:
[...]

PR #542, opened 2026-07-30
Make root-anchored check selectors xsl:transform-robust

maxonfjvipon:
Addresses pattern 2 of #498. Nine declarative rules anchored on `/xsl:stylesheet` only, so a stylesheet rooted at `xsl:transform` — an exact synonym of `xsl:stylesheet` — silently escaped every one of them: `missing-version`, `missing-id`, `not-using-output`, `stylesheet-has-no-templates`, `too-many-templates`, `output-method-xml`, `null-output-from-stylesheet`, `function-template-is-not-child-of-stylesheet`, and `using-not-outermost-stylesheet`.

Each now matches both roots, e.g.:

```yaml
xpath: (/xsl:stylesheet | /xsl:transform)[not(@version)]
```

The nested-stylesheet check also catches a nested `xsl:transform`, not just a nested `xsl:stylesheet`.

And it is machine-enforced so the miss cannot recur — a new conformance test fails any check whose `xpath`/`declaration`/`usage` anchors on a root `/xsl:stylesheet` without the paired `/xsl:transform`:

```js
it('pairs every root xsl:stylesheet with an xsl:transform', function() {
  const rooted = /(?<!\/)\/xsl:stylesheet/
  const paired = /(?<!\/)\/xsl:transform/
  ...
})
```

Six transform-root packs prove the fixed checks fire (including `using-not-outermost` on a nested `xsl:transform`), the four motives that named only `xsl:stylesheet` now name both roots, and CLAUDE.md documents the convention. The code-based linters already read the root via `documentElement`, which is root-name-agnostic, so they needed no change.

Patterns 1 (lexical QName compares) and 3 (substring reference matching in corpus checks) of #498 remain open.


maxonfjvipon:
@rultor merge

PR #543, opened 2026-07-30
Flag XSLT 2.0 constructs used in a version=1.0 stylesheet

### LINKED ISSUE #533: Extend 1.0-dialect checking: flag 2.0+ constructs beyond xsl:function in a version=1.0 sheet
`function-use-in-xslt-1` is the *only* check that flags a 2.0+ construct used in a declared XSLT 1.0 stylesheet, yet its name implies broader coverage. Other 2.0/3.0-only constructs in a `version="1.0"` sheet are silently accepted:

- functions: `lower-case()`/`upper-case()`, `exists()`/`empty()`, `matches()`, `tokenize()`, `current-group()`, …
- instructions: `xsl:for-each-group`, `xsl:sequence`, `xsl:analyze-string`, `xsl:next-match`, …
- syntax: the `as` sequence-type attribute, `*:name` wildcards, the `||` concatenation operator, `if/then/else`.

The XPath validator can't catch these because it compiles under an XPath 3.1 engine (by design — so 1.0 numeric coercion isn't a false positive), so `invalid-xpath-expression` accepts them. A 1.0 processor (or Saxon-HE in 1.0 mode) would reject them, so a declared-1.0 sheet using them is a real portability bug.

Proposal: a check (e.g. `modern-construct-in-xslt-1`, generalizing `function-use-in-xslt-1`) that flags a curated set of 2.0+ instructions/functions/syntax when the stylesheet declares `version="1.0"`. Scope carefully to avoid false positives (custom functions in a user namespace are fine; the list must be 2.0+-reserved names). Enhancement, surfaced by an XSLT-version-sensitivity audit.

maxonfjvipon:
Closes #533. `function-use-in-xslt-1` was the *only* check flagging a 2.0+ construct in a declared-1.0 stylesheet, so other 2.0-only constructs were silently accepted — a real portability bug, since a conformant 1.0 processor rejects them and the XPath validator can't catch them (it compiles under a 3.1 engine by design, so 1.0 numeric coercion isn't a false positive).

The new `modern-construct-in-xslt-1` check flags, on a `version="1.0"` root, a curated set of 2.0 instructions plus the `as` sequence-type attribute:

```yaml
xpath: "(/xsl:stylesheet | /xsl:transform)[@version = '1.0']//(xsl:for-each-group | xsl:sequence | xsl:analyze-string | xsl:next-match | xsl:perform-sort | xsl:namespace | xsl:character-map | xsl:result-document | xsl:import-schema | xsl:*[@as])"
```

Deliberate scope decisions:

- **Only `xsl:`-namespace elements are examined for `@as`**, so a literal result element that legitimately carries one — `<link rel="preload" as="script"/>` in HTML output — is never a false positive. A pack proves this.
- **`xsl:function` is left to `function-use-in-xslt-1`** (which also catches unversioned/`1.1`), so the two checks don't double-flag.
- **The 2.0 functions/operators inside XPath** (`lower-case()`, `||`, `if/then/else`, `*:name`) are not flagged — telling a reserved built-in from a user-namespaced call needs token-aware parsing, which waits on the full-fidelity parser (#228). The motive says so.
- **Report-only**, since rewriting a 2.0 construct to its 1.0 equivalent is structural.

Three packs cover it (a 1.0 sheet firing on four constructs, a 2.0 sheet that stays silent, the literal-`as` guard). Four existing 1.0-with-2.0-construct fixtures record the new co-fire, except `stylesheet-has-templates` whose incidental `@as` scaffolding moves to `version="2.0"`.

### DIFF
--- a/src/resources/checks/xpath/modern-construct-in-xslt-1.yaml
+++ b/src/resources/checks/xpath/modern-construct-in-xslt-1.yaml
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
+xpath: "(/xsl:stylesheet | /xsl:transform)[@version = '1.0']//(xsl:for-each-group | xsl:sequence | xsl:analyze-string | xsl:next-match | xsl:perform-sort | xsl:namespace | xsl:character-map | xsl:result-document | xsl:import-schema | xsl:*[@as])"
+severity: error
+message: "This XSLT 2.0 construct is used in a stylesheet declared version 1.0, which a conformant 1.0 processor rejects. Use a 1.0 equivalent, or set the version to 2.0."
--- a/src/resources/motives/xpath/modern-construct-in-xslt-1.md
+++ b/src/resources/motives/xpath/modern-construct-in-xslt-1.md
@@ -1,0 +1,53 @@
+# Modern construct in a 1.0 stylesheet
+
+A stylesheet that declares `version="1.0"` must run on an XSLT 1.0 processor,
+where the instructions and syntax introduced in XSLT 2.0 simply do not exist. A
+1.0 processor (or Saxon-HE in 1.0 mode) rejects them, so a declared-1.0
+stylesheet that reaches for a 2.0 construct is a real portability bug — it will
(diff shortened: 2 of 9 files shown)


maxonfjvipon:
@rultor merge

PR #545, opened 2026-07-30
Hoist the version guard out of function-use-in-xslt-1's predicate

### LINKED ISSUE #544: Hoist the version guard in function-use-in-xslt-1 out of the per-node predicate
An audit of every declarative check selector (following a review of the root-anchored rules in #498) turned up one clean optimization of the "switch the root and non-root parts" kind.

`function-use-in-xslt-1` selects `//xsl:function[not(/*/@version = '2.0' or /*/@version = '3.0')]`. The `/*/@version` test is a whole-document fact, yet it sits in the per-node predicate, so it is re-evaluated for every `xsl:function` the `//` step walks. Because the version gates the *entire* rule (a single regime — the check fires only when the root is not 2.0/3.0), it hoists cleanly to the front with no union and no duplicated predicate:

```
/*[not(@version = '2.0' or @version = '3.0')]//xsl:function
```

Now the version is read once at the root, and the `//xsl:function` subtree is not walked at all when the stylesheet is 2.0/3.0. The two forms are equivalent across every case (1.0, 2.0, 3.0, unversioned, `xsl:transform` root, multiple functions), and the reported node is unchanged, so the packs are untouched.

The `/*` root step is inherently root-name-agnostic, so it also stays `xsl:transform`-robust without the explicit union the `(/xsl:stylesheet | /xsl:transform)` root checks need.

Note that this hoist is only clean because the guard gates the whole rule. `empty-variable`, where the version gates only the `@as` sub-clause, deliberately keeps its guard nested — hoisting it would need a two-branch union with a duplicated emptiness predicate.

maxonfjvipon:
Closes #544. From the audit of every declarative selector, one clean "switch the root and non-root parts" optimization.

`function-use-in-xslt-1` read the whole-document `/*/@version` inside every `xsl:function`'s predicate:

```yaml
xpath: "//xsl:function[not(/*/@version = '2.0' or /*/@version = '3.0')]"
```

Because the version gates the entire rule (a single regime — it fires only when the root is not 2.0/3.0), it hoists to the root step, read once, and the `//xsl:function` subtree is skipped entirely on a 2.0/3.0 stylesheet:

```yaml
xpath: "/*[not(@version = '2.0' or @version = '3.0')]//xsl:function"
```

Verified equivalent across every case (1.0, 2.0, 3.0, unversioned, `xsl:transform` root, multiple functions) with the same reported node, so the packs are untouched. The `/*` root step is inherently root-name-agnostic, so it stays `xsl:transform`-robust without the explicit union the root checks need.

This hoist is clean only because the guard gates the whole rule. `empty-variable`, where the version gates just the `@as` sub-clause, keeps its guard nested on purpose — hoisting it would need a two-branch union with a duplicated predicate.

### DIFF
--- a/src/resources/checks/xpath/function-use-in-xslt-1.yaml
+++ b/src/resources/checks/xpath/function-use-in-xslt-1.yaml
@@ -1,6 +1,6 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
 # SPDX-License-Identifier: MIT
 ---
-xpath: "//xsl:function[not(/*/@version = '2.0' or /*/@version = '3.0')]"
+xpath: "/*[not(@version = '2.0' or @version = '3.0')]//xsl:function"
 severity: error
 message: "'xsl:function' requires XSLT 2.0 or later, but this stylesheet is not version 2.0 or 3.0. Use a named template, or set the version to 2.0."


maxonfjvipon:
@rultor merge

PR #585, opened 2026-07-30
Audit starts-with-double-slash to maturity

### LINKED ISSUE #583: `starts-with-double-slash` fix shifts template priority in `src/fixers.js`
The `startsWithDoubleSlash` builder in `src/fixers.js:100` returns a fix with no
`suggestion: true`, so a plain `--fix` rewrites `match="//item"` into
`match="item"`:

```js
const startsWithDoubleSlash = function(node) {
  const match = node.getAttributeNode('match')
  return {
    line: match.lineNumber,
    col: match.columnNumber - match.name.length - 1,
    value: `${match.name}="${match.value}"`,
    replacement: `${match.name}="${match.value.slice(2)}"`,
  }
}
```

The two patterns select the same nodes, but they do not rank the same. A pattern
carrying a `/` step has a default priority of 0.5, while a lone name test has 0,
so dropping the slashes costs the template half a point against every rule it
competes with. On this stylesheet xsltproc prints `DOUBLESLASH` before the fix
and `SPECIFIC` after it:

```xsl
<xsl:template match="list/item">SPECIFIC</xsl:template>
<xsl:template match="//item">DOUBLESLASH</xsl:template>
```

That is a behavior change, so the edit belongs behind `--fix-suggestions` rather
than `--fix`. `CLAUDE.md` names this exact trap in its motive-quality bar, and
the motive in `src/resources/motives/xpath/starts-with-double-slash.md` walks
into it anyway: it promises the `//` "adds nothing but noise", and never says
which fix tier the check ships.

The selector in `src/resources/checks/xpath/starts-with-double-slash.yaml:4`
reads the attribute raw, where its sibling `select-starts-with-double-slash`
normalizes first:

```yaml
xpath: //xsl:template[starts-with(@match, '//')]
```

So `<xsl:template match=" //spaced">` slips past, and `use-double-slash` picks it
up instead — with the wrong advice, since that message tells the author to "use a
specific path" when the leading `//` is merely redundant. The `slice(2)` above
would not survive the obvious repair either: on `" //spaced"` it yields
`/spaced`, turning an unanchored pattern into an absolute one.

The same selector misses every pattern that is not a template's. A leading `//`
is redundant in any XSLT pattern, because all of them are matched unanchored, yet
`<xsl:key match="//keyed"/>`, `<xsl:number count="//counted" from="//started"/>`,
`<xsl:for-each-group group-starting-with="//grouped">` and
`xsl:accumulator-rule/@match` all go unreported — a stylesheet holding the lot of
them draws nothing from this check.

None of that is covered, because
`test/resources/xpath-packs/starts-with-double-slash.yaml` pins a single clean
occurrence at one position: no lookalike negatives, no non-`xsl` prefix, no
second and third occurrence to catch a first-match bug. `README.md:44` is stale
on top of it, quoting a sample defect as "starts with //, which scans the entire
document tree" long after the message became "which is redundant since an
unanchored pattern already matches at any depth".

Selecting the pattern attributes themselves rather than their elements —
`((//xsl:template | //xsl:key | //xsl:accumulator-rule)/@match |
//xsl:number/(@count | @from) | //xsl:for-each-group/(@group-starting-with |
@group-ending-with))[starts-with(normalize-space(), '//')]` — covers the whole
set at once and hands the fixer the one attribute it must edit. With that in
place, mark the fix `suggestion: true`, strip from the real `//` instead of by
`slice(2)`, normalize the negation in `use-double-slash` so the two stay
disjoint, rewrite the message and the motive around priority, grow the pack to
the hard cases, and refresh `README.md` and the docs site; the check earns
`mature: true` once all of that holds.

maxonfjvipon:
The `--fix` for this check was not safe. Dropping the leading `//` off a
`match` keeps the node set but not the priority: a pattern with a `/` step
defaults to 0.5, a lone name test to 0. On this sheet xsltproc printed
`DOUBLESLASH` before a plain `--fix` and `SPECIFIC` after it.

```xsl
<xsl:template match="list/item">SPECIFIC</xsl:template>
<xsl:template match="//item">DOUBLESLASH</xsl:template>
```

So the edit moves behind `--fix-suggestions`, and it now deletes only the
whitespace and the slashes rather than rebuilding the whole attribute. A
pattern that is nothing but slashes gets no fix at all — emptying it would
trade one broken pattern for another, and patterns reach no XPath validator.

The selector moves off the element and onto the pattern attributes
themselves, which picks up the five it never saw — `xsl:key/@match`,
`xsl:accumulator-rule/@match`, `xsl:number/@count` and `@from`, and the
grouping patterns of `xsl:for-each-group` — and yields one defect per
attribute, so `<xsl:number count="//a" from="//b"/>` reports and fixes both.
Leading whitespace no longer hides the slashes either; `match=" //x"` used to
slip through to `use-double-slash`, which told the author to "use a specific
path" when the `//` was merely redundant. That check normalizes its negation
now so the two stay disjoint.

Both motives are rewritten around the priority shift, the fix tier, and the
line between the two checks. The README sample output was quoting a message
the check stopped emitting some time ago, and the check moves from the `--fix`
list to the suggestions list. Packs grow to the hard-case bar — whitespace,
buried, three or more, negative neighbours, a non-`xsl` prefix, an
`xsl:transform` root, XSLT 1.0/2.0/3.0 — with a new pack for the non-template
patterns. The check carries `mature: true`.

Locally: 590 tests pass, coverage is 100% on all four metrics, and xcop,
yamllint, markdownlint, typos and the fixtures gate are clean.

One thing left alone, worth its own issue: a `//` opening a later union branch
(`match="a | //b"`) still lands in `use-double-slash` with the wrong wording,
and that check is still confined to `xsl:template`.

Closes #583

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,9 +40,10 @@ Given a stylesheet like this:
 xslint points at each problem with its exact position and how to fix it:
 
 ```text
+[WARNING] sheet.xsl(2:1) The '@id' attribute is missing in the 'xsl:stylesheet' element. Declare it to specify the unique identifier explicitly. (missing-id-in-stylesheet)
 [ERROR] sheet.xsl(2:1) The xsl:output instruction is missing. Declare it to specify the serialization format explicitly. (not-using-output)
-[WARNING] sheet.xsl(3:3) The match attribute of xsl:template starts with //, which scans the entire document tree. Use a more specific pattern. (starts-with-double-slash)
 [WARNING] sheet.xsl(4:5) A variable, function, or template has a single-character name. Use a descriptive name that reveals intent. (short-names)
+[WARNING] sheet.xsl(3:23) The pattern starts with //, which is redundant since a pattern already matches at any depth. Remove it, minding that it lowers a template's default priority. (starts-with-double-slash)
 ```
 
 In CI, use the [GitHub Action](https://github.com/xslint/xslint-action) to
@@ -263,9 +264,6 @@ Today this covers nine checks:
   `exists($x)` and `count($x) = 0` becomes `empty($x)`; on 1.0 (where those
   functions don't exist), `boolean($x)` — or a bare `$x` in a whole `@test` —
   and `not($x)`.
-- `starts-with-double-slash` — the redundant leading `//` of a template's
-  `@match` is dropped: `match="//para"` becomes `match="para"`, which selects
-  the same nodes.
 - `redundant-import` — a duplicate `xsl:import`/`xsl:include` of a module
   already imported in the same stylesheet is removed; the module stays imported
   by the first reference.
@@ -305,6 +303,11 @@ Today this covers:
 - `select-starts-with-double-slash` — the leading `//` of a `@select` is
   anchored as `.//`: `select="//title"` becomes `select=".//title"`, scanning
   descendants of the context node instead of the whole document.
+- `starts-with-double-slash` — the redundant leading `//` of a pattern is
+  dropped: `match="//para"` becomes `match="para"`, which selects the same
+  nodes. A suggestion because the priority does not survive the edit: a
+  template matching `//para` defaults to priority 0.5 and one matching `para`
+  to 0, so a rule it used to beat, or tie with, can start winning.
 - `confusing-variable-and-node` — a bare variable name used as a node selector
   gets its `$`: `<xsl:apply-templates select="items"/>` becomes
   `select="$items"`, assuming the variable was meant over a child element.
--- a/src/fixers.js
+++ b/src/fixers.js
@@ -67,21 +67,28 @@ const modeOrPriority = function(node) {
 }
 
 /**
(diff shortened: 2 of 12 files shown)


maxonfjvipon:
Closing this in favour of a different split. The tier demotion — the actual defect #583 reports — lands on its own in #587, small enough to read in one pass.

The rest of this branch presupposes the declarative implementation and would be deleted shortly after merging: detection by `contains()` over raw attribute text cannot tell a step from a `//` inside a string literal or an XPath comment, and cannot see a `//` opening a union branch, so the widened selector still misdiagnoses `match="a | //b"` and still reports `match="a[@n = 'p//q']"`. `src/tokens.js` already resolves all three, and `src/import-linter.js` shows one linter owning two check names. A follow-up will move both double-slash checks onto the lexer and take `mature: true` there.

The branch stays on the remote — the motives and fix fixtures here were worth keeping and are already in #587.

Issue #615, opened 2026-07-31
`isValid` rejects legal whitespace around an axis's `::` in `src/xpath.js`


maxonfjvipon:
`isValid` in `src/xpath.js:128` hands the expression to fontoxpath and takes its
verdict as final, with one exception carved out for the namespace axis.
fontoxpath will not parse whitespace between an axis name and the `::` that
follows it, nor between a node test's name and its `(`, so `select="child ::a"`
is reported as `invalid-xpath-expression` even though XPath 1.0 §3.7 allows
both: an NCName is recognised as an AxisName when the two characters following
it are `::` "possibly after intervening ExprWhitespace", and as a NodeType when
the following character is `(` under the same proviso.

```js
const isValid = function(xpath) {
  return compiles(xpath) ||
    (NAMESPACE_AXIS.test(xpath) &&
      compiles(xpath.replace(/namespace\s*::/g, 'child::')))
}
```

Every spaced spelling fails. `parent ::node()`, `parent:: node()`,
`parent :: node()`, `self :: node()`, `parent::node ()` and `parent::node( )`
all come back false, while the unspaced `parent::node()` comes back true. The
intolerance is specific to axes and node tests — `count (x)` compiles without
complaint. libxml2, an XPath 1.0 engine, accepts all of them; both
`xmllint --xpath '/r/child ::a' t.xml` and
`xmllint --xpath '/r/a[parent::node ()]' t.xml` select the node.

The positioned lexer in `src/tokens.js:179` already reads these forms, matching
the axis name and its colons across the gap, and the code-based linters take
the corpus rather than the validator's filtered expressions, so they still see
`child ::a` and abbreviate it correctly. Only the validator's verdict is wrong,
and it surfaces as an error sitting beside a defect that was reported and fixed
properly.

The rewrite-and-retry that already covers the namespace axis is the smallest
place to put this: fall back to a copy with the whitespace around `::` and
inside a `node ()` test squeezed out, and call the expression valid when that
copy compiles.

maxonfjvipon:
The half of this that is ours is fixed in #630: the fix replaced only as far as
the colons, so `attribute::  name` became `@  name` and `child::  title` became
a value opening with two spaces. XPath puts that gap between the axis and its
node test, so it belongs to the axis and the fix takes it — `@name` and `title`.

Leaving this open for the other half, which is fontoxpath's:

```text
child ::x        isValid -> false
attribute:: x    isValid -> false
parent :: node() isValid -> false
```

[...]

PR #616, opened 2026-07-31
bug(#564): abbreviate self::node(), version-guard the fix

### LINKED ISSUE #564: `unabbreviated-axis` misses `self::node()` and whitespaced axis forms
`unabbreviated-axis` misses two forms. `self::node()` (which abbreviates to `.`) is not among the axes it handles, and whitespaced spellings such as `parent::node ()` or `parent :: node()` slip past the lexer's contiguous-axis matching.

Verified: `select="child::a"` is flagged while `select="self::node()"` beside it is not.

Add `self::node()` -> `.` to the handled set and make the axis / `node()` matching tolerant of surrounding whitespace.

maxonfjvipon:
Closes #564. Does **not** claim `mature: true` — see the last section.

### What was missing

XPath defines four axis abbreviations. The check handled three, and one of those
brittlely:

| Longhand | Short | Before |
| --- | --- | --- |
| `child::name` | `name` | handled |
| `attribute::name` | `@name` | handled |
| `parent::node()` | `..` | matched by a fixed 14-character slice |
| `self::node()` | `.` | missed entirely |

The slice meant the whitespace tolerance PR #599 put into the lexer never
reached the linter: `parent :: node()` tokenized correctly and then failed the
literal comparison. Both halves of #564 are closed — `afterNode` reads the node
test off the token stream, so `parent::node ()`, `parent :: node()` and
`self :: node ( )` all abbreviate.

### The abbreviation is version-dependent

`.` and `..` are an `AbbreviatedStep` in XPath 1.0, and an `AbbreviatedStep`
takes no predicate:

```text
[4]  Step ::= AxisSpecifier NodeTest Predicate* | AbbreviatedStep
[12] AbbreviatedStep ::= '.' | '..'
```

So `self::node()[1]` is legal on 1.0 and `.[1]` is a syntax error there, and
plain `--fix` was turning a working stylesheet into one xsltproc will not
compile. fontoxpath is 3.1 and accepts `.[1]`, so the validator could not catch
what the fixer had written. The linter now reads `versionOf` and withholds the
fix — report only — when a predicate follows the step on a pre-2.0 sheet:

| | 1.0 | 2.0 |
| --- | --- | --- |
| `self::node()[1]` | reported, not fixed | `.[1]` |
| `parent::node()[1]` | reported, not fixed | `..[1]` |
| `self::node()` | `.` | `.` |
| `child::a[1]` | `a[1]` | `a[1]` |
| `(self::node())[1]` | `(.)[1]` | `(.)[1]` |

The last two rows are the ones worth checking: `@a[1]` and `a[1]` are full steps
and take predicates on 1.0, and a parenthesised `(.)[1]` is a filter expression,
legal in every version. Only the bare abbreviated step is restricted. A sheet
with no readable version is treated as old and keeps its longhand.

### A false positive with a destructive fix

`opensAxis` matched at any offset, so the `child::` at the tail of a longer name
read as an axis and `--fix` rewrote `ns:child::x`. The guard for that is
`opensStep`, and it is called only from `afterOther`: mid-run a match can land
inside a name, but at a token boundary the preceding character belongs to the
previous token, and guarding there loses real steps — `count(a)-child::b` and
`1-child::b` both compile and neither draws `invalid-xpath-expression`.

### Deliberately out of scope

`descendant-or-self::node()` also abbreviates, to `//`. Not adopted: a leading
`//` in a `select` rescans the whole document, which is what
`select-starts-with-double-slash` exists to flag, so recommending it here would
have two checks contradict each other. The motive teaches the trade and points
at `.//`; a pack pins that `a/descendant-or-self::node()/b` stays quiet.

### Coverage

Twelve new packs, taking the check from 4 to 16. They exercise the construct
buried in a predicate, three times in one expression, and beside the lookalikes
that must not fire — `self::a`, `self::*`, `self::text()`, `parent::node` with no
parentheses, an axis in a string literal and in an XPath comment. Per the
node-kind rule from #610 they also cover a non-`xsl` prefix on an `xsl:transform`
root, a simplified 1.0 stylesheet whose literal result element carries a `select`
bound for the result tree (left alone) beside an AVT that is read, a 3.0 text
value template, a shadow attribute, and a CDATA section. `unabbreviated-axis-in-xslt-1.xsl`
joins `UNCHANGED` under `--fix`, so a fix that creeps back onto 1.0 fails the build.

`using-namespace-axis` shares the lexer, so its one behavior delta here —
`my:namespace::x`, a false positive removed — is recorded in
`namespace-axis-inside-a-name.yaml`. That pack also pins
`count(a)-namespace::b`, which the first push had regressed. Its `mature: true`
comes off in this PR as well: `ornamespace::x` and `exceptnamespace::x` are live
false positives on master, so the attestation was already untrue, and #617 is now
the gate for both checks.

### Why the flag is not set

Writing the pack for `grandchild::x` showed the guard cannot reach it, and the
reason is not the axis code at all:

```text
border  -> ["b", "or", "der"]      grandchild -> ["gr", "and", "child"]
agent   -> ["a", "ge", "nt"]       exception  -> ["except", "ion"]
```

`afterOther` breaks a run at any word operator, so the axis ends up at a genuine
token boundary. Filed as #617. `--fix` still rewrites `grandchild::x` to
`grandx` — malformed input, already reported as `invalid-xpath-expression`, but
the file is edited all the same, so "no false positive" is not true yet. #617 is
the gate, and it names the one dimension #613's audit of `using-namespace-axis`
did not cover.

Also filed: #615, the validator calling a legal whitespaced axis malformed. It
does not reach this check, which is a document linter and reads the corpus
rather than the validator's filtered expressions.

883 tests, 100% branch coverage (920/920), ESLint, xcop, markdownlint and the
fixtures gate all clean locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,7 +253,8 @@ Today this covers ten checks:
 - `redundant-whitespace` — a doubled space is collapsed to one, and a space
   leading or trailing an XPath expression is removed.
 - `unabbreviated-axis` — a verbose axis specifier is shortened: `child::x`
-  becomes `x`, `attribute::x` becomes `@x`, and `parent::node()` becomes `..`.
+  becomes `x`, `attribute::x` becomes `@x`, `parent::node()` becomes `..`, and
+  `self::node()` becomes `.`.
 - `redundant-namespace-declarations` — a namespace prefix declared on the
   stylesheet but never used is deleted.
 - `use-node-set-extension` — the redundant `node-set()` extension is unwrapped
--- a/README.md
+++ b/README.md
@@ -254,7 +254,8 @@ Today this covers ten checks:
   leading or trailing an XPath expression is removed.
 - `unabbreviated-axis` — a verbose axis specifier is shortened: `child::x`
   becomes `x`, `attribute::x` becomes `@x`, `parent::node()` becomes `..`, and
-  `self::node()` becomes `.`.
+  `self::node()` becomes `.`. On a 1.0 stylesheet a predicated step keeps its
+  longhand, since `.` and `..` took no predicate before XPath 2.0.
 - `redundant-namespace-declarations` — a namespace prefix declared on the
   stylesheet but never used is deleted.
 - `use-node-set-extension` — the redundant `node-set()` extension is unwrapped
--- a/src/resources/checks/format/unabbreviated-axis.yaml
+++ b/src/resources/checks/format/unabbreviated-axis.yaml
@@ -3,4 +3,3 @@
 ---
 severity: warning
 message: Verbose axis specifiers child::, attribute::, parent::node(), or self::node() are used. Replace them with the node name alone, @, .., or . respectively.
-mature: true
--- a/src/resources/checks/format/unabbreviated-axis.yaml
+++ b/src/resources/checks/format/unabbreviated-axis.yaml
@@ -2,4 +2,5 @@
 # SPDX-License-Identifier: MIT
 ---
 severity: warning
-message: Verbose axis specifiers child::, attribute::, or parent::node() are used. Replace them with the node name alone, @, or .. respectively.
+message: Verbose axis specifiers child::, attribute::, parent::node(), or self::node() are used. Replace them with the node name alone, @, .., or . respectively.
+mature: true
--- a/src/resources/checks/format/using-namespace-axis.yaml
+++ b/src/resources/checks/format/using-namespace-axis.yaml
@@ -3,4 +3,3 @@
 ---
 severity: warning
 message: "The namespace:: axis is deprecated in XSLT 2.0 and later. Use in-scope-prefixes() and namespace-uri-for-prefix() instead."
-mature: true
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -9,8 +9,8 @@ competing for the reader's attention. Worse, `child::` is noise that says
 nothing — the child axis is the one a bare name already uses, so writing it out
 draws the eye to the one part of the step that carries no information.
 
-The abbreviations are exact synonyms defined by XPath itself, so the shorter
-form selects precisely the same nodes:
+The abbreviations are defined by XPath itself and select precisely the same
+nodes as the longhand:
 
 | Longhand | Short |
 | --- | --- |
@@ -43,6 +43,16 @@ Those four are the whole of the trade. `parent::chapter` and `self::chapter`
 name an element rather than any node at all, so they have no shorter spelling
 and stay as they are, and the remaining axes — `descendant::`, `ancestor::`,
 `following-sibling::` and the rest — were never given one.
+
+The node-sets match, but the grammars do not quite, and on XSLT 1.0 that shows.
+There `.` and `..` are an *abbreviated step*, a production that takes no
+predicate, while the longhand is a full step and a full step accepts them. So
+`self::node()[1]` is legal in a 1.0 stylesheet and `.[1]` is a syntax error in
+it — XPath 2.0 gave the context item its own predicate list and lifted the
+restriction, which is why the same expression compiles under a later `version`.
+Where a predicate follows the step in a 1.0 stylesheet, either leave it spelled
+out or parenthesise the abbreviation, since `(.)[1]` is a filter expression and
+is legal in every version.
 
 `descendant-or-self::node()` is the exception that is not worth taking. Its
 short form is `//`, but the two are easy to confuse in use: a `//` that opens a
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -1,8 +1,23 @@
 # Can use abbreviated axis specifier
 
-XSLT supports short forms for common axis specifiers: `child::` can be
-omitted entirely, `attribute::` is abbreviated as `@`, and `parent::node()`
-is written as `..`.
+The steps an XPath uses most often have short forms, and those short forms are
+what every XSLT reader scans for. Spelling them out buries the shape of the
+path under axis ceremony: `child::chapter/child::section/attribute::id` selects
+exactly what `chapter/section/@id` selects, in nearly three times the width,
+and the difference only grows inside a predicate, where the path is already
+competing for the reader's attention. Worse, `child::` is noise that says
+nothing — the child axis is the one a bare name already uses, so writing it out
+draws the eye to the one part of the step that carries no information.
+
+The abbreviations are exact synonyms defined by XPath itself, so the shorter
+form selects precisely the same nodes:
+
+| Longhand | Short |
+| --- | --- |
+| `child::name` | `name` |
+| `attribute::name` | `@name` |
+| `parent::node()` | `..` |
+| `self::node()` | `.` |
 
 Incorrect:
 
@@ -10,6 +25,8 @@ Incorrect:
 <xsl:value-of select="child::title"/>
 <xsl:value-of select="attribute::name"/>
 <xsl:apply-templates select="parent::node()"/>
+<xsl:copy-of select="self::node()"/>
+<xsl:value-of select="child::book/child::author/attribute::id"/>
 ```
 
 Correct:
@@ -18,4 +35,18 @@ Correct:
 <xsl:value-of select="title"/>
 <xsl:value-of select="@name"/>
 <xsl:apply-templates select=".."/>
+<xsl:copy-of select="."/>
+<xsl:value-of select="book/author/@id"/>
 ```
+
+Those four are the whole of the trade. `parent::chapter` and `self::chapter`
+name an element rather than any node at all, so they have no shorter spelling
+and stay as they are, and the remaining axes — `descendant::`, `ancestor::`,
+`following-sibling::` and the rest — were never given one.
+
+`descendant-or-self::node()` is the exception that is not worth taking. Its
+short form is `//`, but the two are easy to confuse in use: a `//` that opens a
+`select` walks the whole document from the root on every evaluation, which is
+rarely what the longhand step meant. Where the abbreviation is wanted there,
+anchor it — `.//author` searches the current subtree, `//author` searches the
+entire document.
--- a/src/tokens.js
+++ b/src/tokens.js
@@ -168,6 +168,14 @@ const opensComment = function(xpath, at) {
 }
 
 /**
+ * Characters a name is spelled with. An axis name only ever opens a step, so
+ * one of these just before it means the match landed inside a longer name —
+ * the `child::` that ends `grandchild::` — and no axis starts there.
+ * @type {RegExp}
+ */
+const NAMED = /[\w.:-]/
+
+/**
  * The axis opening at the given offset, or null when none does. XPath allows
  * whitespace between the axis name and its `::`, so `child ::` names the same
  * axis as `child::`; the name and the colons are matched across that gap and
@@ -187,7 +195,7 @@ const opensAxis = function(xpath, at) {
   }
   const name = `${xpath.slice(at, end)}::`
   if (end === at || xpath[colons] !== ':' || xpath[colons + 1] !== ':' ||
-    !AXES[name]) {
+    !AXES[name] || (at > 0 && NAMED.test(xpath[at - 1]))) {
     return null
   }
   return {name: name, length: colons + 2 - at}
--- a/src/tokens.js
+++ b/src/tokens.js
@@ -168,9 +168,7 @@ const opensComment = function(xpath, at) {
 }
 
 /**
- * Characters a name is spelled with. An axis name only ever opens a step, so
- * one of these just before it means the match landed inside a longer name —
- * the `child::` that ends `grandchild::` — and no axis starts there.
+ * Characters a name is spelled with.
  * @type {RegExp}
  */
 const NAMED = /[\w.:-]/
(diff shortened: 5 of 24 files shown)


maxonfjvipon:
Not mature yet, and one of the two problems is a `--fix` that destroys a working
stylesheet. The `self::node()` work itself is right; the trouble is what it is
tiered as, and what the `tokens.js` guard costs.

### 1. The safe fix breaks XSLT 1.0

`.` and `..` are `AbbreviatedStep` in the XPath 1.0 grammar, and
`AbbreviatedStep` takes no predicates:

```text
[4]  Step ::= AxisSpecifier NodeTest Predicate* | AbbreviatedStep
[12] AbbreviatedStep ::= '.' | '..'
```

The longhand is a full `Step` and does take them, so the abbreviation is not a
drop-in wherever a predicate follows. libxslt agrees:

```text
XPath1.0 accept  self::node()[1]      XPath1.0 REJECT  .[1]
XPath1.0 accept  parent::node()[1]    XPath1.0 REJECT  ..[1]
```

End to end on a `version="1.0"` sheet, with plain `--fix`, no suggestions flag:

```diff
 [INFO] Fixed 2 defects in 1 files
-    <xsl:copy-of select="self::node()[1]"/>
+    <xsl:copy-of select=".[1]"/>
-    <xsl:copy-of select="parent::node()[1]"/>
+    <xsl:copy-of select="..[1]"/>
```

```text
$ xsltproc one.xsl in.xml
XPath error : Invalid expression
.[1]
 ^
xsl:copy-of : could not compile select expression '.[1]'
```

`src/xpath-axis-linter.js` never imports `xsl-version` — the check has no
version guard at all. Detection is right to be version-blind; the fix is not.
This is the mandatory rule verbatim: never emit a fix the declared version
cannot run. `.[…]` and `..[…]` are XPath 2.0 constructs, where `ContextItemExpr`
gained a `PredicateList`, and fontoxpath, being 3.1, accepts them — so xslint's
own validator cannot catch what xslint's own fixer just wrote. `//.`, `a/.` and
a bare `.`/`..` are all fine on 1.0; the whole hazard is the predicate.

The `parent::node()` half is pre-existing, but maturity is a claim about the
check as it stands, so both land here. Guard it on `versionOf`: withhold the fix
when a predicate follows the step and the sheet is 1.0. The check still reports,
and report-only for that one occurrence is the version-appropriate answer.

### 2. The `NAMED` guard loses valid XPath

`NAMED = /[\w.:-]/` includes `-`, and a `-` before an axis is not always
name-internal. After `)` or a number it is the subtraction operator, and the
step that follows is a real step:

| expression | master | this PR | libxslt 1.0 |
| --- | --- | --- | --- |
| `count(a)-child::b` | 1 | 0 | compiles |
| `1-child::b` | 1 | 0 | compiles |
| `count(a) - child::b` | 1 | 1 | compiles |

Master flags these, the branch misses them, and both are accepted by a real
XPath 1.0 engine and by fontoxpath — no `invalid-xpath-expression` co-fires.
That is a false negative on valid input, which the bar rules out.

Worth noting what the guard is buying in exchange. Every example in the
description — `grandchild::x`, `a-child::b`, `ns:child::x` — is malformed XPath,
already reported as `invalid-xpath-expression`. The bug is real, `--fix` should
not mangle a broken expression into `grandx`, but it is confined to input the
run has already rejected, while the cost falls on input that compiles.

The guard is also in the wrong place. `tokenized` calls `opensAxis` only at a
token boundary, where the guard can never help and can only hurt; `afterOther`
calls it mid-run, which is the only place a match can land inside a longer name.
Move the test into `afterOther` and every case comes out right: `a/child::x`
still breaks the run, since `/` precedes it, `grandchild::x` and `a-child::b`
still do not, and `count(a)-child::b` is recognized at the boundary as before.

### 3. It silently thaws a frozen check

`src/tokens.js` is shared, so this changes `using-namespace-axis` — frozen
`mature: true` by #613, merged three commits ago — in both directions:

| expression | master | this PR | |
| --- | --- | --- | --- |
| `mynamespace::x` | 1 | 0 | false positive removed; the expression is malformed anyway |
| `count(a)-namespace::b` | 1 | 0 | false negative added; the expression compiles |

`CLAUDE.md` says a mature check changes only by first removing the flag and
re-earning it. Fixing the point above removes the regression, but the
false-positive half still alters a frozen check and belongs in the record rather
than slipped in under another rule's PR. It also says something about #613: that
audit tested a prefix named `namespace` but never a name ending in it, so the
false positive was there and went unseen.

### What is right

Everything else holds up. `afterNode` reads the node test off the token stream
instead of the 14-character slice, so #564's real complaint is closed —
`parent::node ()`, `parent :: node()` and `self :: node ( )` all abbreviate, and
`SPAN = 6` is exactly the token budget those forms need. `self::node` without
parentheses, `self::a`, `self::*`, `self::text()` and `ancestor-or-self::node()`
all stay quiet. Leaving `descendant-or-self::node()` alone rather than
recommending `//` is the right call, and the motive explains that trade
honestly. The nine new packs cover the construct buried in a predicate, three in
one expression, the lookalikes, and every node kind #610 added. Coverage is 100%
(914/914 branches) and 641 tests pass locally.

The gap in the packs mirrors the gap in the code: not one fixture is a 1.0
stylesheet, and not one puts a predicate after `self::node()` or
`parent::node()`. Add that row and `fixer.test.js` fails today.

One smaller thing, for when the first point is fixed: the motive's "exact
synonyms defined by XPath itself, so the shorter form selects precisely the same
nodes" is true of the node-set and false of the syntax — `.` cannot stand where
`self::node()` stands in a 1.0 predicate. Worth a clause.

Guard the fix on version, move the `NAMED` test into `afterOther`, add the
1.0-with-predicate fixture, and record what happened to `using-namespace-axis`.
Then the flag is earned.

maxonfjvipon:
All three confirmed, all three fixed. The flag comes off — a fourth problem fell
out of the second one and it is not mine to fix here.

### 1. The safe fix breaks XSLT 1.0 — right, and worse than described

Reproduced exactly, and libxml2 agrees with your grammar reading:

```text
/r/a[self::node()[1]]     libxml2 OK
/r/a[.[1]]                XPath error : Invalid predicate
/r/a[..[1]]               XPath error : Invalid predicate
```

`versionOf` now gates it. `afterNode` already walked the tokens after the node
test, so it reports whether an `[` follows; when one does and the sheet is not
2.0 or later, the defect ships with no fix at all. Detection stays version-blind.

```text
1.0  self::node()[1]   reported, not fixed     parent::node()[1]  reported, not fixed
1.0  self::node()      -> .                    child::a[1]        -> a[1]
2.0  self::node()[1]   -> .[1]                 parent::node()[1]  -> ..[1]
```

The 1.0 sheet compiles under xsltproc after `--fix`, which it did not before.
Two packs pin both sides, and `unabbreviated-axis-in-xslt-1.xsl` joins
`UNCHANGED` with `--fix` rather than `--fix-dry-run`, so a fix that creeps back
fails the build.

### 2. The `NAMED` guard loses valid XPath — right, and the diagnosis was right too

Your table reproduces, and `invalid-xpath-expression` does not co-fire on any of
the three. The guard is now `opensStep`, called only from `afterOther`, exactly
as you said: mid-run it suppresses, at a boundary `tokenized` asks `opensAxis`
directly.

| expression | master | first push | now |
| --- | --- | --- | --- |
| `count(a)-child::b` | 1 | 0 | 1 |
| `1-child::b` | 1 | 0 | 1 |
| `count(a)-namespace::b` | 1 | 0 | 1 |
| `ns:child::x` | 1 | 0 | 0 |

### 3. It silently thaws a frozen check — and there is a fourth thing

Chasing the `grandchild::x` example to write its pack turned up why the relocated
guard does not catch it. It never was a name-character problem:

```text
border      -> ["b", "or", "der"]
agent       -> ["a", "ge", "nt"]
grandchild  -> ["gr", "and", "child"]
modal       -> ["mod", "al"]
exception   -> ["except", "ion"]
```

`afterOther` breaks a run at any word operator, so `grandchild` lexes as
`gr`/`and`/`child::` and the axis lands at a genuine token boundary, where the
guard correctly declines to interfere. `a-child::b` is the same shape with the
minus. Filed as #617 with the fix — require that neither neighbour of a word
operator is a name character, the same test `opensStep` applies one function up.

So the flag comes off. `--fix` still rewrites `grandchild::x` to `grandx`; the
input is malformed and already reported, but the file is still edited, and I am
not going to claim "no false positive" twice. #617 is the gate.

On the frozen check: after the relocation the only remaining delta to
`using-namespace-axis` is `my:namespace::x`, a false positive removed. It is
recorded rather than slipped in — `namespace-axis-inside-a-name.yaml` pins both
that and `count(a)-namespace::b`. #617 notes the dimension #613 missed; whether
that warrants un-freezing is your call, since the same lexer defect gives it
`grandnamespace::x` today.

### Smaller points

The motive's "exact synonyms" clause is gone, replaced by a paragraph on why the
node-sets match and the grammars do not, ending in the hand-fix: `(.)[1]` is a
filter expression and libxml2 accepts it on 1.0.

One correction to the packs note — `axis-in-simplified-stylesheet.yaml` was
already a `version="1.0"` sheet. You are right that none had a predicate after
the step; both now do.

883 tests, 100% branches (920/920).

maxonfjvipon:
All three points are closed, and #617 is a better finding than any of them —
`grandchild -> ["gr", "and", "child::", "x"]` is the actual mechanism, and my
`NAMED`-in-`afterOther` suggestion only ever hid part of it. Verified on
`b1c7384`:

```text
                                    before   now
self::node()[1]   on 1.0            .[1]     reported, no fix
parent::node()[1] on 1.0            ..[1]    reported, no fix
self::node()[1]   on 2.0            .        .
count(a)-child::b                   missed   reported
count(a)-namespace::b               missed   reported
```

100% branch coverage (920/920), conformance green, and
`unabbreviated-axis-in-xslt-1.xsl` in `UNCHANGED` under `--fix` — so a fix that
creeps back onto 1.0 fails the build, which is the right shape for that guard.
Two things left.

### `using-namespace-axis` should lose its flag in this PR

It still carries `mature: true`, and #617's own last line says the frozen flag
rests on it too. It is not a latent risk — the false positive is live on both
master and this head:

```text
ornamespace::x       using-namespace-axis
exceptnamespace::x   using-namespace-axis
```

`or` and `except` split the name, `namespace::` lands at a fresh boundary, and a
check that advertises "no false positive" reports one. On top of that, this
branch already changes what that check does — `mynamespace::x` and
`my:namespace::x` were reported on master and are silent now — and
`CLAUDE.md` is explicit that a mature check changes only by first removing the
flag. Recording the delta in `namespace-axis-inside-a-name.yaml` is the right
instinct and the pack should stay, but the pack is evidence, not the
attestation. The flag is the attestation, and it is wrong today.

Dropping it here costs one line and leaves the tree honest: two checks with one
defect, both waiting on #617, neither claiming to be finished. Leaving it set
means nobody looks at that check again, because frozen means do not re-audit.

### #617 will not unblock the flag as scoped

Its closing paragraph says the symbol operators are unaffected "since `(`, `)`,
`/` and the rest cannot appear inside a name to begin with". `-` can — it is a
legal `NCName` character after the first, and it is also in `SINGLE`:

```text
a-child::b -> ["a", "-", "child::", "b"]
```

`afterOther` stops at the `-` because of the `!SINGLE[xpath[at]]` test, which is
a different branch from the word-operator one #617 proposes to guard. So
`child::` still reaches a genuine token boundary, `opensAxis` still fires, and
`--fix` still rewrites the expression — to `a-b` this time rather than `grandx`.
Requiring that neither neighbour of a word operator is a name character leaves
that untouched.

Two rows quietly moved to record this, and both are worth keeping in view: the
first push had `a-child::b` in `axis-name-inside-a-name.yaml` and
`['a-child::abc', TOKENS.CHILD]` in the "finds no axes" table of
`tokens.test.js`, and this push dropped each of them alongside the
`grandchild::x` pair. `grandchild::x` is named in #617; `a-child::b` is not, and
it fails through a door #617 currently declares safe.

Worth widening #617 to "a name character on either side of an operator that can
appear inside a name", covering the word operators and `-` in one rule, before
the work starts — otherwise the issue gets closed and the flag is still blocked
by a case nobody wrote down.

maxonfjvipon:
Both right. Done in `41feb44`.

### `using-namespace-axis` loses its flag

Verified live on master and on this head before touching it:

```text
ornamespace::x       -> ["or", "namespace::", "x"]       using-namespace-axis
exceptnamespace::x   -> ["except", "namespace::", "x"]   using-namespace-axis
```

The distinction between evidence and attestation is the part I had wrong. I
treated the pack as discharging the obligation; it records what the check does,
it does not certify that what it does is right. The flag is gone, the pack
stays.

Worth noting these two arrive through a door `border` and `grandchild` do not.
`or` and `except` start the name, so `tokenized` matches them before
`afterOther` is ever reached — a guard confined to the run scanner would leave
both reporting. That is now written into #617 as a separate case.

### #617 was wrong about the symbol operators, and the obvious rule has a trap

`a-child::b -> ["a", "-", "child::", "b"]` reproduces, through
`!SINGLE[xpath[at]]` rather than the word-operator branch, exactly as you say.
`--fix` writes `a-b`.

Chasing that turned up why the rule has to be phrased carefully:

```text
a-child::b -> ["a", "-", "child::", "b"]    one name, the "-" is internal
1-child::b -> ["1", "-", "child::", "b"]    a number minus a step, both real
```

Identical token shapes. "A name character on either side" is true of both, since
a digit is a name character, so that phrasing would suppress `1-child::b` — valid
XPath, accepted by libxml2, no `invalid-xpath-expression`, and the exact
false negative the first push was pulled up for.

What separates them is not the characters but the run: `1` is a `NUMBER`,
because `opensNumber` wins before `afterOther` is called, and `count(a)` ends in
`)`. So the `-` test belongs inside `afterOther` alone, where a run always began
at a name-start character and never a digit. `axis-after-an-operator.yaml`
already pins all three, so an overreaching fix fails the build.

#617 is rewritten around the three doors — word operator mid-run, word operator
at the start of a token, `-` mid-run — with that scoping spelled out, and it now
names `a-child::b` alongside `grandchild::x`. On the two rows I dropped: fair
hit. They came out when the relocated guard stopped suppressing them and I
recorded `grandchild::x` in the issue but not `a-child::b`, which left the
weaker case with no home. Both are in #617 now.

Two checks, one defect, neither claiming to be finished.

883 tests, 100% branches (920/920).

maxonfjvipon:
Both points closed on `41feb44`. Nothing left blocking — this is good to merge.

Verified on that head: no `mature: true` remains anywhere under
`src/resources/checks/`, so neither check claims a freeze it has not earned and
both wait on #617 in the open. The version guard holds — a predicated step is
reported with no fix on 1.0 and abbreviated on 2.0 — `count(a)-child::b`,
`1-child::b` and `count(a)-namespace::b` are all reported again, 647 tests pass,
100% branch coverage (920/920), ESLint clean, and `npx grunt docs` leaves no
diff.

The false positives that remain — `grandchild::x`, `a-child::b`,
`ornamespace::x` — are all confined to malformed XPath, all already reported as
`invalid-xpath-expression`, and all recorded in #617 with nothing in the tree
saying otherwise. That is the honest state.

One note left on #617 rather than here: its scoping argument rests on an
`afterOther` run always beginning at a name-start character, and `.` is an
exception — `.-child::b` is valid XPath that the lexer handles correctly today
and that the proposed `-` rule would break. Details on the issue, since that is
where the fix gets written.

Issue #618, opened 2026-08-01
`versionOf` reads only the root, missing a subtree effective version in `src/xsl-version.js`


maxonfjvipon:
`versionOf` in `src/xsl-version.js:29` reads the declared version from
`xsl.documentElement` and nowhere else:

```js
const versionOf = function(xsl) {
  const root = xsl.documentElement
  return (root.namespaceURI === XSLT ?
    root.getAttribute('version') : root.getAttributeNS(XSLT, 'version')) || ''
}
```

XSLT 2.0 and 3.0 make `version` a standard attribute that may sit on any XSLT
element, and `xsl:version` one that may sit on any literal result element; each
sets the effective version of that element and everything below it. A stylesheet
that declares `1.0` at the root and raises one template to `2.0` is therefore
judged, everywhere, by the root.

What follows is a false negative in every code-based linter gating on the
reader. `src/using-namespace-axis-linter.js:58` asks
`MODERN.includes(versionOf(xsl))` before it looks at anything, so this sheet
reports nothing, though the deprecated axis in it stands in a 2.0 context:

```xsl
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="1.0">
  <xsl:template match="/" version="2.0">
    <xsl:copy-of select="namespace::x"/>
  </xsl:template>
</xsl:stylesheet>
```

The literal-result spelling is silent for the same reason: `<html
xsl:version="2.0">` wrapping that same `xsl:copy-of` inside a 1.0 sheet yields
no defect either. `count`, `name`, `node-set`, and `translate` inherit the gap,
and `src/attributes.js:127` inherits it one layer deeper — `const three =
versionOf(xsl) === '3.0'` decides whether text value templates and shadow
attributes are read at all, again from the root alone, so a 3.0 subtree of a 2.0
sheet contributes no expressions to any linter at all. The mirror case falls out
of the same line: a subtree lowered to `1.0` inside a 2.0 sheet, where the
flagged construct is standard and the report is a false positive.

This is the shape of defect #603 fixed one level up, where the root's own
version was read from the wrong attribute. The reader learned to ask which kind
of root it had, but not to ask anything below it.

Reading the version from the node under judgement rather than from the document
covers the class: let `versionOf` take a node and walk up to the nearest
ancestor carrying `version` on an XSLT element or `xsl:version` on a literal
result element, falling back to the root as today — the same ancestor walk
[...]

Issue #617, opened 2026-07-31
`src/tokens.js` splits an ordinary name at an operator it happens to contain


maxonfjvipon:
`src/tokens.js` ends a token at anything that looks like an operator, without
asking whether a name is in progress. Names that merely contain an operator's
letters are therefore torn apart:

```text
border      -> ["b", "or", "der"]
agent       -> ["a", "ge", "nt"]
grandchild  -> ["gr", "and", "child"]
modal       -> ["mod", "al"]
unionist    -> ["union", "ist"]
exception   -> ["except", "ion"]
orange      -> ["or", "ange"]
a-child     -> ["a", "-", "child"]
```

Every one of those is a legal XML name and a legal XPath node test. `border` is
one token to any XPath engine; here it is three, and the middle one claims to be
the `or` operator.

Most linters read past the wreckage because they look for a token type the split
does not produce, so the damage stays latent — until a split leaves an axis name
at a fresh token boundary, where `opensAxis` is asked directly and answers yes:

```text
grandchild::x      -> ["gr", "and", "child::", "x"]   unabbreviated-axis, --fix writes "grandx"
a-child::b         -> ["a", "-", "child::", "b"]      unabbreviated-axis, --fix writes "a-b"
ornamespace::x     -> ["or", "namespace::", "x"]      using-namespace-axis
exceptnamespace::x -> ["except", "namespace::", "x"]  using-namespace-axis
```

All four are malformed XPath and are already reported as
`invalid-xpath-expression`, so the blast radius is confined to input the run has
rejected — but `--fix` still edits the file, turning one broken expression into a
different broken expression.

## Three doors, not one

The splits arrive through three different branches, and a fix that closes one
leaves the others open.

**A word operator mid-run.** `afterOther` at `src/tokens.js:345` breaks on
`DOUBLE`, `TRIPLE` and `MORE`. This is `border`, `grandchild`, `modal`.

**A word operator at the start of a token.** `tokenized` tests the same three
maps before falling through to `afterOther`, so a name that *begins* with one
never reaches the run scanner at all. This is `orange` and `ornamespace::x`.
Guarding only `afterOther` misses these.

**A `-` mid-run.** `-` is in `SINGLE`, so `afterOther` stops at it through the
`!SINGLE[xpath[at]]` test — a different branch from the word-operator one. But
`-` is a legal `NCName` character after the first, so `a-child` is one name.
This is the case the word-operator rule does not reach.

[...]

Issue #614, opened 2026-07-31
An unrecognized `@version` silently disables every version gate in `src/xsl-version.js`


maxonfjvipon:
Every version gate in the tree matches the declared version as a string.
`src/xsl-version.js:17` holds the whole vocabulary:

```js
const MODERN = ['2.0', '3.0']
```

`count`, `name`, `node-set`, `translate`, and `using-namespace-axis` each ask
`MODERN.includes(versionOf(xsl))` (`src/count-linter.js:121`), and
`src/attributes.js:127` gates text value templates and shadow attributes on
`versionOf(xsl) === '3.0'`. Anything outside those two literals is therefore read
as pre-2.0, and nothing anywhere says so. `@version` is `xs:decimal` in the W3C
schema for XSLT, so `3.00`, `2`, and a value carrying the whitespace the spec
tells a processor to ignore are all legal spellings of versions that exist; a
typo like `2,0` is not legal but reaches the same place.

What follows is silent. On `version="3.00"` the text value template scan finds
nothing where the identical sheet at `3.0` reports a defect. On `version="2"`,
`count-compared-to-zero` still fires but rewrites `count(i) = 0` to `not(i)`
rather than `empty(i)` — and that fix is safe-tier, so `--fix` writes the 1.0
idiom into a 2.0 stylesheet without asking. The mandatory version-dependence rule
in `CLAUDE.md` asks for the opposite: emit the version-appropriate form. A reader
of the report sees one thing go wrong at most, never that five checks and a whole
expression surface quietly switched off.

Parsing the version numerically would paper over the good spellings and make the
bad ones worse: `2.00` and `2` would start working, a future `4.0` would be
silently guessed as modern, and `2,0` would parse to `NaN` and be silently
treated as ancient — the same silence, one layer down. The tool should say what
it does not recognize. Nothing checks the value today; the neighbouring
`missing-version-in-stylesheet` tests only that the attribute is present.

An `xpath` check on the version value, at `error` severity so the run cannot
quietly continue with reduced coverage, covers the whole class in one place: it
flags any `@version`/`@xsl:version` that is not `1.0`, `2.0`, or `3.0`, carries a
safe fix where the value is the same decimal as one of those (`3.00` becomes
`3.0`), and stays report-only for a version xslint has never heard of. It needs
the usual root guard, and it must leave an absent version alone so it does not
double-report against `missing-version-in-stylesheet`. This supersedes #612,
which proposed the numeric comparison instead.

[...]

Issue #619, opened 2026-08-01
`using-namespace-axis` motive rests on an unsourced processor claim and credits XSLT 2.0 for an XPath 2.0 deprecation


maxonfjvipon:
The motive for `using-namespace-axis` opens by telling the reader their
stylesheet may be unrunnable, at
`src/resources/motives/format/using-namespace-axis.md:3`:

```md
The `namespace::` axis is deprecated in XSLT 2.0 and remains so in 3.0, where
some processors drop it entirely — a stylesheet that leans on it may not run at
all.
```

That is the entire harm the check rests on, and it has no source. XPath 3.1
§3.3.2.1 puts the decision on the host language, not on the processor: "As of
XPath 2.0, the namespace axis is deprecated and need not be supported by a host
language." XSLT is that host language, and nothing in the XSLT 3.0 conformance
chapter releases an XSLT processor from supporting the axis — a conforming
processor must implement XPath 1.0 backwards-compatible behaviour, which is the
one mode XPath 2.0 requires the axis in. Saxon, libxslt, and Xalan all support
it. So the sentence a user is asked to rewrite working code over describes a
failure nobody has been shown.

The attribution is off by one specification too, in the motive and again in the
reported message at
`src/resources/checks/format/using-namespace-axis.yaml:5`:

```yaml
message: "The namespace:: axis is deprecated in XSLT 2.0 and later. Use in-scope-prefixes() and namespace-uri-for-prefix() instead."
```

XSLT 2.0 deprecates nothing here; XPath 2.0 does, and XSLT 2.0 inherits it by
adopting that XPath. A reader who goes looking for the deprecation in the XSLT
spec will not find it.

The real harm is the one the rest of the motive already teaches well: namespace
nodes are a data-model wart that XPath 2.0 pushed out in favour of
`in-scope-prefixes()` and `namespace-uri-for-prefix()`, the axis is the only
part of the 1.0 data model deprecated outright, and code leaning on it is
writing against a construct the working group has been retiring for two decades
— with the migration made awkward by the fact that no function returns a
namespace *node*, so an `xsl:copy-of` of one has to become `xsl:namespace`. That
is a portability and longevity argument, and it stands on its own without the
unsourced claim about processors.

The check itself is right and its severity is right; only the prose overreaches.
This is the same family as #552, which sweeps motives that assert a rule without
a why or contradict their check, but that issue names neither this motive nor
this kind of defect — a stated harm that the specification does not support — so
[...]

PR #620, opened 2026-08-01
bug(#617): no axis from a name that carries one

maxonfjvipon:
Fixes the false positive half of #617. Leaves #617 open, and does **not** set
`mature: true` — #618 and #614 both block, see the last section.

### The rule

An axis name only ever opens a step, so one reached with a name already in
progress is the tail of that name. `opensAxis` now walks the run of name
characters behind the match back to its start and declines when that run begins
the way a name may:

| expression | before | now |
| --- | --- | --- |
| `grandchild::x` | reported, `--fix` writes `grandx` | silent |
| `a-child::b` | reported, `--fix` writes `a-b` | silent |
| `item2child::x` | reported, `--fix` writes `item2x` | silent |
| `ornamespace::x` | reported | silent |
| `exceptnamespace::x` | reported | silent |
| `h1namespace::x` | reported | silent |
| `caféchild::x` | reported | silent |
| `count(a)-child::b` | reported | reported |
| `1-child::b` | reported | reported |
| `count(a)-namespace::b` | reported | reported |

The run test is the whole trick, because the last three tokenize just like the
first ones. `a-child::b` is one name — a `-` continues a name a letter started.
`1-child::b` subtracts from a number, and a name may not begin with a digit.
`count(a)-child::b` reaches its `-` with no run behind it at all.

Reading the raw string rather than the previous token is deliberate. A review
suggestion on #616 was to guard only where a run is scanned, on the reasoning
that a token boundary is always trustworthy — but a boundary can itself be the
product of a split, which is exactly how `grandchild::x` and `ornamespace::x`
got through. The guard does not care how the characters in front were tokenized.

XML names run well past ASCII, so the name classes are `\p{L}`, `\p{N}` and
`\p{M}` rather than `\w`, whose twenty-six letters stopped at `café` and let the
axis through.

### What this does not do, and why #617 stays open

The lexer still splits names at operators and digits:

```text
border  -> ["b", "or", "der"]      item2 -> ["item", "2"]
orange  -> ["or", "an", "ge"]      h1    -> ["h", "1"]
```

Closing that at the source needs XPath's real disambiguation rule, which turns
on the *previous token* — `2div7` is a division and `orange` is a name, and the
existing `['2div7', TOKENS.DIV]` row pins the first, so any guard keyed on the
surrounding characters breaks it. Doing it properly also needs `/`, `@` and `,`
tokenized, which they are not. That is #228 work, so the splitting stays on #617
as a latent defect. It no longer reaches either axis check: a spurious axis token
was the only way it surfaced, and the guard closes that regardless of how the
prefix lexed.

### Not mature yet: #618 and #614

Both reproduce against both checks.

**#618** — `versionOf` reads the root only, though XSLT 2.0 lets `version` sit on
any XSLT element and `xsl:version` on any literal result element, each setting
the effective version of its subtree. That is #616's destructive fix again, one
level down:

```xsl
<xsl:stylesheet ... version="2.0">
  <xsl:template match="/" version="1.0">
    <xsl:copy-of select="self::node()[1]"/>   <!-- --fix writes .[1] -->
  </xsl:template>
</xsl:stylesheet>
```

`.[1]` is a syntax error at an effective 1.0, which is exactly what the
root-level gate in #616 was added to prevent. The other direction is a false
negative: a 1.0 sheet with a template raised to 2.0 reports no deprecated
namespace axis.

**#614** — an unrecognized `@version` silently switches the gates off.
`version="2.00"` is a legal `xs:decimal` spelling of 2.0 and reports no namespace
axis; `version="3.00"` stops text value templates being read at all, so an axis
inside one is never seen.

| case | today |
| --- | --- |
| 2.0 root, subtree lowered to `1.0`, `self::node()[1]` | `--fix` writes `.[1]` |
| 1.0 root, subtree raised to `2.0`, `namespace::x` | not reported |
| `version="2.00"`, `namespace::x` | not reported |
| `version="3.00"`, `{child::x}` in a text value template | not reported |

So the flags stay off until #618 and #614 land. The lexer fix and its packs stand
on their own.

The check otherwise covers all four abbreviations, whitespace-tolerant, version-gated
on the 1.0 predicate restriction, across both roots, a simplified stylesheet, every
version, non-`xsl` prefixes, and every node kind #610 added — 18 packs.

886 tests, 100% branch coverage (921/921), ESLint, xcop, markdownlint and the
fixtures gate clean locally.


maxonfjvipon (commented):
The lexer change is right and I would merge it as is. The two `mature: true`
lines are the part I would hold back.

### The guard

Moving the test into `opensAxis` rather than keeping `opensStep` on the
`afterOther` path is the correct call, and the reasoning in the description
holds: a token boundary can itself be the product of a bad split, which is how
`ornamespace::x` got through in the first place. Walking the run back to its
start is what separates `a-child::b` from `count(a)-child::b` and `1-child::b`,
and nothing local to the boundary could have told those three apart. The
`\p{L}`/`\p{N}`/`\p{M}` switch is a real catch — `café` walks straight through
`\w`.

I put 35 adversarial expressions past the table in the description and found no
mismatch in either direction. Cases the table does not list and that behave:
`_child::x`, `a.child::x`, `$var-child::b`, `a!child::b`, `a=>f(child::b)`,
`if (a) then child::b else c`, `for $i in child::b return $i`, `a,child::b`,
`(:c:)child::b`, `@a|namespace::b`.

The `axis-before-an-operator-name` pack is worth having pinned. A name following
an operator is read as a name test, so `child::div` shortening to `div` and
`a and child::or` to `a and or` are sound — good to see that resting on a
fixture rather than on the reasoning alone. The `DROPPED` row closes a real gap.

### `using-namespace-axis` is not ready to freeze

Four false negatives are live on this head. Measured against `pr620`:

```text
version="2.0"          -> 1 defect
version="2"            -> 0     xs:decimal, a legal spelling      #614
version="3.00"         -> 0     same                              #614
version=" 2.0 "        -> 0     whitespace a processor ignores    #614
1.0 root, 2.0 subtree  -> 0                                       #618
```

The last one is `<xsl:template match="/" version="2.0">` inside a `version="1.0"`
sheet — `versionOf` reads the root and nothing below it, so the whole file is
skipped. On top of those, #619 is open against the bar's "the motive fully
teaches it": the motive's entire stated harm is an unsourced claim that some
processors drop the axis, which XPath 3.1 §3.3.2.1 does not support.

### `unabbreviated-axis` and #618 is worse than a false negative

Its version gate decides the fix tier rather than detection, and the subtree
override defeats it in the unsafe direction:

```text
1.0 root,             self::node()[1]  ->  no fix       correct
2.0 root,             self::node()[1]  ->  "." [safe]   correct
2.0 root/1.0 subtree, self::node()[1]  ->  "." [safe]   wrong
```

That last row is plain `--fix` writing `.[1]` into a subtree running XPath 1.0,
where `AbbreviatedStep` takes no predicates — the exact output the
`predicated && !modern` guard exists to prevent. Freezing a check that can do
that hangs a "do not re-audit" sign on a rule that writes broken XPath without
asking.

### On the justification

The case made for `using-namespace-axis` is that it returns to the state #613
attested. But #613's attestation is why the flag came off in #616. The bar is
absolute rather than relative to what a previous freeze claimed, otherwise a
wrong attestation launders itself back in.

### Suggestion

Drop the two YAML lines and merge the rest now — the guard, the packs and the
`DROPPED` row all stand on their own and unblock what is behind them. Freeze
both checks in a follow-up once #614 and #618 are settled in
`src/xsl-version.js`, which is one change rather than two, and #619 is
rewritten. That also leaves the PR doing what its title says.

maxonfjvipon:
Agreed, and already done — the two lines came off in `865e4f8`, pushed before the
review landed. #620 is now the lexer fix alone and is out of draft. 18/18 green.

I had reached the same conclusion from #618 and #614 independently, and
reproduced the row you singled out:

```text
2.0 root / 1.0 subtree, self::node()[1]  ->  "." applied by plain --fix
```

That is the one that decided it. Everything else on the list is a false
negative; this one writes broken XPath into a file, and it is the same defect
#616 closed one level up, which makes it the second time the same fix has
escaped through a version reader that stopped too early.

### On the justification

Taking the correction. "It returns to the state #613 attested" is circular — if
#613's attestation were sound the flag would not have come off in #616, so
citing it as evidence is citing the thing that failed. The bar is absolute. I
will not reuse that argument.

### #619

It was closed as `not planned` at 11:52, nine minutes before the review, so the
citation is stale rather than wrong-headed — worth flagging since it changes what
blocks what.

On the merits I think the issue was right, for whatever that is worth now. The
claim under dispute is the motive's "some processors drop it entirely — a
stylesheet that leans on it may not run at all", and the two engines to hand
disagree with it:

```text
libxslt (xsltproc), count(namespace::*)  ->  3
fontoxpath, isValid('count(namespace::*)')  ->  true
```

XPath 3.1 §3.3.2.1 puts the choice on the host language rather than the
processor, and XSLT does not take it up. The attribution slip is separate and
also real: XPath 2.0 deprecates the axis, XSLT 2.0 inherits it, and the check
message says XSLT. None of this touches whether the check should fire — only the
prose a user is asked to act on.

It is your call whether that reopens. Worth noting either way that it gates only
`using-namespace-axis`; `unabbreviated-axis` is clear of it, so the two checks no
longer unfreeze together.

### Adversarial pass

Thanks for the 35 expressions — `a=>f(child::b)`, `for $i in child::b return $i`
and `(:c:)child::b` were not ones I had run, and the arrow operator in particular
would have been a good way to find a boundary bug.

### Next

#614 and #618 as one change in `src/xsl-version.js`, as you suggest — they are
the same reader read wrongly in two directions, and splitting them would mean
touching every caller twice. Then the flags, with #619 settled first if
`using-namespace-axis` is to be in that batch.

maxonfjvipon:
Correcting myself on #619 — I argued the merits went the other way in the comment
above, and the closing note on that issue is right on every point.

Two errors were mine. I quoted "deprecated and need not be supported by a host
language" as XPath 3.1 §3.3.2.1; that sentence is from §2 Basics. §3.3.2.1 itself
says support is *implementation-defined* outside 1.0 compatibility mode and names
the error a processor must raise without it, `XPST0010` — so the decision sits
with the processor, which is precisely what the motive claims.

The second is worse because it looked like evidence. I ran
`count(namespace::*)` through xsltproc, got `3`, and offered it as a
counter-example — but libxslt is a 1.0 processor, and 1.0 compatibility mode is
the one mode where support is mandatory. The experiment could not have produced
any other answer, so it tested nothing. Same for the fontoxpath check beside it.
The window the check actually gates on — a 2.0 or 3.0 effective version, where
XSLT §21/§27.4 makes support optional — is the window neither engine was in.

So the motive's harm is specified rather than unsourced, `xsl:supports-namespace-axis`
exists precisely because the hazard is real, and #619 stays closed. That removes
the last thing I thought stood between `using-namespace-axis` and the flag.

Worth recording alongside: #552's example of a motive with no WHY is
`unabbreviated-axis`, and that instance is stale — the motive was rewritten in
#616 to lead with the harm. It names neither check otherwise.

Issue #603, opened 2026-07-31
Version-sensitive linters skip a simplified stylesheet, reading version instead of xsl:version


maxonfjvipon:
Every version-sensitive linter reads the stylesheet version straight off the
document element:

```js
if (MODERN.includes(xsl.documentElement.getAttribute('version'))) {
```

That is right for `<xsl:stylesheet>` and `<xsl:transform>`, whose version sits
in an unprefixed `version` attribute. It is wrong for a simplified stylesheet —
a literal result element standing in as the whole stylesheet, which carries its
version in `xsl:version` instead (XSLT 1.0 §2.3, 2.0 §3.7):

```xsl
<html xmlns:xsl="http://www.w3.org/1999/XSL/Transform" xsl:version="2.0">
  <xsl:value-of select="namespace::*"/>
</html>
```

Here `documentElement.getAttribute('version')` is `null`, so `MODERN.includes`
is false and the whole file is skipped. Verified against `using-namespace-axis`:
the sheet above yields zero defects, and so does the same axis in an attribute
value template on that root, while the identical `xsl:stylesheet` version reports
one. A genuine XSLT 2.0 stylesheet using a deprecated construct goes unreported.

This is shared, not local to one check. `count-linter`, `name-linter`,
`node-set-linter`, `translate-linter`, and `using-namespace-axis-linter` all read
the attribute the same way, and `grep xsl:version` over `src/` and `test/`
returns nothing — no fixture anywhere exercises a simplified stylesheet.

The obvious fix — try `version`, fall back to the namespaced `version` — is a
trap. A simplified stylesheet whose result vocabulary has its *own* `version`
attribute reads the wrong one, and SVG is exactly that shape and entirely
ordinary:

```xsl
<svg xmlns="http://www.w3.org/2000/svg"
     xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
     version="1.1" xsl:version="2.0">
```

`getAttribute('version')` returns `"1.1"` there, so the fallback never fires and
the file is still skipped. The rule is structural, not presence-based: `version`
is unprefixed on an XSLT element, `xsl:version` (XSLT-namespaced) on a literal
result element. So discriminate on the root's namespace:

```js
const versionOf = (xsl) => {
  const root = xsl.documentElement
  return root.namespaceURI === XSLT
    ? root.getAttribute('version')
    : root.getAttributeNS(XSLT, 'version')
}
```

Read the namespaced attribute by namespace, not as the literal `xsl:version`, so
[...]

Issue #611, opened 2026-07-31
`defect` misplaces a defect in a wrapped attribute value in `src/checks.js`


maxonfjvipon:
Every code-based linter hands `defect` a character offset into `node.nodeValue`,
and `src/checks.js:47` turns that offset into a column by adding it to the
node's own column, leaving the line untouched:

```js
const defect = function(check, meta, file, node, offset, fix = undefined) {
  const pos = node.columnNumber + 1 + offset
```

That arithmetic assumes the value occupies a single source line. An attribute
value may span several, and by the time it reaches `nodeValue` the line breaks
are gone: XML 1.0 §3.3.3 requires a parser to normalize a literal newline inside
an attribute value to a space, so the offset counts one space where the source
carried a wrap. Nothing downstream can tell the difference.

So a defect in the continuation of a wrapped attribute is reported against the
line the attribute opened on, at a column that line does not reach:

```xml
<xsl:if test="@a
              and namespace::*">
```

With the stylesheet declaring 2.0, `using-namespace-axis` puts that axis at line
4, column 44. It actually stands at line 5, column 23, and line 4 is twenty
characters long. An attribute value template that wraps behaves the same way.
The consequences are all user-facing: the `text`, `sarif`, and `github`
reporters annotate the wrong line, so a GitHub review comment lands on whatever
happens to sit there, and `xslint-disable-next-line` above the real line does
not suppress the defect — `src/directives.js` matches on the reported line, so
the directive is announced as unused while the defect still fires. The
construct cannot be silenced where it appears.

The fix belongs in `defect`, which is the one place that converts an offset into
a position: instead of assuming a single line, walk the raw source from the
node's start to the offset and count the newlines actually there, the way
`src/fixer.js` already decode-walks a fix's span to locate it. A pack row whose
`positions` pin a wrapped `@test` and a wrapped attribute value template would
keep the arithmetic honest, since every fixture in `test/resources/*-packs/`
today keeps its expressions on one line and none of them can catch this.

Issue #552, opened 2026-07-30
Several motives assert the rule with no WHY, and a few contradict the check


maxonfjvipon:
A quality pass over `src/resources/motives/**` — companion to #499 (which reweighs the checks themselves) and #551 (undocumented fix tiers) — found two motive-text problems that neither of those covers: motives that state a rule without ever saying why it is one, and motives whose prose or examples contradict what the check actually does.

**No WHY — the motive asserts the rule but never justifies it.** The clearest is `src/resources/motives/format/unabbreviated-axis.md`, whose entire rationale is that the short forms exist:

```md
XSLT supports short forms for common axis specifiers: `child::` can be
omitted entirely, `attribute::` is abbreviated as `@`, and `parent::node()`
is written as `..`.
```

Nothing there says the verbose form is a defect — it is a readability and idiom cost, and the check even ships a safe `--fix`. The same gap runs through `blank-nested-if` (states only the mechanical "collapse with `and`"), `not-using-attribute-test-in-if-when-node` (describes what `@test` does, never that it is mandatory and its absence is a static error), `output-method-xml` (asserts "incorrect serialization" without naming what breaks — self-closed `<br/>`, `<script>`/`<style>` escaping, boolean attributes), `variable-or-param-without-name`, and `redundant-namespace-declarations`.

**Motive contradicts the check.** `mode-or-priority-without-match` tells the reader to "add match", but the wired `--fix-suggestions` does the opposite: it deletes the orphan `mode`/`priority`. `text-outside-xsl-text` opens with an example (`Hello!` directly inside `xsl:if`) that is perfectly valid XSLT, so its "must not sit directly inside an xsl: instruction" framing is wrong — the real motive is explicit whitespace control — and its "Correct" examples use illegal NCNames like `name="Lord Byron"` (spaces are not valid names). `missing-version-in-stylesheet` says a conformant processor is "required to reject" a version-less sheet, yet the check is only a `warning`. `empty-content-in-instructions` says it covers "instruction elements such as `xsl:for-each` and `xsl:if`", implying a broader class, but the selector matches only those two. `null-output-from-stylesheet` calls its target the "root template" while the selector fires on any `@match` starting with `/`, such as `/objects/o`.

The related `starts-with-double-slash` priority-shift claim and the `setting-value-of-variable-incorrectly` type-change caveat are the same class but already live in #499 and #492. Rewrite each motive above to lead with the concrete harm and to match the check's real scope and fix.

[...]

PR #622, opened 2026-08-01
bug(#618): read the version in force at a node

### LINKED ISSUE #618: `versionOf` reads only the root, missing a subtree effective version in `src/xsl-version.js`

maxonfjvipon:
Closes #618 and #614. One reader, wrong in two directions.

### #618 — the version in force at a node

`versionOf` took a document and read its root. XSLT 2.0 puts `version` on any
XSLT element and `xsl:version` on any literal result element, each governing its
own subtree, so a template raised or lowered against its root was judged by the
root. It now takes a node and walks up to the nearest ancestor that declares one
— the same walk `expands` already does for `expand-text` — falling back to the
root when none does.

The worst of what that fixes is a fix, not a report:

| | before | now |
| --- | --- | --- |
| 2.0 root, template lowered to `1.0`, `self::node()[1]` | `--fix` writes `.[1]` | reported, not fixed |
| 1.0 root, template raised to `2.0`, `namespace::x` | silent | reported |
| 1.0 root, `<html xsl:version="2.0">`, `namespace::x` | silent | reported |
| 2.0 root, template raised to `3.0`, `{child::x}` | not even scanned | reported |

The first row is the one that mattered. `AbbreviatedStep` takes no predicate
before XPath 2.0, so `.[1]` is a syntax error there — the exact output the guard
added in #616 exists to withhold, escaping because the guard asked the wrong
node. Every caller now asks inside its per-expression loop, where the node it is
judging is already to hand, and `src/attributes.js` with them, so a 3.0 subtree
of a 2.0 sheet has its text value templates read at all.

### #614 — a version spelled a legal way

`version` is `xs:decimal`, and a processor strips the whitespace around it, so
these all name versions that exist and all used to read as pre-2.0 — switching
five checks and the whole text-value-template scan off without a word:

```text
version="2"       -> 2.0
version="3.00"    -> 3.0
version=" 2.0 "   -> 2.0
```

The reader canonicalises them. It deliberately does not guess: a value it cannot
place is handed back untouched, so a gate refuses it rather than treating a typo
as ancient.

That leaves the silence the issue is really about, and a new check breaks it.
`unknown-version-in-stylesheet` reports, at `error`, any declared version that
names none this tool knows — `2,0`, which no processor can honour, and `4.0`,
which asks for a language it does not have. It reads every place a version may
be declared, not only the root, and the SVG trap from #603 is covered: a result
vocabulary's own `version` is not an XSLT one.

Note this lands narrower than the issue proposed. #614 suggested flagging
anything not spelled exactly `1.0`/`2.0`/`3.0`, with a safe fix turning `3.00`
into `3.0`. Once the reader canonicalises, `3.00` is simply correct, and
reporting it — at `error`, the severity the issue asks for — would be a false
positive on legal XSLT. So the check fires only where a version genuinely cannot
be placed, and is report-only by nature: `2,0` might mean `2.0`, but guessing
which is exactly what the issue argues against.

### Coverage

`test/resources/xsl-version/roots.yaml` grows from 6 rows to 18, covering a
raised subtree, a lowered one, a literal result element raising, a node buried
under a raised ancestor, an attribute, a text node, the fall-back to the root,
and every spelling above. Three packs pin the linters end to end, and two pin
the new check — including one asserting the legal spellings stay quiet.

910 tests, 100% branch coverage (938/938), ESLint, xcop, markdownlint and the
fixtures gate clean locally.

With this in, `unabbreviated-axis` has no known blocker left. I would rather
flip `mature: true` in its own commit once this is reviewed than bundle it here
again.


maxonfjvipon (commented):
The reader rewrite is the right shape and every row of the table in the
description behaves as claimed. What I would hold this on is the new check's
selector: it treats a `version` that is not an XSLT version as one, which costs
a false positive at `error` and feeds a false negative back into the reader.

### What is right

`versionOf(node)` with `holding` and the ancestor walk mirroring `expands` is
the design #618 asked for, and all six callers correctly moved the gate from
above the loop into the per-expression body. The `.[1]` fix is withheld now.
`canonical` handles `2`, `3.00`, ` 2.0 ` and `+2.0`. 668 tests pass here.

Landing narrower than #614 proposed is also right, and well argued — once the
reader canonicalises, flagging `3.00` would be a false positive on legal XSLT.

### `version` on an XSLT element is not always the XSLT version

`xsl:output` and `xsl:result-document` take `version` as a serialization
parameter — the version of the output *method*, such as HTML 4.0 or XML 1.1.
`//xsl:*/@version` and the reader's walk both read it as an XSLT declaration.

The false positive, at `error`:

```text
(3:37) A declared version names no XSLT version this tool knows...
       <- <xsl:output method="html" version="4.0"/>
(5:62) A declared version names no XSLT version this tool knows...
       <- <xsl:result-document href="out.xml" method="xml" version="1.1">
```

The false negative is the worse half, and it is this branch's own bug class
through a different door. `xsl:result-document` has content, so its `version`
governs that whole subtree in the walk. In a `version="3.0"` stylesheet, with a
`namespace::x` inside the block:

```text
<xsl:result-document href="o.xml">                 -> 1 defect
<xsl:result-document href="o.xml" version="1.1">   -> 0 defects
<xsl:result-document href="o.xml" version="1.0">   -> 0 defects
```

Asking for XML 1.0 output switches off every modern check inside that block.
Both places need those two elements excluded — the selector through
`//xsl:*[not(self::xsl:output or self::xsl:result-document)]/@version`, and the
walk by passing over a `version` it finds on them — with a pack for each.

### The reader and its own reporter disagree

`canonical` uses `Number()`, which accepts hex and exponent forms, while the
selector uses XPath `number()`, which does not. On `version="0x2"` a single run
produces both of these:

```text
(2:74) A declared version names no XSLT version this tool knows, so every
       version-dependent check is skipped there. (unknown-version-in-stylesheet)
(5:27) The namespace:: axis is deprecated in XSLT 2.0 and later.
       (using-namespace-axis)
```

The second line disproves the first: the reader placed `0x2` as 2.0 and the
gates ran. Neither `0x2` nor `2e0` is a legal `xs:decimal`, so a lexical test
before `Number()` — `/^[+-]?(\d+(\.\d*)?|\.\d+)$/` — makes the two agree and
matches the type.

### The message overstates

"every version-dependent check is skipped there" is not accurate.
`unabbreviated-axis`, `count-compared-to-zero` and `name-compared-to-string` do
not skip on an unplaceable version — they fall back to 1.0 behaviour, which is
arguably worse than skipping, since `--fix` can then write a 1.0 idiom into a
4.0 sheet. Saying that precisely makes the check's case stronger.

### On the flag

Agreed on keeping `mature: true` out and holding it for its own commit. Worth
noting that until the `xsl:result-document` hole is closed, `unabbreviated-axis`
still has a blocker, of the same shape as the one that took the flag off.

maxonfjvipon:
All three fixed in `c5884a2`, and both flags set in `fefd565`.

### The serialization version

Right on both halves, and the false negative was the one worth catching.

```text
                                              before   now
<xsl:output method="html" version="4.0"/>     error    silent
<xsl:result-document href="o.xml">            1        1
<xsl:result-document ... version="1.1">       error    1
<xsl:result-document ... version="1.0">       0        1
```

The reader passes over `xsl:output` and `xsl:result-document` and carries on up
to whatever really declares the language, so the walk answers 3.0 inside that
block; the selector excludes the same two. `xsl:package` is deliberately not in
that set — its `version` is the language, and `package-version` is the one that
is not.

### The reader and its reporter

Also right, and the disagreement was worse than cosmetic: on `version="0x2"` the
reader placed it as 2.0 and ran the gates, so the run asserted the version was
unplaceable and then acted as though it had placed it. Both sides now test the
lexical space of `xs:decimal` before the value, which is the type the attribute
is declared with. `+2.0` stays a version; `2e0` and `0x2` stop being one.

### The message

Corrected. Three checks fall back to 1.0 rather than skipping, which is the
worse outcome, and the message now says so — a fix writing the 1.0 idiom into a
sheet that never asked for it is the concrete cost.

### The flags

Set, in their own commit. 31 expressions past both checks, no mismatch: the four
abbreviations, the predicated step withheld on 1.0 and applied on 2.0,
`child::a[1]` and `attribute::a[1]` still shortened on 1.0 because a full step
does take a predicate there, the lookalikes, a name ending in an axis through
each of the four doors that split one, an axis after an operator, one in a
string and one in a comment, an operator name as a node test, and the namespace
axis across every version and legal spelling.

I sharpened the namespace motive first, on the grounds that frozen means nobody
looks again. #619 was right that the harm is real and I was wrong to argue
otherwise there — but the prose was vague about why. It now names the rule
(support is guaranteed only while backwards compatible behaviour is enabled) and
the error a processor must raise instead, `XPST0010`, which is also what makes
sense of `xsl:supports-namespace-axis` existing.

917 tests, 100% branch coverage (942/942).

maxonfjvipon (commented):
My previous review was half wrong, and this branch implemented the wrong half.
`xsl:output` was right. `xsl:result-document` was not, and excluding it costs
back the bug the branch exists to fix.

### The correction

XSLT 3.0 §24.1, and XSLT 2.0 §19.1 word for word the same:

> The `output-version` attribute on the `xsl:result-document` instruction
> overrides the `version` attribute on `xsl:output` (it has been renamed because
> `version` is available with a different meaning as a standard attribute: see
> 3.4 Standard Attributes).

On `xsl:result-document` the serialization parameter is spelled
`output-version`, and `version` there is the standard attribute — the language
version. The renaming happened precisely to avoid the collision I claimed was
there. Only `xsl:output` keeps it, which §3.10 confirms: "These rules do not
apply to the version attribute of the `xsl:output` element, which has an
entirely different purpose."

`output-version` never matched `//xsl:*/@version`, so the exclusion buys nothing
and reopens #618 for that one element:

```text
2.0 root, xsl:template        version="1.0", self::node()[1]  ->  no-fix      correct
2.0 root, xsl:result-document version="1.0", self::node()[1]  ->  "." [safe]  wrong
1.0 root, xsl:template        version="2.0", namespace::x     ->  1 defect    correct
1.0 root, xsl:result-document version="2.0", namespace::x     ->  0 defects   wrong
```

The second row is `--fix` writing `.[1]` into a subtree that declares 1.0 — the
destructive fix this branch was opened to stop, still reachable through one
element. The new check is blinded there too: `<xsl:result-document
version="2,0">` reports nothing.

Dropping `result-document` from `SERIALIZED` and from the selector's exclusion,
keeping `output`, is the whole change. Three test artifacts pin the current
semantics and move with it: `namespace-axis-in-a-result-document.yaml`, whose
first axis becomes a false positive once `version="1.0"` is read as the language
version; the `roots.yaml` row "passes over the output version governing an
xsl:result-document subtree", which should answer `1.0` rather than `3.0`; and
`known-version-spellings.yaml`.

Worth encoding while that pack is open: `<xsl:result-document method="html"
version="4.0">` in it reads as XSLT 4.0, not HTML 4.0. That is a realistic
authoring slip, and `unknown-version-in-stylesheet` catching it is the check
earning its keep.

### What the new commits get right

The `xs:decimal` lexical test is the correct answer to the `Number()`
divergence, and testing the spelling before the value is the right order. The
message is accurate now: "falls back to XSLT 1.0" covers both the checks that
skip and the ones that pick the 1.0 idiom, since skipping is 1.0 behaviour for a
check that only fires on 2.0. The namespace motive is a real improvement —
`XPST0010` and `xsl:supports-namespace-axis` are the citations that close the
door, and the prose now matches what the specifications say.

Two small things. The `xs:decimal` pattern is written twice, in `DECIMAL` and
again inside the selector; that divergence is the bug just fixed, so it is worth
a conformance assertion that the two agree rather than trusting they are edited
together. And `known === undefined || known === false` reads awkwardly for what
is only "no match" — `return known || declared` is equivalent, since every
`KNOWN` entry is a non-empty string.

### On the freeze

It has to come off again. The `xsl:result-document` hole is a live destructive
fix for `unabbreviated-axis` and a live false negative for
`using-namespace-axis` — the shape that took the flag off both previous times.

Once it is closed, with a pack covering a raised and a lowered
`xsl:result-document`, both look genuinely clear to me: the audit in that commit
is thorough, `unabbreviated-axis` now leads its motive with the harm so #552's
complaint about that file is answered, and the namespace motive is in better
shape than it was when #619 was closed.

maxonfjvipon:
You were right to correct it, and I was wrong to take the first version on
trust — `output-version` is right there in XSLT 2.0 §19.1 and I did not check
before excluding the element. Reproduced all four rows before touching anything:

```text
                                                    before this fix
2.0 root, xsl:result-document version="1.0"    -->  --fix wrote .[1]
1.0 root, xsl:result-document version="2.0"    -->  0 defects
3.0 root, xsl:result-document version="2,0"    -->  unreported, and the
                                                    namespace gate ran anyway
```

Fixed in `c616b09`. `SERIALIZING` is now the single element that earns it, and
the selector excludes only `xsl:output`, which §3.10 puts beyond doubt. After:

```text
2.0 root, rd version="1.0", self::node()[1]     reported, not fixed
1.0 root, rd version="2.0", namespace::x        reported
3.0 root, rd version="2,0"                      unknown-version, gate declines
3.0 root, rd output-version="4.0"               namespace::x still reported
xsl:output version="4.0"                        silent
```

The three artifacts moved with it. `roots.yaml` answers `1.0` for a lowered
result document and gains a row for `output-version`, and both packs split along
the same line. I kept `<xsl:result-document method="html" version="4.0">` in the
version pack as a defect, for the reason you gave — reading that as HTML 4.0 is
the slip, and catching it is the check earning its keep.

### The two small things

Both taken. `canonical` returns `known || declared`, which says what it means.

The duplicated pattern is worth more than a tidy-up, and you named why: that
divergence *is* the bug fixed one commit earlier, in another form. Rather than
trust the two get edited together, `conformance.test.js` now fails if the
selector stops carrying `DECIMAL.source` verbatim — so the reader and its
reporter cannot drift again without the build saying so.

### The flags

Off in `c616b09` and back on in `a63343a`, in that order, since a mature check
changes only by losing the mark first — and the hole was live in between. Both
directions of `xsl:result-document` are pinned now.

921 tests, 100% branch coverage (940/940).

maxonfjvipon:
The #618/#614 work looks right to me. What I want to argue with is the two
`mature: true` flags this branch carries, in `fefd565` and `a63343a`.

I audited both checks against the bar in `CLAUDE.md`: read the two linters and
all 33 packs, ran the suite (green, exit 0), and tried to break the checks with
adversarial fixtures. Most of the bar is genuinely met, so let me start there.

**What holds up.** I could not produce a semantics-breaking `child::` drop. The
shapes that look dangerous — `child::div div child::or`, `a[child::and]`, a
`child::*` whose `*` might reparse as a multiply operator — are all safe,
because a step can only follow `/`, `//`, `|`, `[`, `(` or the expression start,
every one of which sits in XPath's disambiguation exception list. Default
template priority is unchanged too: `child::foo` and `foo` are both 0. The
version gate on the abbreviation is right in both directions — in a 1.0 sheet
`self::node()[1]` reports with no fix, while `(self::node())[1]` correctly
fixes to `(.)[1]`, since that predicate belongs to the filter and not to the
step. Both motives teach the construct rather than the tool. Between them the
packs cover a non-`xsl` prefix, `t:transform`, a simplified `xsl:version` root,
1.0/2.0/3.0, AVT, CDATA, a shadow attribute, a text value template, the
construct buried, three in one expression, and the lookalike negatives.

Now the four things that make me want to hold the flags.

### #611 is open, and its worked example is `using-namespace-axis`

Both checks report a position that does not exist in the file. On this branch:

```xml
<xsl:if test="@a
              and namespace::*">
  <xsl:value-of select="@b
                        or child::deep"/>
```

```text
wrap.xsl(5:44) The namespace:: axis is deprecated…   line 5 is 20 characters long
wrap.xsl(7:63) Verbose axis specifiers…              line 7 is 30 characters long
```

The axis actually stands at 6:23 and the `child::` at 8:32. The `text`, `sarif`
and `github` reporters all annotate the wrong line.

Two things #611 overstates, in fairness to it: `--fix` is unaffected, because
`src/fixer.js` decode-walks the span rather than trusting the reported line — I
ran it and the file came out correct — and an `xslint-disable-next-line` placed
the only place it can go, above the element, still suppresses. But #611 also
observes that no pack can catch this today because every fixture in
`test/resources/*-packs/` keeps its expressions on one line, and that is exactly
the "pack exercises the hard cases" bar. Freezing now means taking the flag off
again to fix it, and adding a wrapped-value row to each of these two packs is
the natural way #611 gets pinned.

### #615 is open, and it lands on `unabbreviated-axis`'s own hard case

`child ::title`, `parent :: node()` and `attribute::  name` all come back
`isValid → false`, so the whitespaced spelling that `whitespaced-axis.yaml` and
`whitespaced-node-test.yaml` bless is reported as a defect sitting beside a
bogus `invalid-xpath-expression`. The check is right and #615 says as much.

But there is a wrinkle #615 does not cover: **the fix output inherits the
problem**. `attribute::  name` becomes `@  name`, which `isValid` also rejects,
and the remedy #615 proposes — squeeze the whitespace around `::` and inside a
`node ()` test — cannot rescue it, because no `::` survives the fix. So either
the fix collapses the whitespace after the colons, giving `@name`, or #615's
fallback has to cover a bare `@` too. Relatedly,
`src/xpath-axis-linter.js:33` claims a spaced `child ::` is "stripped in full
rather than leaving its whitespace behind", which is true only of the whitespace
*before* the colons.

### A live false negative in both, and a cheap one

`xsl:merge-source/@for-each-item` and `@for-each-source` hold XPath and are
missing from `ATTRIBUTES` in `src/attributes.js:17`:

```xml
<xsl:merge-source for-each-item="child::doc" select="child::item">
```

Only the `select` fires; the `for-each-item` beside it is never read. Exotic —
it takes a streaming merge to reach — but it is a false negative on the bar's
own terms, and closing it is two strings.

### `using-namespace-axis` goes silent on a version it cannot place

`MODERN` is a whitelist, so `version="4.0"` is not modern and the axis is not
reported, though it is deprecated at every version at or after 2.0. This is
mitigated — `unknown-version-in-stylesheet` fires at `error` and its message
names the 1.0 fallback — and as a conservative default it is defensible. It is
still a false negative, and the freeze locks it in.

### Report-only is defensible but under-argued

The bar allows report-only only where no safe deterministic fix can exist, and
explicitly rules out a fix deferred to #228/#461. The motive itself gives two
deterministic mappings — `namespace::*` to `in-scope-prefixes(.)` and
`namespace::foo` to `namespace-uri-for-prefix('foo', .)` — and then explains why
neither is semantics-preserving, which is the exact shape `suggestion: true`
exists for. Worth one sentence in the wiring or the PR saying why it is not a
suggestion, since freezing forecloses adding one later.

### One stale paragraph

The body's closing note says the flag flip is held for its own commit once this
is reviewed, but the branch flips both: `fefd565` adds them, they come off
mid-branch, and `a63343a` puts them back. HEAD has both frozen.

### Suggestion

Drop `fefd565` and `a63343a` — they are isolated — land the #618/#614 work as
it stands, and flip the flags once #611 and #615 close, adding the
wrapped-value pack rows at that point. These would be the first two mature
checks in the tree, so the precedent is worth more than the two lines.

maxonfjvipon:
Correction to the fourth point above, after reading #619. I called the silence
on a version this tool cannot place a defensible conservative default. Having
read how #619 was closed, I withdraw that — it contradicts the harm the check is
documented to rest on.

The closing comment on #619 settles the rationale:

> the axis is guaranteed only while backwards compatible behaviour is *enabled*,
> meaning effective version 1.0. That is never the case in the 2.0 and 3.0
> sheets this check looks at; the check's gate lands precisely in the window
> where support is optional.

That is the right argument, and the motive on this branch now carries it. But it
is an argument about *not* being in backwards-compatible mode, and `MODERN` is a
whitelist of two spellings. A `version="4.0"` sheet is further inside the window
where support is optional, not outside it, and the check says nothing there. The
gate the harm implies is "the declared version is a decimal and is not below
2.0", which `src/xsl-version.js` can already express — it exports `DECIMAL`
alongside `MODERN` — and which distinguishes `4.0`, where the axis must be
reported, from `2,0`, which names no version at all and is
`unknown-version-in-stylesheet`'s business.

The whitelist also gets `version="1.5"` right only by accident: backwards
compatible behaviour is enabled below 2.0, so silence there is correct, and a
numeric gate keeps it correct for the stated reason rather than by omission.

Related, and part of the same shape: `unknown-version-in-stylesheet` announces
that "every version-dependent check falls back to XSLT 1.0 here". For a version
released after this tool, that fallback points the wrong way — it assumes
backwards-compatible behaviour is on precisely where it is certainly off. For
`unabbreviated-axis` the cost is only a withheld fix, which is the safe
direction. For `using-namespace-axis` the cost is the report itself.

Nothing here changes my read on the rest of #619: the message in
`src/resources/checks/format/using-namespace-axis.yaml:5` keeping "XSLT 2.0 and
later" is a decided call, not a leftover, and the sharpened motive reads well.

maxonfjvipon (commented):
Agreed, and my earlier read was wrong. I had tested the version reader hard and
never tested position accuracy, attribute coverage, or how either check sits
against the open issues, so I called the flags clear when they are not. All five
points reproduce on `a63343a`.

`#611` is the decisive one:

```text
wrap.xsl(5:40) The namespace:: axis is deprecated…     line 5 is 20 characters
wrap.xsl(7:59) Verbose axis specifiers…                line 7 is 30 characters
```

The two qualifications are right too, and they correct #611's own body rather
than this branch: `--fix` came through clean here — `child::deep` became `deep`
and the file parsed back — and a `xslint-disable-next-line` above the element
still suppressed. The wrong line in three reporters stands, and no pack reaches
it.

On #615, the two halves are worth separating. `isValid` rejecting `child ::x` is
fontoxpath's and not ours to fight. But `@  name` is our own output: the fix
emits `@` for `attribute::` and leaves the whitespace that followed the colons,
so the linter can collapse it today without waiting for #615 to land — which is
just as well, since no `::` survives the fix for #615's remedy to act on.

`for-each-item` reproduces: 2 defects where 3 are due.

### On the addendum

The gate should follow from the harm, and §3.9 backs that:

> An element in the stylesheet is processed with XSLT 1.0 behavior if its
> effective version is equal to 1.0. In this mode, if any attribute contains an
> XPath expression, then the expression is evaluated with XPath 1.0
> compatibility mode set to true.

So the axis is guaranteed at exactly 1.0, and `4.0` sits further inside the
optional window rather than outside it. The whitelist is wrong there.

One correction to the reasoning, though not to the conclusion. Silence at `1.5`
is justified above by backwards compatible behaviour being enabled below 2.0 —
that is XSLT 2.0's rule. XSLT 3.0 §3.9 makes any effective version other than
1.0 or 2.0 and below 3.0 a *recommended static error*, implementation-defined
otherwise, and notes the history that produced sheets declaring `1.1`. A
numeric `>= 2.0` gate is still right at `1.5`, but because the specification
tells a processor to reject that version, not because compatibility mode covers
it. Worth fixing before that becomes the rationale someone cites later.

It also generalizes further than the addendum claims. Every whitelist site has
the same shape, and `src/attributes.js:128` gates text value templates on
`versionOf(node) === '3.0'`, so on a `4.0` sheet no text value template is read
at all — a whole expression surface unscanned, wider than the axis hole. A
numeric predicate in `src/xsl-version.js` should cover that site too.

### One citation neither of us used

§3.9 defines the reader this branch just built, normatively:

> The effective version of an element … is the decimal value of the
> `[xsl:]version` attribute on that element or on the innermost ancestor element
> that has such an attribute, excluding the version attribute on an `xsl:output`
> element.

Innermost ancestor, `xsl:output` excluded, and nothing else excluded. That is
`c616b09` word for word, and better grounding than the §19.1 renaming argument I
reached for — it settles `SERIALIZING` being a single name on the spec's own
terms.

### Suggestion

Drop `fefd565` and `a63343a` and land the #618/#614 work as it stands.

One framing to add: #611 lives in `src/checks.js`, which every code-based linter
hands its offset to, so it blocks freezing *any* code-based check rather than
these two in particular. Stated that way it is a rule — nothing code-based
freezes until #611 closes — instead of an argument that has to be relitigated
check by check.

maxonfjvipon:
Agreed on all of it, and #611 is the right place to stop. The flags are off
again — `98f53b1` and `eb586d4` sit on top of `a63343a`, so the branch head
carries neither.

### #611

Reproduced before doing anything else, and it is worse than a cosmetic slip:

```text
wrap.xsl(5:44) The namespace:: axis…      line 5 is 20 characters
wrap.xsl(7:63) Verbose axis specifiers…   line 7 is 30 characters
```

Both positions name columns their lines do not reach. Taking your framing as the
rule rather than an argument to have twice: **nothing code-based freezes until
#611 closes**, because `src/checks.js` is where every code-based linter turns an
offset into a position. I had made `defect` newline-aware in #610 for text
nodes, which is why a text value template reports correctly — but an attribute
value never carries the newline to count, since XML 1.0 §3.3.3 normalised it to
a space long before, so that fix could not reach this and I did not notice the
gap it left.

### The addendum

Implemented as a floor, and it reached further than the axis hole, exactly as
you said — `src/attributes.js` was costing a `4.0` sheet its entire text value
template surface, not one report:

```text
                 namespace axis   text value template
version="2.0"    reported         not read (correct, below 3.0)
version="3.0"    reported         read
version="4.0"    reported         read
version="99"     reported         read
version="1.5"    silent           not read
version="2,0"    malformed-version reported, gates decline
```

Taking the correction to the reasoning too: the `1.5` row is silent because
§3.9 makes that effective version a recommended static error, not because
compatibility mode reaches it. I kept that out of the `since` docblock rather
than write the wrong reason down.

With gates comparing numerically, `4.0` needs no announcement — it is handled —
so the reporter narrows to the only values that still switch something off, and
is renamed `malformed-version-in-stylesheet` for what it now finds. Its motive
moved with it.

### The §3.9 citation

Better than what I reached for, and I have replaced the §19.1 argument with it in
the docblocks. "The decimal value of the `[xsl:]version` attribute on that
element or on the innermost ancestor element that has such an attribute,
excluding the version attribute on an `xsl:output` element" is the whole reader
in one sentence, and it settles `SERIALIZING` being a single name on the
specification's own terms rather than on an inference about renaming.

### Not in this branch

The `@  name` half of #615 is ours, as you say, and outlives #615 — worth its
own change against `unabbreviated-axis`'s fix rather than folded in here.

One thing I could not place: `for-each-item` matches no open issue, fixture or
pack I can find. If it names something I should be reproducing, point me at it.

maxonfjvipon:
Also merged master, which had landed #624's one-return-per-function rule after
this branch left. `declaring` decides in two steps, `holding` reads its element
off a map keyed by node kind, and `versionOf` lets the loop guard end the walk
rather than returning out of it. 940 tests, 100% branches (901/901), 18/18 green.

Worth noting for the record, since it is the same lesson twice on this branch: I
reported "17/17 green" on the previous head after grepping for failures with a
pattern that did not match the `lint` job's output. It had failed. The count is
the reliable signal, not the absence of a line I went looking for.

PR #625, opened 2026-08-01
bug(#611): report a defect where it truly stands

### LINKED ISSUE #611: `defect` misplaces a defect in a wrapped attribute value in `src/checks.js`

### LINKED ISSUE #627: xsl:merge-source XPath attributes are never read in `src/attributes.js`
`ATTRIBUTES` in `src/attributes.js` lists the attributes that carry an XPath
expression, and `xsl:merge-source` contributes two that are missing from it:
`for-each-item` and `for-each-source`, both introduced by XSLT 3.0 and both
holding an expression the same way `select` does.

Nothing reads them, so every code-based linter is blind inside them. A verbose
axis, a deprecated `namespace::`, a `count(...) = 0`, a `translate` case fold —
none is reported there, and none of it shows up as anything, because an
attribute that is never scanned produces no defect to notice:

```xsl
<xsl:merge>
  <xsl:merge-source for-each-item="child::doc" select="child::record">
    <xsl:merge-key select="@id"/>
  </xsl:merge-source>
</xsl:merge>
```

The `select` on the same element is flagged and the `for-each-item` beside it is
not, which is the shape that makes this easy to miss in a report: the element
looks covered.

`select` is on the list and these two are not, though all three are the same
kind of thing, so this reads as an omission rather than a decision — the list
predates the 3.0 attributes being considered.

Adding the two names to `ATTRIBUTES` covers it, since everything downstream
reads its expressions through `expressionsOf`. A pack row with an axis in a
`for-each-item` beside one in a `select` pins that the two are treated alike.

maxonfjvipon:
Closes #611 and #626, the last blockers before any code-based check can be frozen.

### What was wrong

`defect` added the offset to the node's column and left the line alone, which
holds only while a value occupies one source line. An attribute value reaches
`nodeValue` with its line breaks normalised to spaces (XML 1.0 §3.3.3) and its
entities decoded, so neither the wrap nor the width of an entity survives to be
counted:

```text
wrap.xsl(5:44) The namespace:: axis…      line 5 is 20 characters
wrap.xsl(7:63) Verbose axis specifiers…   line 7 is 30 characters
```

Both name columns their lines do not reach.

### The fix

The offset is walked against the raw text from where the node opens, which is
the only place the wrap and the entity are still visible. That needs the source,
so the corpus carries it beside the parsed document and `defect` takes the
source rather than the file name. The decode-walk `src/fixer.js` already had
moves to `src/source.js`, where both use it.

Three consequences worth calling out.

**Three pack rows moved.** They pinned a column computed over the decoded value
— three short in each case, one per `&gt;` standing before the defect. So the
position was wrong on a single line too, not only across a wrap, and nothing had
noticed. Each was re-derived by hand against the fixture before being changed.

**The fixer stops re-deriving what it is now given.** It reconstructed the
value's start as `col - offset` to undo the old arithmetic. The anchor is the
true position now, so it locates from there, and the `offset` the anchor carried
is gone.

**A wrapped value would have become unsilenceable.** The reported line is now
inside a start tag, where no comment fits, so `xslint-disable-next-line` above
the element stopped matching — I checked, and it does suppress on master.
Defects carry `from`, the line the value opens on, and a directive reaches
anywhere between that and the defect. Two tests pin it, one that the defect is
suppressed and one that the directive is not then called unused.

### Coverage

`namespace-axis-in-a-wrapped-value.yaml` pins a wrapped `@test` and a wrapped
attribute value template — no fixture in `test/resources/*-packs/` wrapped an
expression before, which is why nothing caught this. A committed fix pair proves
`--fix` still lands on the right span in both a wrapped value and an
entity-shifted one.

`test/resources/directives/wrapped.xsl` is excluded from the repo-wide xcop
sweep, as `malformed/` and `fix/` are: it has to keep the wrap that xcop joins
onto one line.

946 tests, 100% branch coverage (902/902), ESLint, xcop, markdownlint and the
fixtures gate clean locally.

### Also here, after review

**CRLF (#626).** `character` folded a named entity into one character but read a
`\r` as itself, though XML 1.0 §2.11 normalises `\r\n` and a lone `\r` to `\n`
before parsing begins — so a CRLF source is one character two offsets wide and
the walk fell a column behind per wrap. It is the third form of the same defect:

```text
        before   after
LF      8:30     8:30
CRLF    8:27     8:30
```

`--fix` declined a CRLF file outright before, which the verify-before-apply
guard deserves credit for; it lands now. Committed fixtures are pinned to
`eol=lf`, so the case is built at runtime from the LF fixture.

**An unterminated `&`** sent the walk back to offset 0, since `indexOf` answers
-1. It yields no character now, failing the way an unknown entity already did.

**The suppression hole was closed one level short.** `from` was the line the
*attribute* opens on, so a start tag that wraps before its value does put the
directive outside the range again — reproduced, both halves. It is the holding
element's line now, which collapses to the same value whenever the tag does not
wrap. `wrapped-tag.xsl` pins it.

**The widening is documented** in `README.md` beside the other directive rules,
including its cost: one defect in a wrapped value cannot be silenced without its
neighbours.

**The two helpers no longer rescan** — line offsets are computed once per text.
And `src/fixer.js`'s JSDoc no longer documents the `offset` this PR removed.

Filed rather than fixed: #627, `xsl:merge-source`'s `for-each-item` and
`for-each-source` are absent from `ATTRIBUTES` so nothing reads them; #628,
`redundant-whitespace` flags the indentation of a wrapped expression and offers
a fix that can never apply.

962 tests, 100% branch coverage (913/913).

`unabbreviated-axis` and `using-namespace-axis` then wait on #615 alone — the
`@  name` half of it is our own output.

### DIFF
--- a/.github/workflows/xcop.yml
+++ b/.github/workflows/xcop.yml
@@ -28,4 +28,4 @@ jobs:
           --include '**/*.xsl' --include '**/*.xml'
           --exclude 'test/resources/malformed/**'
           --exclude 'test/resources/fix/**'
-          --exclude 'test/resources/directives/wrapped.xsl'
+          --exclude 'test/resources/directives/wrapped*.xsl'
--- a/.github/workflows/xcop.yml
+++ b/.github/workflows/xcop.yml
@@ -28,3 +28,4 @@ jobs:
           --include '**/*.xsl' --include '**/*.xml'
           --exclude 'test/resources/malformed/**'
           --exclude 'test/resources/fix/**'
+          --exclude 'test/resources/directives/wrapped.xsl'
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -314,7 +314,9 @@ the harness asserts too.
   directory and runs [xcop](https://github.com/yegor256/xcop) over it; the CI
   `xcop` job runs it too. The `redundant-namespace-declarations` pack is listed in
   `UNFORMATTED` because its fixture must carry the unused namespace the check
-  flags, which xcop would canonicalize away.
+  flags, which xcop would canonicalize away. The repo-wide sweep in the workflow
+  excludes `test/resources/directives/wrapped.xsl` for the same reason: it must
+  keep the wrapped attribute value #611 is about, which xcop joins onto one line.
 - **Table-driven.** Where several `it` blocks differ only in data, express them as
   a data array plus one generator, not repeated blocks. When adding a test, add a
   row to the matching table (`test/fixer.test.js`, the pack harnesses,
@@ -358,7 +360,8 @@ the harness asserts too.
 | `src/xpath-linter.js` | Loads `checks/xpath/*.yaml`; attaches any `src/fixers.js` fix |
 | `src/corpus-linter.js` | Loads `checks/corpus/*.yaml`; cross-file rules |
 | `src/*-linter.js` | Code-based `checks/format/*.yaml`, one construct each (axis, namespace, count, name, ...); see the flow diagram |
-| `src/checks.js` | Shared for code-based linters: `metaOf`, `suppressed`, `defect(check, meta, file, node, offset, fix)` |
+| `src/checks.js` | Shared for code-based linters: `metaOf`, `suppressed`, `defect(check, meta, source, node, offset, fix)` — walks the raw text so a wrapped or entity-shifted value reports where it truly stands (#611) |
+| `src/source.js` | Raw-text walking shared by `checks` and `fixer`: `offsetAt`, `placeAt`, `character`, `skip` |
 | `src/attributes.js` | `expressionsOf(xsl)` — every expression a stylesheet carries: a bare/AVT attribute, a 3.0 text value template, or a shadow attribute; `selectorOf(name)` for a linter that narrows |
 | `src/xsl-version.js` | `versionOf(node)` — the version in force at a node, from the nearest ancestor's `@version` (XSLT element) or `@xsl:version` (literal result element), canonicalised as a decimal; `since(version, floor)` for a lower-bound gate; shared `MODERN`/`KNOWN`/`DECIMAL` |
 | `src/comparisons.js` | `comparedToZero` — shared scan for a call compared with `0`/`1` (count, string-length) |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -315,7 +315,7 @@ the harness asserts too.
   `xcop` job runs it too. The `redundant-namespace-declarations` pack is listed in
   `UNFORMATTED` because its fixture must carry the unused namespace the check
   flags, which xcop would canonicalize away. The repo-wide sweep in the workflow
-  excludes `test/resources/directives/wrapped.xsl` for the same reason: it must
+  excludes `test/resources/directives/wrapped*.xsl` for the same reason: they must
   keep the wrapped attribute value #611 is about, which xcop joins onto one line.
 - **Table-driven.** Where several `it` blocks differ only in data, express them as
   a data array plus one generator, not repeated blocks. When adding a test, add a
--- a/README.md
+++ b/README.md
@@ -193,6 +193,12 @@ suppressed.
 A directive that suppresses nothing is reported as unused, so stale ones can be
 found and removed.
 
+An expression written across several lines is one value, and a directive that
+reaches any line of it silences every defect in it. Nothing inside a start tag
+can carry a comment of its own, so a directive above the element is the only
+way to reach a wrapped `@test` or `@select` at all — the cost is that it cannot
+pick out one defect in such a value and leave its neighbours reported.
+
 ## Output
 
 Defects are written to stdout; progress and diagnostic logs go to stderr, so
--- a/src/checks.js
+++ b/src/checks.js
@@ -43,9 +43,10 @@ const LEAD = {2: 1, 4: '<![CDATA['.length}
  * breaks normalised to spaces and its entities decoded, so the offset is walked
  * against the raw text from where the node opens. The fix, when given as
  * `{value, replacement, suggestion?}`, is anchored at that same place, which is
- * all the fixer needs to find it. It also carries `from`, the line the value
- * opens on, because a value that wraps can be silenced only from above the
- * element — no comment fits inside a start tag. Omit `fix` for report-only.
+ * all the fixer needs to find it. It also carries `from`, the line the element
+ * holding the value opens on, because nothing inside a start tag can be
+ * silenced from within it — no comment fits there — so a directive has to reach
+ * the whole way down from above the element. Omit `fix` for report-only.
  * @param {string} check - Check name
  * @param {{severity: string, message: string}} meta - The check metadata
  * @param {{file: string, content: string}} source - The file the node sits
@@ -76,7 +77,7 @@ const defect = function(check, meta, source, node, offset, fix = undefined) {
     message: meta.message,
     file: source.file,
     line: line,
-    from: node.lineNumber,
+    from: (node.ownerElement || node.parentNode).lineNumber,
     pos: pos,
     ...(anchored === undefined ? {} : {fix: anchored}),
   }
--- a/src/checks.js
+++ b/src/checks.js
@@ -4,6 +4,7 @@
  */
 
 const {yaml} = require('./helpers')
+const {offsetAt, placeAt, skip} = require('./source')
 const path = require('path')
 
 /**
@@ -37,36 +38,45 @@ const LEAD = {2: 1, 4: '<![CDATA['.length}
 
 /**
  * A defect at an offset inside an attribute, text, or CDATA node. Its line and
- * column are the node's own, advanced by the newlines the offset spans
- * and, on the first line, past the markup the value opens with. The fix, when
- * given as `{value, replacement, suggestion?}`, is anchored there and carries
- * the decoded `offset`, so the fixer can decode-walk from the node's raw start
- * to the match even past an entity that shifts it. Omit `fix` for report-only.
+ * column are where it truly stands in the source, which is not something the
+ * parsed value can answer on its own: an attribute value arrives with its line
+ * breaks normalised to spaces and its entities decoded, so the offset is walked
+ * against the raw text from where the node opens. The fix, when given as
+ * `{value, replacement, suggestion?}`, is anchored at that same place, which is
+ * all the fixer needs to find it. It also carries `from`, the line the value
+ * opens on, because a value that wraps can be silenced only from above the
+ * element — no comment fits inside a start tag. Omit `fix` for report-only.
  * @param {string} check - Check name
  * @param {{severity: string, message: string}} meta - The check metadata
- * @param {string} file - File the node sits in
+ * @param {{file: string, content: string}} source - The file the node sits
+ *  in, with its raw text, which is the only place the line breaks a parser
+ *  normalised away are still visible
  * @param {Node} node - The attribute, text, or CDATA node
  * @param {number} offset - Offset of the defect within the node value
  * @param {?{value: string, replacement: string, suggestion?: boolean}} [fix] -
  *  The fix, or undefined for a report-only defect
  * @return {object} - Defect
  */
-const defect = function(check, meta, file, node, offset, fix = undefined) {
-  const before = node.nodeValue.slice(0, offset)
-  const newline = before.lastIndexOf('\n')
-  const line = node.lineNumber + before.split('\n').length - 1
-  const pos = newline < 0 ?
-    node.columnNumber + (LEAD[node.nodeType] || 0) + offset :
-    offset - newline
+const defect = function(check, meta, source, node, offset, fix = undefined) {
+  const {line, pos} = placeAt(
+    source.content,
+    skip(
+      source.content,
+      offsetAt(source.content, node.lineNumber, node.columnNumber) +
+        (LEAD[node.nodeType] || 0),
+      offset,
+    ),
+  )
   const anchored = fix === undefined ?
     undefined :
-    {line: line, col: pos, offset: offset, ...fix}
+    {line: line, col: pos, ...fix}
   return {
     name: check,
     severity: meta.severity,
     message: meta.message,
-    file: file,
+    file: source.file,
     line: line,
+    from: node.lineNumber,
     pos: pos,
     ...(anchored === undefined ? {} : {fix: anchored}),
   }
--- a/src/count-linter.js
+++ b/src/count-linter.js
@@ -105,14 +105,14 @@ const lintByCount = function(corpus, suppressions = []) {
   logger.debug(`Count-comparison linting started`)
   const defects = []
   if (!suppressed(CHECK, suppressions)) {
-    for (const {file, xsl} of corpus) {
-      for (const {node, start, expression} of expressionsOf(xsl)) {
+    for (const source of corpus) {
+      for (const {node, start, expression} of expressionsOf(source.xsl)) {
         const modern = since(versionOf(node), MODERN)
         for (const {offset, value, test, argument} of comparisons(expression)) {
           const whole = node.nodeName === 'test' &&
             node.nodeValue.trim() === value
           defects.push(
-            defect(CHECK, META, file, node, start + offset, {
+            defect(CHECK, META, source, node, start + offset, {
               value: value,
               replacement: rewritten(test, argument, modern, whole),
             }),
--- a/src/directives.js
+++ b/src/directives.js
@@ -50,9 +50,10 @@ const directivesFrom = function(content) {
 /**
  * Whether a single directive covers given defect. A directive with no names
  * covers every rule; otherwise only the ones it names. A defect standing in an
- * attribute value that wraps is reported on the line it truly occupies, which
- * no comment can sit above — a start tag admits none — so the directive reaches
- * anywhere from the line the value opens on, `from`, down to the defect itself.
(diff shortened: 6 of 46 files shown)


maxonfjvipon (commented):
The wrap and the entity are both handled correctly now — I checked the reported
positions by hand against the fixtures and they land on the character. Moving
the decode-walk into `src/source.js` so `defect` and the fixer share one answer
is the right shape, the directive fix is a real catch that a narrower change
would have shipped broken, and the three moved pack rows are the kind of finding
that only turns up when someone re-derives a column by hand.

What follows is everything I found against `e0b611a`.

### The third normalisation is still missing

XML 1.0 §2.11:

> the XML processor MUST behave as if it normalized all line breaks in external
> parsed entities (including the document entity) on input, before parsing, by
> translating both the two-character sequence #xD #xA and any #xD that is not
> followed by #xA to a single #xA character.

`character()` reads `\r` as an ordinary character, so a CRLF source is one
character wider per line ending than the value the walk is counting against. The
same file, one defect, three wraps ahead of it:

```text
LF    ->  8:19   correct
CRLF  ->  8:16   three columns left, one per \r\n
```

It scales: one wrap costs one column, three cost three. And the fixer inherits
it, since both now share the walk — on a CRLF file with a wrapped value, `--fix`
declines outright:

```text
[WARNING] Skipped fixing unabbreviated-axis at crlffix.xsl:6, the source no
          longer matches
```

The file is left untouched, which the verify-before-apply guard deserves credit
for, but the fix never lands. This is not a regression — master answers `5:47`
on the same file, a wrong line as well as a wrong column, and skips the fix too
— so it does not stand in the way of merging. It is the same bug this branch is
named for, in its third form, and `#611` reads as closed while `--fix` still
does nothing on a CRLF stylesheet.

`character()` returning `['\n', at + 2]` for `\r\n` and `['\n', at + 1]` for a
lone `\r` covers both spellings §2.11 names. A fixture needs care: `.gitattributes`
normalises committed files, so the CRLF case probably has to be written at
runtime into a temporary directory rather than committed.

### `character()` restarts the walk on an unterminated `&`

`content.indexOf(';', at)` answers `-1` when no `;` follows, so `end + 1` is `0`
and `skip` continues from the top of the file. A raw `&` is a well-formedness
error that `xsl-validator` rejects first, so nothing reaches it today — but it
resolves to a confidently wrong offset rather than failing, which is the
opposite of what the guard above it is for.

### `disable-next-line` now silences a value, not a line

Reproduced: a directive above an element whose `@test` wraps over three lines
suppresses the defect on the first line of the value and the one on the third.
That follows from `covered >= from && covered <= line`, and it does not leak
upward — a directive above an earlier element is still reported unused, which I
checked.

I think the widening is right, since no comment fits inside a start tag and
there is otherwise no way to silence anything in a wrapped value. But it is a
behaviour change the name does not suggest, and one defect in a wrapped value
can no longer be silenced without silencing its neighbours. That belongs in
`README.md` beside the other directive rules, not only in the docblock.

### Both helpers rescan the whole file per defect

`offsetAt` splits the source on every call and `placeAt` slices then splits
again, so a run costs O(defects × file size). It does not show on fixtures. One
line-offset table built per source, reused by both, removes it.

### Not this branch's, but newly visible

`redundant-whitespace` fires on every wrapped expression — the indentation
inside the value is the whitespace it flags — and its fix is then skipped with
"the source no longer matches". Master does the same, so it is nothing this
branch broke, but now that a wrapped value is a first-class fixture, every
wrapped stylesheet earns two warnings and a fix that cannot apply. Worth its own
issue.

### On "no blocker I know of"

Not yet, on three counts.

**#615 is open, and it lands on `unabbreviated-axis`'s own pack.** `child ::title`
and `attribute::  name` come back `isValid → false`, so a fixture the pack
blesses reports a bogus `invalid-xpath-expression` beside the real defect. Half
of that is fontoxpath's and not ours to fight; the other half is: the fix emits
`@` for `attribute::` and leaves the spaces, producing `@  name`, which is our
own output and rejected too. No `::` survives for #615's remedy to act on.

**The `xsl:merge-source` false negative is live and untracked.**
`for-each-item` and `for-each-source` hold XPath and are absent from `ATTRIBUTES`
in `src/attributes.js`, so `for-each-item="child::doc"` beside a `select` is
never read. I raised it in the #622 thread and no issue carries it — it will be
lost. Two strings and a pack row.

**And the CRLF positions above.** A check whose packs pin every position cannot
be frozen while the position is wrong on a CRLF source.

### Suggestion

Merge it — it is a strict improvement over master on every input I tried, 700
tests pass here and CI is green. But close the CRLF case first, or drop `#611`
from the closing keywords: on the input this branch is about, a wrapped value in
a CRLF file, the defect still is not reported where it truly stands.

Then #615, the `merge-source` gap as its own issue, and the flags after those.

maxonfjvipon:
Reviewed this on the branch. The core of it is right, and I checked the claims
rather than taking them: on a value wrapped four times, master reports `5:151`
against a line 28 characters long and this branch reports `9:30`, which is
exactly where the axis stands. `npm test` and `npm run coverage` pass locally,
branches 100% at 902/902.

I also re-derived the three moved pack rows by hand instead of trusting the
shift. They are right. In `count-in-complex-expressions.yaml` the second defect
anchors on the `0` of `0 = count(@x)`, which sits at raw column 51 and at
decoded column 48, so the `+3` is precisely the `&gt;` standing in front of it —
and it confirms the point that these were wrong on a single line all along. The
new pack's `[[5, 23], [7, 22]]` matches its fixture to the character.

One thing I would fix before this merges.

### The suppression hole is closed one level short

`from` is `node.lineNumber`, which for an attribute is the line the *attribute*
opens on, not the line the element does. The two part company as soon as the
start tag itself wraps, and then the directive falls outside `[from, line]`
again:

```xml
<!-- xslint-disable-next-line using-namespace-axis -->
<xsl:if
    test="@a
          and namespace::*">
```

```text
[WARNING] Unused xslint-disable directive at tagwrap.xsl:5
[WARNING] tagwrap.xsl(8:19) The namespace:: axis is deprecated…
```

The directive is announced unused and the defect fires anyway — both halves of
the symptom this PR added two tests to prevent, one nesting level further out.
The element opens on line 6, the attribute on line 7, so `from` is 7 and the
directive reaching line 6 misses by one.

`node.ownerElement.lineNumber` for an attribute covers it, and collapses to
exactly today's value whenever the tag does not wrap, so the existing fixture
keeps passing. A `wrapped-tag.xsl` beside `wrapped.xsl` would pin it.

### Filed separately

#626, for CRLF. `character` folds a named entity into one decoded character but
not a `\r\n`, which XML 1.0 §2.11 normalizes the same way, so the position
drifts one column left per preceding wrap — the fixture above reports `9:26`
against a true `9:30`. `--fix` degrades safely, skipping every wrapped value
with "the source no longer matches", and master is far worse than this branch,
so it is the remainder of #611 rather than anything this PR broke. Worth knowing
that `.gitattributes` pins every fixture to `eol=lf`, so the green
`build (windows-2022)` job cannot reach it.

### Two nits

`src/fixer.js:78` still documents `fix: {line, col, offset: …}` in its JSDoc,
though `offset` is gone from the anchor this PR simplified.

`offsetAt` and `placeAt` each split the whole file, and `defect` calls both for
every defect, so positioning is O(file × defects). Nothing to act on at the
sizes involved; noting it since the file is new and the shape is easy to fix
later by hoisting the line index.

maxonfjvipon:
All four fixed in `1e84c55`, and the tag-wrap catch was the one that mattered —
it is the same symptom the two new tests were meant to prevent, one level out,
and I had stopped at the attribute without asking what held it.

```text
before   Unused xslint-disable directive at tagwrap.xsl:5
         tagwrap.xsl(8:19) The namespace:: axis…
after    (both silent)
```

`from` is now the holding element's line, which collapses to exactly the old
value whenever the tag does not wrap, so `wrapped.xsl` keeps passing and
`wrapped-tag.xsl` pins the new case.

### CRLF

Fixed here rather than left to #626, since the work was already done and
shipping a position I know to be wrong is worse than a slightly wider PR:

```text
        before   after
LF      8:30     8:30
CRLF    8:27     8:30
```

`--fix` lands on a CRLF file now instead of skipping it. Your note about
`.gitattributes` saved me a fixture that would have silently normalised to LF
and passed for the wrong reason — the case is built at runtime from the LF
fixture instead. The unterminated `&` is closed in the same function.

### The nits

Both done: the `offset` is gone from `src/fixer.js`'s JSDoc, and the line
offsets are computed once per text rather than per call, so positioning is no
longer O(file × defects).

The widening is in `README.md` now, with its cost stated — one defect in a
wrapped value cannot be silenced without its neighbours. You were right that a
docblock was the wrong place for a behaviour change a user meets.

### Filed

#627 for `xsl:merge-source` — `for-each-item` and `for-each-source` are absent
from `ATTRIBUTES`, so the `select` beside them is flagged and they are not,
which is what makes it easy to miss. Thanks for pressing on it; that is the
reference I could not place earlier and it would have been lost.

#628 for `redundant-whitespace` on a wrapped expression.

That leaves #615 as the last thing between these two checks and the flag, and
only the half that is ours: `attribute::  name` shortens to `@  name`.

962 tests, 100% branches (913/913).

maxonfjvipon:
Re-reviewed at 1e84c55. Everything I raised is addressed, and I re-ran each
check rather than reading the commit message for it. 716 passing locally, 100%
of statements (5157) and branches (913), all 18 CI checks green.

The CRLF file now reports positions identical to the LF one — `9:30`, `5:28`,
`6:31`, `7:31`, `8:31`, every one of them where the text actually stands — and
`--fix` lands on the span instead of declining the file. `wrapped-tag.xsl` is
silent with no unused-directive warning. `(node.ownerElement || node.parentNode)`
is a better answer than the `ownerElement` I suggested, since it carries a text
node to its element too, which my version would have left behind.

The unclosed `&` is a good catch. I had looked at that `indexOf` answering `-1`
and talked myself out of it on the grounds that well-formed XML cannot carry a
bare `&` — which is true of the document and beside the point, because the guard
exists for source that has already been half-edited.

Two things left, neither of them blocking.

### A CR-only file reports `1:NaN`

```text
multi-cr.xsl(1:NaN) Verbose axis specifiers child::, attribute::, …
```

`character` now takes a lone `\r` as a line ending, but `starts` still splits on
`\n` alone, so the two halves of the same commit disagree about what ends a
line. `@xmldom/xmldom` counts the `\r`, hands back a `lineNumber` past the end of
the offsets array, and `starts(text)[line - 1]` is `undefined` — every defect in
the file then reports a `NaN` column. Splitting on `/\r\n|\r|\n/` settles it.

This one is worth folding in here rather than filing, because it is the branch
that introduces it: on master `defect` never called `offsetAt`, so a CR-only
file got a wrong column rather than a `NaN` one. The input is a museum piece, but
`NaN` in front of a user is a bad way to find that out.

### #629, for the fix that cannot cross a wrap

Filed rather than raised inline, since it predates this branch. `decodes` folds
entities and line endings but not XML 1.0 §3.3.3, so a fix whose value spans a
wrap never matches and is declined as "the source no longer matches" — which
blames a file nothing has touched. #628 covers the `redundant-whitespace` half
by having the check stop flagging indentation; the other two want the opposite,
since `count(item)\n &gt; 0` becoming `exists(item)` absorbs the wrap rather
than reflowing anything:

```text
[WARNING] Skipped fixing unabbreviated-axis at sheet.xsl:7, the source no longer matches
[WARNING] Skipped fixing count-compared-to-zero at sheet.xsl:5, the source no longer matches
```

Positions are untouched by it — one raw character still maps to one decoded
character either way — which is exactly why the work in this PR stands on its
own and that one can wait.

maxonfjvipon:
Folded in at `21b4f52`. You were right that it belongs here — the walk learned
that a lone carriage return ends a line and the line index did not, so the two
halves of one commit disagreed about what a line is. xmldom counted the return,
handed back a line past the end of the offsets, and `starts(text)[line - 1]`
came out `undefined`.

```text
before   cronly.xsl(1:NaN)
after    cronly.xsl(5:27)     its LF twin: (5:27)
```

`starts` splits on all three spellings §2.11 names now, and the source test
carries a row per ending so the two halves cannot drift apart again.

Agreed on #629 being separate — positions are untouched by it, one raw
character still maps to one decoded character whichever way a fix spans a wrap,
which is why it does not reach anything here.

966 tests, 100% of statements (5163) and branches (913). Out of draft and ready
to merge.

After it lands: #615's `@  name` half, then the flags.

maxonfjvipon (commented):
Everything from the last round is resolved, and I checked each against
`1e84c55` rather than taking the description for it:

| | before | now |
| --- | --- | --- |
| CRLF, three wraps ahead of the defect | `8:16` | `8:19`, identical to LF |
| `--fix` on a CRLF wrapped value | skipped, file untouched | applied, `\r\n` kept |
| unterminated `&` | walk restarted at offset 0 | no character, advances one |
| directive widening | docblock only | in `README.md` |
| per-defect rescan | whole file split, twice | one line table per text |

`from` moving to the holding element is better than what I reviewed — the start
tag can wrap before the value does, and `node.lineNumber` would have missed it.
#627 and #628 give both of the loose observations a home.

One new thing, and it comes from the fix to my own performance note.

### `LINES` never lets a source go

The memo is a module-level `Map` keyed by the whole source string, and nothing
clears it:

```text
LINES entries after 200 lint() calls: 200
heap retained: 25 MB          (measured after an explicit global.gc())
```

For the CLI this costs nothing, since the process exits. But `lint` and `fixed`
are the package `main` export precisely so an embedder can lint a buffer without
shelling out, and the LSP server in #336 is the stated reason. An editor
re-lints on a keystroke, so every buffer version of every open file is retained
for the life of the server, each one holding its entire text as a live key.

The docblock reads `@type {WeakMap|Map}`, which cannot be satisfied as written —
a `WeakMap` rejects a string key. Keying the memo on the source object instead,
the `{file, content}` the corpus already carries and which is a valid `WeakMap`
key, makes the annotation true and lets the table fall away with the corpus.
Clearing at the end of `lint` would do as well, in one line.

Worth closing before this merges rather than after: it sits in the shipped
`lint` path, and the embedder that trips over it is the one this API exists for.

### Everything else

`starts` still splits on `\n` alone, so a source using a lone `\r` collapses to
one line — §2.11 names that spelling, and `character` already honours it. That
looked like work in progress rather than an oversight, so I have left it here
rather than filing it.

Nothing else stands in the way. On the freeze, #615 and #627 both land on these
two checks — #615 on a fixture `whitespaced-axis.yaml` blesses, and on `@  name`,
which is our own fix output rather than fontoxpath's — so the flags still wait
on those two, not on this branch.

maxonfjvipon:
Right, and it is the worse half of my fix to your performance note — I traded an
O(file × defects) rescan for unbounded retention and only annotated the trade,
with a `@type {WeakMap|Map}` that, as you say, cannot be satisfied as written.
Measured before touching it:

```text
heap growth over 2000 lint() calls, ~2 KB each
before   5 MB
after    1 MB      the weight of the run itself
```

Fixed in `594573b`, though not the way you suggested. Keying on the source
object would work for `defect`, which holds one, but `fixed` calls `offsetAt`
with a plain string out of `sources`, so the module would have had to learn the
shape of a corpus entry to stay a text utility. Positions are asked for one
source at a time — every defect in a file, then every fix in it — so a
single-entry memo gets the same hit rate and holds one text rather than all of
them. No signature changes, and nothing survives the source in hand.

Clearing at the end of `lint` was the other option and I left it, because it
only holds while every entry point remembers to do it, and `fixed` is exported
beside `lint`.

### On `starts`

That one is already in — `21b4f52`, pushed after the commit you reviewed. It
splits on all three spellings now, and a CR-only file reports `5:27` where it
reported `1:NaN`, identical to its LF twin. The source test carries a row per
ending so the walk and the index cannot drift apart again.

966 tests, 100% of statements (5167) and branches (913).

Agreed on where the freeze stands: #615 and #627 both land on these two checks,
and neither is this branch's.

Issue #629, opened 2026-08-02
`decodes` cannot match a fix value across a line wrap in `src/fixer.js`


maxonfjvipon:
The decode-walk in `decodes` at `src/fixer.js:75` compares a fix's decoded
`value` against the raw source one character at a time, and `character` in
`src/source.js` folds a named entity and a line ending into what a parser makes
of them. It never applies the one normalisation left, XML 1.0 §3.3.3, which
turns a line ending inside an attribute value into a space. So the value holds a
space exactly where the source holds a newline, the comparison fails on that
character, and the whole fix is declined:

```js
const matched = [...value].every((char) => {
  const [decoded, next] = raw < content.length ?
    character(content, raw) :
    [null, raw]
  raw = next
  return decoded === char
})
```

Any fix whose matched text crosses a wrap is unreachable, whatever check offers
it. One stylesheet with three wrapped expressions loses all three:

```text
[WARNING] Skipped fixing unabbreviated-axis at sheet.xsl:7, the source no longer matches
[WARNING] Skipped fixing count-compared-to-zero at sheet.xsl:5, the source no longer matches
[WARNING] Skipped fixing redundant-whitespace at sheet.xsl:5, the source no longer matches
```

None of those files had been touched. "The source no longer matches" is the
wording for a span an earlier edit has already moved, so a user reading it goes
looking for a change that never happened, and the defect sits in the report
across every rerun with no way to clear it.

#628 covers the `redundant-whitespace` half by having the check stop flagging
indentation, which is right on its own terms and leaves the other two checks
where they are. The rest of the wrapped-expression cases want the opposite
outcome: `count(item)\n &gt; 0` should become `exists(item)` and `parent::\n
node()` should become `..`, both of which absorb the wrap into a replacement
that no longer spans anything.

Letting `decodes` accept a line ending in the source where the value carries a
space is the smallest change that reaches all of them, and it belongs in the
matcher rather than in `character`, which has no way of telling an attribute
value from element content — where a newline stays a newline and no such licence
applies. That distinction is the reason to keep the relaxation out of the shared
decoder.

Issue #607, opened 2026-07-31
Pack harnesses assert no fix when a pack omits `found.fixes`


maxonfjvipon:
#519 gave every code-based linter pack a `found.fixes` array and every harness
the assertion that reads it. The assertion is opt-in per pack, though: it starts
from `yml.found.fixes || []`, so a pack that never declares the key hands the
loop an empty array and the body never runs.

```js
const fixes = yml.found.fixes || []
fixes.forEach((expected, index) => {
  assert.equal(
    defects[index].fix ? defects[index].fix.replacement : null,
    expected,
  )
})
```

That block sits, copied verbatim, in thirteen test files —
`test/using-namespace-axis-linter.test.js:31` is one — and 29 of the 63 packs
they discover carry no `fixes` key. Three of those pin real positions:
`test/resources/xpath-format-packs/redundant-whitespace.yaml` declares three
positions and no fixes, and the linter attaches the replacements `" "`, `""`,
and `""` at exactly those positions with nothing asserting them, which is the
fixable check #519 set out to pin in the first place. The other two —
`result-namespace-packs/two-leaking-prefixes.yaml` and
`name-packs/name-compared-to-invalid-name.yaml` — are report-only today, so a
fix wrongly attached there would also pass unnoticed.

Nothing catches the omission. `test/conformance.test.js` never looks at pack
shape, while CLAUDE.md states the convention as settled fact — "A code-based
linter's pack also carries `found.fixes` aligned with `positions` ... which the
harness asserts too" — and demands that every consistency convention be
machine-enforced. So the next pack drops the key with a green build, and a
report-only contract such as `using-namespace-axis`, whose six packs were
completed in #600 precisely to lock it, can be broken by an accidentally
attached fix without anything turning red.

Add one `conformance.test.js` case that walks the `*-packs` directories of the
code-based harnesses and requires `found.fixes` to exist with the same length as
`found.positions`, then fill the 29 missing keys. Folding that thirteen-times
copy into one shared pack harness is the natural home for the check, since the
length assertion then lives in a single place.

PR #630, opened 2026-08-02
bug(#615): shorten an axis cleanly, and not at all inside a pattern

### LINKED ISSUE #627: xsl:merge-source XPath attributes are never read in `src/attributes.js`

maxonfjvipon:
Closes #627 and the half of #615 that is ours, fixes the #583-class defect the
review turned up, and takes both axis checks to `mature: true`.

The branch grew past its title. In order of weight:

### The #583-class defect: an abbreviated step inside a pattern

`--fix` was rewriting `match` attributes in the safe tier, and a pattern is not
an expression. XSLT 3.0 §5.5.2 makes `.` a `PredicatePattern` — an alternative to
the whole union grammar rather than a term inside it — so three shapes stopped
compiling, and a fourth changed which template wins:

```text
match="y|self::node()"     ->  match="y|."     will not compile
match="(self::node())"     ->  match="(.)"     will not compile
match="y|self::node()[1]"  ->  match="y|.[1]"  will not compile
match="self::node()"       ->  match="."       demoted: -0.5 becomes -1
```

Before 3.0 it is simpler still: the self axis is not a pattern axis at all, so
the longhand was never legal there.

`src/attributes.js` now says which of the names it reads hold a pattern — nothing
could ask it before, which is why no fixer could withhold on that basis — and the
check leaves a longhand step in one alone entirely. Not just the fix: the message
told the reader to write `.` where the motive tells them not to, and for
`match="y|self::node()"` that edit does not compile. The check already answers
this shape, staying quiet on `parent::n` because nothing shorter exists; a
pattern is the same case.

`child::` and `attribute::` keep their fixes there, both being axes a pattern
step admits in every version — and, tested by declaration order on two
processors, both leaving the weighing untouched.

### #615, the gap the colons left behind

| | before | now |
| --- | --- | --- |
| `attribute::  name` | `@  name` | `@name` |
| `child::  title` | `  title` | `title` |

The gap belongs to the axis, so the fix takes it. Master's output was a fresh
defect: `@  name` draws `invalid-xpath-expression` and `  title` draws
`redundant-whitespace`. The other half of #615 — `isValid` refusing `child ::x`
outright — is fontoxpath's and stays open there.

### #627, two attributes nothing read

```text
<xsl:merge-source for-each-item="child::doc" select="child::record">
                                 ^ silent                ^ reported
```

Both read now, and `use-when` with them. The element looking half-covered rather
than uncovered is what kept it quiet.

### Positioning, from the review of #611

`skip` walked the raw text one character at a time for every defect. Nothing in a
span can be wider than one character unless the span holds an `&` or a `\r`, so
where it holds neither the answer is the count itself: 62.9 ms to 6.6 ms over
80,000 calls, with the walk unchanged for the spans that need it and both paths
pinned in `test/source.test.js`.

### Coverage

Two packs pin what no pack could before: `abbreviated-step-inside-a-pattern.yaml`
holds all eight withheld shapes at `amount: 0` and fails on either mistake —
restoring the fix or restoring the report — and a `values:` list beside `fixes:`
pins the span a fix replaces, which is the part of a fix nothing asserted.
Reverting either change now fails the suite; neither did before.

Three packs named a check that does not exist, so the gate counting a format
check's coverage never saw them. Pack names are checked against the real checks
across every directory now, which found nine more of the same.

### `mature: true`

Both checks. 37 expressions past them with no mismatch, on top of 22 packs for
`unabbreviated-axis` and 16 for `using-namespace-axis`: both roots, a simplified
stylesheet, non-`xsl` prefixes, every version and every legal spelling of one, a
raised and a lowered subtree, an `xsl:result-document`, a wrapped value, patterns,
and every node kind — attribute, AVT, text value template, shadow attribute,
CDATA.

The commit subjects on this branch read `bug(#583)` for the pattern work. #583
itself is `starts-with-double-slash` in `src/fixers.js`, still open and untouched
here — the reference is to its class, not to that instance, and merging does not
close it (`closingIssuesReferences` is `#627` alone). The title now reads #615,
which the branch is named for and genuinely advances. No issue is filed for the
instance, since it is fixed in this branch rather than outstanding.

Filed rather than folded in: #631, a pattern axis illegal before 3.0 goes
unreported, which is cheap now that `pattern` and `versionOf` are both available
at the node; #632, `using-namespace-axis` naming two functions that cannot stand
as a pattern step.

974 tests, 100% branch coverage (920/920), ESLint, xcop, markdownlint and the
fixtures gate clean locally.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -113,7 +113,14 @@ a linter and its suppression names cannot drift apart.
 No code-based linter selects attributes by name on its own. `src/attributes.js`
 hands it every expression a stylesheet carries: an XPath or pattern attribute *of
 an XSLT element*, whole, plus each expression an attribute value template encloses
-in braces, offset by where it starts inside the value (#579). In an XSLT 3.0
+in braces, offset by where it starts inside the value (#579). Each one says
+whether it is a `pattern`, because the two are different languages and a rewrite
+legal in one can be a syntax error in the other — `.` stands alone in an
+expression but is a whole pattern rather than a step inside one, so `y|.` does
+not parse and a bare `match="."` outranks the `self::node()` it replaced. A
+fixer touching a step withholds inside a pattern; where the shorter form does
+not exist there at all, the check does not report it either, the way it stays
+quiet on a `parent::n` (#583). In an XSLT 3.0
 stylesheet it also reads a **text value template** — the braces of a text node
 whose nearest `expand-text`/`xsl:expand-text` is on — and a **shadow attribute**
 (`_select` for `select`), the same expressions the modern idiom hides outside an
@@ -177,7 +184,8 @@ motive, or ships untested fails the build.
   (`metaOf`, `suppressed`, `defect`) and reads its expressions from
   `src/attributes.js`'s `expressionsOf` (every XPath/pattern attribute of an XSLT
   element, plus every expression an attribute value template, a 3.0 text value
-  template, or a shadow attribute carries) unless it has a documented reason to
+  template, or a shadow attribute carries, each flagged `pattern` or not) unless
+  it has a documented reason to
   narrow — then it narrows through `selectorOf`, never a hand-written `//@name`,
   which an ESLint `no-restricted-syntax` selector bans.
 
@@ -362,7 +370,7 @@ the harness asserts too.
 | `src/*-linter.js` | Code-based `checks/format/*.yaml`, one construct each (axis, namespace, count, name, ...); see the flow diagram |
 | `src/checks.js` | Shared for code-based linters: `metaOf`, `suppressed`, `defect(check, meta, source, node, offset, fix)` — walks the raw text so a wrapped or entity-shifted value reports where it truly stands (#611) |
 | `src/source.js` | Raw-text walking shared by `checks` and `fixer`: `offsetAt`, `placeAt`, `character`, `skip` |
-| `src/attributes.js` | `expressionsOf(xsl)` — every expression a stylesheet carries: a bare/AVT attribute, a 3.0 text value template, or a shadow attribute; `selectorOf(name)` for a linter that narrows |
+| `src/attributes.js` | `expressionsOf(xsl)` — every expression a stylesheet carries: a bare/AVT attribute, a 3.0 text value template, or a shadow attribute, each saying whether it is a `pattern`; `PATTERNS` names the five attributes that hold one; `selectorOf(name)` for a linter that narrows |
 | `src/xsl-version.js` | `versionOf(node)` — the version in force at a node, from the nearest ancestor's `@version` (XSLT element) or `@xsl:version` (literal result element), canonicalised as a decimal; `since(version, floor)` for a lower-bound gate; shared `MODERN`/`KNOWN`/`DECIMAL` |
 | `src/comparisons.js` | `comparedToZero` — shared scan for a call compared with `0`/`1` (count, string-length) |
 | `src/expressions.js` | `masked`/`closes` lexer helpers (node-set, double-negation, boolean-call); `enclosed` — the expressions an AVT holds in its braces |
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -20,7 +20,19 @@ const ATTRIBUTES = [
   'select', 'test', 'use', 'value', 'group-by', 'group-adjacent', 'key',
   'initial-value', 'xpath', 'context-item', 'with-params', 'namespace-context',
   'match', 'count', 'from', 'group-starting-with', 'group-ending-with',
-  'for-each-item', 'for-each-source',
+  'for-each-item', 'for-each-source', 'use-when',
+]
+
+/**
+ * The attributes among those that hold a *pattern* rather than an expression.
+ * The two are not one language: a pattern has its own grammar, narrower where
+ * an expression is free — `.` may stand alone there but not inside a union or
+ * a parenthesis — and it decides which template wins rather than what a step
+ * selects. A linter rewriting one has to know which of the two it is holding.
+ * @type {Array.<string>}
+ */
+const PATTERNS = [
+  'match', 'count', 'from', 'group-starting-with', 'group-ending-with',
 ]
 
 /**
@@ -103,9 +115,12 @@ const carried = function(node, bare, three) {
     (bare.has(node) || (three && shadow(node)))
   const braced = node.nodeType === 2 || (three && expands(node))
   return whole ?
-    [{node: node, start: 0, expression: node.nodeValue}] :
+    [{
+      node: node, start: 0, expression: node.nodeValue,
+      pattern: PATTERNS.includes(node.nodeName.replace(/^_/, '')),
+    }] :
     braced ? enclosed(node.nodeValue).map((found) => ({
-      node: node, start: found.offset, expression: found.value,
+      node: node, start: found.offset, expression: found.value, pattern: false,
     })) : []
 }
 
@@ -118,7 +133,7 @@ const carried = function(node, bare, three) {
  * reported — and fixed — where it truly stands.
  * @param {Document} xsl - XSL document parsed as {@link Document}
  * @return {Array.<{node: Node, start: number, expression: string}>} - The
- *  expressions found
+ *  expressions found, each saying whether it is a pattern
  */
 const expressionsOf = function(xsl) {
   const bare = new Set(nodes(xsl, SELECTOR))
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -11,13 +11,16 @@ const {XSLT, since, versionOf} = require('./xsl-version')
  * Attributes that hold an XPath expression or a pattern — every place a
  * construct like an axis, a predicate, or a `count(...)` call can appear.
  * Patterns (`match` and the grouping attributes) are included, so a construct
- * in a template match is caught too, not only in the expression stream.
+ * in a template match is caught too, not only in the expression stream. So are
+ * the two an `xsl:merge-source` holds beside its `select`, which carry an
+ * expression the same way it does (#627).
  * @type {Array.<string>}
  */
 const ATTRIBUTES = [
   'select', 'test', 'use', 'value', 'group-by', 'group-adjacent', 'key',
   'initial-value', 'xpath', 'context-item', 'with-params', 'namespace-context',
   'match', 'count', 'from', 'group-starting-with', 'group-ending-with',
+  'for-each-item', 'for-each-source',
 ]
 
 /**
--- a/src/resources/checks/format/unabbreviated-axis.yaml
+++ b/src/resources/checks/format/unabbreviated-axis.yaml
@@ -3,3 +3,4 @@
 ---
 severity: warning
 message: Verbose axis specifiers child::, attribute::, parent::node(), or self::node() are used. Replace them with the node name alone, @, .., or . respectively.
+mature: true
--- a/src/resources/checks/format/using-namespace-axis.yaml
+++ b/src/resources/checks/format/using-namespace-axis.yaml
@@ -3,3 +3,4 @@
 ---
 severity: warning
 message: "The namespace:: axis is deprecated in XSLT 2.0 and later. Use in-scope-prefixes() and namespace-uri-for-prefix() instead."
+mature: true
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -60,11 +60,12 @@ is legal in every version.
 In a pattern only the first two rows hold, and not by coincidence. Before 3.0 a
 pattern step admits the child and attribute axes and no others — a 1.0 processor
 rejects `match="self::r"` outright — which is exactly the pair whose
-abbreviations are safe there. XSLT 3.0 widened the step to the forward axes —
+abbreviations are safe there. XSLT 3.0 widened the step to the downward axes —
 `descendant`, `self`, `descendant-or-self`, `namespace` — so `match="self::r"`
 compiles on a 3.0 processor; the pair is still the pair, because those two drop
-a step without touching how the rule is weighed. The reverse axes stayed out,
-which is why `parent::` never became a pattern step in any version.
+a step without touching how the rule is weighed. Every axis that reaches
+elsewhere stayed out, `following::` as much as `parent::`, which is why
+`parent::` never became a pattern step in any version.
 
 Neither abbreviated step survives the trip. For `..` there is nothing to weigh:
 `match="parent::node()"` is not a legal pattern in any version, so neither is
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -60,22 +60,25 @@ is legal in every version.
 In a pattern only the first two rows hold, and not by coincidence. Before 3.0 a
 pattern step admits the child and attribute axes and no others — a 1.0 processor
 rejects `match="self::r"` outright — which is exactly the pair whose
-abbreviations are safe there. XSLT 3.0 widened the step to `descendant`, `self`
-and the rest, so `match="self::r"` compiles on a 3.0 processor; the pair is
-still the pair, because those two drop a step without touching how the rule is
-weighed.
+abbreviations are safe there. XSLT 3.0 widened the step to the forward axes —
+`descendant`, `self`, `descendant-or-self`, `namespace` — so `match="self::r"`
+compiles on a 3.0 processor; the pair is still the pair, because those two drop
+a step without touching how the rule is weighed. The reverse axes stayed out,
+which is why `parent::` never became a pattern step in any version.
 
-The abbreviated steps never survive the trip, for a different reason in each
-era. Before 3.0 the longhand is not a legal pattern to begin with — the self
-axis is not among the two — so the question of shortening it never arises. From
-3.0 it is legal, and then the short form is not a synonym: `.` is a pattern in
-its own right rather than a step inside one, so it cannot stand in a union or
-behind a parenthesis, `match="y|self::node()"` being legal where `match="y|."`
-is not; and standing alone it is outranked by the longhand it replaced, `.`
-carrying a default priority of -1 where `self::node()` carries -0.5, so the
-rewrite quietly hands the node to a template that used to lose. Leave a longhand
-step as it is in a `match`, a `count`, a `from`, or an `xsl:for-each-group`
-boundary.
+Neither abbreviated step survives the trip. For `..` there is nothing to weigh:
+`match="parent::node()"` is not a legal pattern in any version, so neither is
+`match=".."`, and the question of shortening one into the other never arises.
+
+`.` is the one with two eras. Before 3.0 the longhand is not a legal pattern
+either, the self axis not being among the two. From 3.0 it is, and then the
+short form is no synonym for it: `.` is a pattern in its own right, not a step
+inside one, so it cannot stand in a union or behind a parenthesis,
+`match="y|self::node()"` being legal where `match="y|."` is not; and standing
+alone it is outranked by the longhand it replaced, `.` carrying a default
+priority of -1 where `self::node()` carries -0.5, so the rewrite quietly hands
+the node to a template that used to lose. Leave a longhand step as it is in a
+`match`, a `count`, a `from`, or an `xsl:for-each-group` boundary.
 
 `descendant-or-self::node()` is the exception that is not worth taking. Its
 short form is `//`, but the two are easy to confuse in use: a `//` that opens a
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -58,22 +58,24 @@ out or parenthesise the abbreviation, since `(.)[1]` is a filter expression and
 is legal in every version.
 
 In a pattern only the first two rows hold, and not by coincidence. Before 3.0 a
-pattern step admits the child and attribute axes and nothing else — `match="c"`,
-`match="child::c"` and `match="@a"` are the whole of it, and a 1.0 processor
+pattern step admits the child and attribute axes and no others — a 1.0 processor
 rejects `match="self::r"` outright — which is exactly the pair whose
 abbreviations are safe there. XSLT 3.0 widened the step to `descendant`, `self`
 and the rest, so `match="self::r"` compiles on a 3.0 processor; the pair is
 still the pair, because those two drop a step without touching how the rule is
 weighed.
 
-The abbreviated steps never survive the trip. Before 3.0 there is nothing to
-shorten to: a pattern has no `.` step, so `match="self::node()"` does not
-compile at all. From 3.0 `match="."` does compile, and is not a synonym — it is
-a pattern in its own right rather than a step inside one, so it cannot stand in
-a union or behind a parenthesis, `match="y|self::node()"` being legal where
-`match="y|."` is not; and standing alone it outranks the longhand it replaced,
-quietly handing the node to a template that used to lose. Leave a longhand step
-as it is in a `match`, a `count`, a `from`, or an `xsl:for-each-group` boundary.
+The abbreviated steps never survive the trip, for a different reason in each
+era. Before 3.0 the longhand is not a legal pattern to begin with — the self
+axis is not among the two — so the question of shortening it never arises. From
+3.0 it is legal, and then the short form is not a synonym: `.` is a pattern in
+its own right rather than a step inside one, so it cannot stand in a union or
+behind a parenthesis, `match="y|self::node()"` being legal where `match="y|."`
+is not; and standing alone it is outranked by the longhand it replaced, `.`
+carrying a default priority of -1 where `self::node()` carries -0.5, so the
+rewrite quietly hands the node to a template that used to lose. Leave a longhand
+step as it is in a `match`, a `count`, a `from`, or an `xsl:for-each-group`
+boundary.
 
 `descendant-or-self::node()` is the exception that is not worth taking. Its
 short form is `//`, but the two are easy to confuse in use: a `//` that opens a
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -10,7 +10,10 @@ nothing — the child axis is the one a bare name already uses, so writing it ou
 draws the eye to the one part of the step that carries no information.
 
 The abbreviations are defined by XPath itself and select precisely the same
-nodes as the longhand:
+nodes as the longhand — in an expression. A `match` is not an expression but a
+pattern, a narrower grammar with its own rules about where `.` may stand and its
+own way of deciding which template wins, so the table below is about `select`
+and `test`, and a pattern is discussed at the end:
 
 | Longhand | Short |
 | --- | --- |
@@ -53,6 +56,15 @@ restriction, which is why the same expression compiles under a later `version`.
 Where a predicate follows the step in a 1.0 stylesheet, either leave it spelled
 out or parenthesise the abbreviation, since `(.)[1]` is a filter expression and
 is legal in every version.
+
+In a pattern only the first two rows hold. A pattern takes the child and
+attribute axes, so `match="child::chapter"` is `match="chapter"` and nothing
+changes. The abbreviated steps are another matter: `.` is a pattern in its own
+right rather than a step inside one, so it cannot stand in a union or behind a
+parenthesis — `match="y|self::node()"` is legal and `match="y|."` will not
+compile — and where it does parse it is weighed differently against a competing
+template. Leave a longhand step alone in a `match`, a `count`, a `from`, or an
+`xsl:for-each-group` boundary.
 
 `descendant-or-self::node()` is the exception that is not worth taking. Its
 short form is `//`, but the two are easy to confuse in use: a `//` that opens a
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -57,9 +57,10 @@ Where a predicate follows the step in a 1.0 stylesheet, either leave it spelled
 out or parenthesise the abbreviation, since `(.)[1]` is a filter expression and
 is legal in every version.
 
-In a pattern only the first two rows hold. A pattern takes the child and
-attribute axes, so `match="child::chapter"` is `match="chapter"` and nothing
-changes. The abbreviated steps are another matter: `.` is a pattern in its own
+In a pattern only the first two rows hold. A pattern is free with axes —
+`match="self::r"` and `match="descendant::c"` are both legal — and the child and
+attribute axes are among them, so `match="child::chapter"` is `match="chapter"`
+and nothing changes. The abbreviated steps are another matter: `.` is a pattern in its own
 right rather than a step inside one, so it cannot stand in a union or behind a
 parenthesis — `match="y|self::node()"` is legal and `match="y|."` will not
 compile — and where it does parse it is weighed differently against a competing
--- a/src/resources/motives/format/unabbreviated-axis.md
+++ b/src/resources/motives/format/unabbreviated-axis.md
@@ -57,15 +57,23 @@ Where a predicate follows the step in a 1.0 stylesheet, either leave it spelled
 out or parenthesise the abbreviation, since `(.)[1]` is a filter expression and
 is legal in every version.
 
-In a pattern only the first two rows hold. A pattern is free with axes —
-`match="self::r"` and `match="descendant::c"` are both legal — and the child and
-attribute axes are among them, so `match="child::chapter"` is `match="chapter"`
-and nothing changes. The abbreviated steps are another matter: `.` is a pattern in its own
-right rather than a step inside one, so it cannot stand in a union or behind a
-parenthesis — `match="y|self::node()"` is legal and `match="y|."` will not
-compile — and where it does parse it is weighed differently against a competing
-template. Leave a longhand step alone in a `match`, a `count`, a `from`, or an
-`xsl:for-each-group` boundary.
+In a pattern only the first two rows hold, and not by coincidence. Before 3.0 a
+pattern step admits the child and attribute axes and nothing else — `match="c"`,
+`match="child::c"` and `match="@a"` are the whole of it, and a 1.0 processor
+rejects `match="self::r"` outright — which is exactly the pair whose
+abbreviations are safe there. XSLT 3.0 widened the step to `descendant`, `self`
+and the rest, so `match="self::r"` compiles on a 3.0 processor; the pair is
+still the pair, because those two drop a step without touching how the rule is
+weighed.
+
+The abbreviated steps never survive the trip. Before 3.0 there is nothing to
+shorten to: a pattern has no `.` step, so `match="self::node()"` does not
+compile at all. From 3.0 `match="."` does compile, and is not a synonym — it is
+a pattern in its own right rather than a step inside one, so it cannot stand in
+a union or behind a parenthesis, `match="y|self::node()"` being legal where
+`match="y|."` is not; and standing alone it outranks the longhand it replaced,
+quietly handing the node to a template that used to lose. Leave a longhand step
+as it is in a `match`, a `count`, a `from`, or an `xsl:for-each-group` boundary.
 
 `descendant-or-self::node()` is the exception that is not worth taking. Its
 short form is `//`, but the two are easy to confuse in use: a `//` that opens a
--- a/src/source.js
+++ b/src/source.js
@@ -99,9 +99,12 @@ const character = function(content, at) {
  * The raw offset reached after skipping the given number of decoded characters,
  * so an offset into a parsed value maps back to its true place in the source
  * even when an entity ahead of it spans several source characters. Where the
- * span holds no `&` and no `\r` nothing can be wider than one character, and
- * the answer is the count itself — which is the ordinary case, and worth not
- * walking a character at a time to reach.
+ * span holds no `&` and no `\r` nothing in it can be wider than one character,
+ * and the answer is the count itself — the ordinary case, and worth not walking
+ * a character at a time to reach. Testing only that span is enough: every
+ * character before the first wide one is read one for one, so the first `&` or
+ * `\r` a walk could reach always sits below `at + count`, inside the span
+ * tested. A count below zero takes the walk, which answers `at`.
  * @param {string} content - Raw source text
  * @param {number} at - Zero-based offset to start from
  * @param {number} count - Number of decoded characters to skip
@@ -110,7 +113,7 @@ const character = function(content, at) {
 const skip = function(content, at, count) {
   const ahead = content.slice(at, at + count)
   let raw = at + count
-  if (ahead.includes('&') || ahead.includes('\r')) {
+  if (count < 0 || ahead.includes('&') || ahead.includes('\r')) {
     raw = at
     for (let seen = 0; seen < count; seen++) {
       raw = character(content, raw)[1]
--- a/src/source.js
+++ b/src/source.js
@@ -98,16 +98,23 @@ const character = function(content, at) {
 /**
  * The raw offset reached after skipping the given number of decoded characters,
  * so an offset into a parsed value maps back to its true place in the source
- * even when an entity ahead of it spans several source characters.
+ * even when an entity ahead of it spans several source characters. Where the
+ * span holds no `&` and no `\r` nothing can be wider than one character, and
+ * the answer is the count itself — which is the ordinary case, and worth not
+ * walking a character at a time to reach.
  * @param {string} content - Raw source text
  * @param {number} at - Zero-based offset to start from
  * @param {number} count - Number of decoded characters to skip
  * @return {number} - The raw offset after `count` decoded characters
  */
 const skip = function(content, at, count) {
-  let raw = at
-  for (let seen = 0; seen < count; seen++) {
-    raw = character(content, raw)[1]
+  const ahead = content.slice(at, at + count)
+  let raw = at + count
+  if (ahead.includes('&') || ahead.includes('\r')) {
+    raw = at
+    for (let seen = 0; seen < count; seen++) {
+      raw = character(content, raw)[1]
+    }
   }
   return raw
 }
--- a/src/xpath-axis-linter.js
+++ b/src/xpath-axis-linter.js
@@ -29,8 +29,10 @@ const names = [CHECK]
 
 /**
  * The abbreviation each axis specifier collapses to on its own, whatever node
- * test follows it. The verbatim text comes from the token, so a spaced
- * `child ::` is stripped in full rather than leaving its whitespace behind.
+ * test follows it. What gets replaced is measured by `spans`, not taken from
+ * the token: the token carries the gap in front of the colons and `spans`
+ * reaches the one behind them, so a spaced `child ::  a` is stripped in full
+ * from either side.
  * @type {{[type: string]: {replacement: string}}}
  */
 const SHORT = {
@@ -63,7 +65,10 @@ const SPAN = 4
  * whitespace behind its colons. XPath allows a gap there, and it belongs to the
  * axis rather than to the node test, so shortening `attribute::  name` has to
  * take the gap with it — leaving `@  name` would be a fix that reads worse than
- * what it replaced.
+ * what it replaced. A gap the source wrote as a line break cannot be told apart
+ * here — the parser turned it into a space before the lexer saw it — so such a
+ * fix reaches the source as a run of spaces, fails to match, and is declined
+ * rather than misapplied. That is #629's to settle, not this function's.
  * @param {Array.<{type: string, value: string, start: number}>} tokens - Tokens
  * @param {number} index - Index of the axis token
  * @return {number} - Offset just past the axis and the gap behind it
@@ -113,9 +118,11 @@ const afterNode = function(tokens, index) {
  * comments are never seen because the lexer keeps those whole.
  * @param {string} expression - Xpath expression or pattern
  * @param {boolean} modern - Whether the stylesheet declares XSLT 2.0 or later
+ * @param {boolean} pattern - Whether the expression is a pattern, where the
+ *  step abbreviations are not the synonyms they are in an expression
  * @return {Array.<{offset: number, fix: ?object}>} - Axes and their fixes
  */
-const abbreviable = function(expression, modern) {
+const abbreviable = function(expression, modern, pattern) {
   const tokens = tokenized(expression)
   const found = []
   tokens.forEach((token, index) => {
@@ -131,7 +138,7 @@ const abbreviable = function(expression, modern) {
     } else if (step !== null) {
       found.push({
         offset: token.start,
-        fix: step.predicated && !modern ? undefined : {
+        fix: pattern || (step.predicated && !modern) ? undefined : {
           value: expression.slice(token.start, step.end),
           replacement: STEP[token.type].replacement,
         },
@@ -154,9 +161,11 @@ const lintByAxis = function(corpus, suppressions = []) {
   const defects = []
   if (!suppressed(CHECK, suppressions)) {
     for (const source of corpus) {
-      for (const {node, start, expression} of expressionsOf(source.xsl)) {
-        const modern = since(versionOf(node), MODERN)
-        for (const {offset, fix} of abbreviable(expression, modern)) {
+      for (const held of expressionsOf(source.xsl)) {
+        const {node, start, expression} = held
+        for (const {offset, fix} of abbreviable(
+          expression, since(versionOf(node), MODERN), held.pattern,
+        )) {
           defects.push(defect(CHECK, META, source, node, start + offset, fix))
         }
       }
--- a/src/xpath-axis-linter.js
+++ b/src/xpath-axis-linter.js
@@ -59,6 +59,24 @@ const STEP = {
 const SPAN = 4
 
 /**
+ * Offset just past the axis at the given token index, counting in any
+ * whitespace behind its colons. XPath allows a gap there, and it belongs to the
+ * axis rather than to the node test, so shortening `attribute::  name` has to
+ * take the gap with it — leaving `@  name` would be a fix that reads worse than
+ * what it replaced.
+ * @param {Array.<{type: string, value: string, start: number}>} tokens - Tokens
+ * @param {number} index - Index of the axis token
+ * @return {number} - Offset just past the axis and the gap behind it
+ */
+const spans = function(tokens, index) {
+  const gap = tokens[index + 1]
+  const axis = tokens[index]
+  return gap !== undefined && gap.type === TOKENS.WHITESPACE ?
+    gap.start + gap.value.length :
+    axis.start + axis.value.length
+}
+
+/**
  * The `node()` test that follows the axis token at the given index, or null
  * when the axis carries another node test. The whitespace XPath allows between
  * the pieces is insignificant, so `node ( )` names the same test as `node()`.
@@ -105,7 +123,10 @@ const abbreviable = function(expression, modern) {
     if (SHORT[token.type]) {
       found.push({
         offset: token.start,
-        fix: {value: token.value, replacement: SHORT[token.type].replacement},
+        fix: {
+          value: expression.slice(token.start, spans(tokens, index)),
+          replacement: SHORT[token.type].replacement,
+        },
       })
     } else if (step !== null) {
       found.push({
--- a/src/xpath-axis-linter.js
+++ b/src/xpath-axis-linter.js
@@ -115,11 +115,12 @@ const afterNode = function(tokens, index) {
  * gave the context item a predicate list, `.` and `..` were an AbbreviatedStep,
  * which takes no predicate, so `self::node()[1]` is only reported on a 1.0
  * sheet: `.[1]` is a syntax error there. A longhand step in a pattern goes
- * unreported for two reasons that meet: before 3.0 a pattern has no `.` step to
- * offer, so there is nothing shorter to name, and from 3.0 `match="."` does
- * compile but is not a synonym — a pattern in its own right rather than a step
- * inside one, illegal in a union and outranking the longhand where it stands
- * alone. Either way the check has no shorter form to recommend. Axes inside
+ * unreported for a different reason in each era: before 3.0 the longhand is not
+ * a legal pattern at all, the self axis not being one a pattern step admits,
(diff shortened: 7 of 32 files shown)


maxonfjvipon:
Reviewed at ad6bf7f on the branch, re-running each claim rather than reading the
diff for it. The suite and the coverage gate pass locally, 916/916 branches.

**The #615 half is right, and the table understates it.** Master does not merely
leave the gap behind — its output is a fresh defect. `attribute::  spaced` became
`@  spaced`, which this tool then reports as `invalid-xpath-expression`, and
`child::  gapped` became `  gapped`, which `redundant-whitespace` flags. So the
fix was trading one warning for an error plus a warning. All four rows check out
on the branch, and so do `child ::  title`, `attribute::  *` and
`self ::  node ( )`.

**#627 is right.** I counted the five positions in the new pack by hand; each
lands on the character it names.

One thing keeps the flag off, one is new here, three are nits.

### Abbreviating `self::node()` inside a pattern changes what the stylesheet does

`.` and `self::node()` select the same nodes in an expression, but a `match` is
not an expression, and as patterns the two do not carry the same default
priority — Saxon puts `self::node()` at -0.5 and `.` at -1. So `--fix` on a
pattern silently demotes the template, and no explicit priority is needed to see
it:

```xsl
<xsl:template match="node()">NODE</xsl:template>
<xsl:template match="self::node()">SELF</xsl:template>
```

Before `--fix` that prints `SELF` — both are -0.5 and the later one wins. After
it, `match="."` sits at -1 and the same input on the same processor prints
`NODE`.

Three more shapes stop compiling outright, since a `.` pattern may only stand at
the outermost level:

```text
match="y|self::node()"      ->  match="y|."      A PredicatePattern can appear only at the outermost level
match="(self::node())"      ->  match="(.)"      same
match="y|self::node()[1]"   ->  match="y|.[1]"   same
```

All three compile before the fix and not after, and all three go through in one
`--fix` run on a 3.0 sheet. `parent::node()` escapes only by accident — no
reverse axis is a legal pattern, so that input was already broken — and the two
abbreviations that drop a step, `child::` and `attribute::`, are safe, because a
pattern takes both axes.

This predates the branch, so it is issue material rather than something to fold
in, but it does keep `mature` off `unabbreviated-axis`: the bar asks for a fix
"correctly tiered", and this one sits in the safe tier while it changes
behaviour and can break a build. The narrowest correction is to withhold `.`
where the expression came from a pattern attribute — `match`, `count`, `from`,
`group-starting-with`, `group-ending-with` — which `attributes.js` does not
currently tell a linter apart from the rest. The motive wants a line too: it
teaches the abbreviations as selecting "precisely the same nodes as the
longhand", which holds for a `select` and not for a `match`.

### A gap that crosses a line wrap is no longer fixable at all

The gap is part of the fix value now, and as soon as that value spans a wrap
#629 catches it:

```xsl
<xsl:value-of select="child::
                      alpha"/>
```

```text
[WARNING] Skipped fixing unabbreviated-axis at wrapgap.xsl:4, the source no longer matches
```

Master fixes that one, badly, leaving the wrap standing where the axis was. The
branch declines it, blames a file nothing has touched, and the defect survives
every rerun. I would still rather have the decline than master's output, so it
is no reason to hold the PR — but the fixture pair does not cover it: both new
rows in `test/resources/fix/unabbreviated-axis.xsl` are inline gaps. Worth
either pinning the decline or holding the gap back when it carries a newline,
which is master's behaviour for exactly the inputs #629 blocks.

### Nits

`test/resources/axis-packs/abbreviated-axis.yaml` declares
`pack: abbreviated-axis`, a name no check has. Nothing catches it: the
`names no check` assertion in `conformance.test.js` walks `PACKED`, which is
`xpath-packs` and `corpus-packs` only, while `gives every format check a test
pack somewhere` reads that same field across every directory — so the field is
load-bearing for a format check and unenforced for one.

I count 20 pack files in `axis-packs`, 19 of them naming the check, and 16 in
`using-namespace-axis-packs` — not the 22 and 12 above.

The `SHORT` docblock still says the verbatim text "comes from the token, so a
spaced `child ::` is stripped in full rather than leaving its whitespace
behind". Half of that moved into `spans`: the token carries the gap in front of
the colons and `spans` carries the one behind it.

maxonfjvipon (commented):
Checked at `ad6bf7f`, re-running each claim rather than reading the diff for it.
721 tests here plus 100% branches (916/916); CI 18/18. The `968` in the
description is right and mine is not — xcop is not installed here, and the 247
cases it registers over the 248 pack inputs make up the difference exactly. No
correction needed there.

**The #615 half is right,** and it does more than the table says. Running the
fixture through the CLI before and after:

```text
unabbreviated-axis.xsl        3 invalid-xpath-expression + 8 unabbreviated-axis
unabbreviated-axis.fixed.xsl  0 + 0
```

So `--fix` now clears the *error* as well, on every whitespaced spelling —
`attribute::  spaced`, `child::  gapped` and the older `parent :: node ( )`.
Until fontoxpath takes the other half of #615, this branch is the only thing a
user has that makes those expressions parse. Worth saying in the issue.

**#627 is right.** Five positions in the new pack, each on the character it
names, and all five fixed in one run.

One correction to the review above, two things I did not see raised, and
agreement on the rest.

### The compile failures are real; the priority mechanism is not

The three shapes that stop compiling reproduce on both processors I have, so
that half needs no defence:

```text
                    Saxon-HE 12.9              SaxonJS 2.6
y|.                 XTSE0340                   XTSE0340
y|.[1]              XTSE0340                   XTSE0340
(.)                 java.lang.AssertionError    "Predicate pattern cannot be parenthesized"
```

`y|self::node()`, `(self::node())` and `y|self::node()[1]` all compile on both.
Saxon-HE crashing on `(.)` rather than diagnosing it is Saxonica's problem, but
it makes the point sharper: `--fix` hands the processor something it cannot
even reject cleanly.

The priority half does not hold up. The exact pair in the review prints `SELF`
both before and after, on both processors — the predicted `NODE` never appears:

```text
                          before (self::node())   after (.)
Saxon-HE 12.9             SELF                    SELF
SaxonJS 2.6               SELF                    SELF
```

Bisecting against an explicit priority says why. `self::node()` crosses over at
exactly -0.5, as claimed. `.` does not cross over at all:

```text
match="self::node()"  vs  match="node()" priority=-0.75  ->  self wins
match="self::node()"  vs  match="node()" priority=-0.5   ->  node wins   (-0.5 confirmed)
match="."             vs  match="node()" priority=1000   ->  dot wins
match="." priority=-3 vs  match="node()"                 ->  dot wins
```

A predicate rule is not being resolved by priority at all on either processor;
it is consulted ahead of the kind chains whatever number is on it. SaxonJS's
own SEF does export `prio: "-1"` for `match="."`, which is presumably where the
-1 came from, but that number does not decide the outcome, and nothing I could
run demotes the template.

Conflict resolution *does* change, just not that way and not consistently.
Declaring `self::node()` before `*` and applying to an element:

```text
                 before      after
SaxonJS 2.6      SELF        STAR
Saxon-HE 12.9    STAR        STAR
```

That is one processor changing its answer and one not — which is a worse
property than a clean demotion, and it is the honest version of the finding.
The conclusion is untouched: the rewrite is not semantics-preserving inside a
pattern and cannot sit in the safe tier. But the issue should lead with the
three static errors and drop the -1 arithmetic, or it will be cited later and
not reproduce.

Worth filing it beside #583 and the last bullet of #499 — `starts-with-double-slash`
is the same defect, a safe-tier fix editing a pattern attribute and moving the
rule with it. Two instances make it a class, and the remedy is shared:
`attributes.js` knows which names are patterns and does not currently tell a
linter, so no fixer can withhold on that basis even when it wants to.

### Nothing in `axis-packs` pins what this branch changed

The pack harness asserts `defects[index].fix.replacement` and nothing else. The
replacement is exactly what did *not* change here — `""` and `"@"` before and
after. `fix.value`, the span the gap was added to, is asserted by no pack at
all:

```js
assert.equal(defects[index].fix ? defects[index].fix.replacement : null, expected)
```

Revert `spans` to `token.value` and all 20 packs still pass; only the
`.fixed.xsl` pair catches it, through two inline rows. For a check being frozen
that is the one dimension of its fix left unpinned, and #607 is already circling
the same field. A `values:` list beside `fixes:` would close it, and would pin
the wrapped-gap decline too — that case has no home in the fixture pair, as the
review above says, but it does have one in a pack.

### `ATTRIBUTES` is still one short

`use-when` holds an XPath expression on every XSLT element, and `xsl:use-when`
on every literal result element. Neither is read:

```text
<xsl:template match="/" use-when="child::x"/>     ->  nothing
<out xsl:use-when="child::x"/>                    ->  nothing
```

For these two checks it costs nothing — `use-when` is evaluated with no context
item, so an axis step there is an error anyway. It costs the others: the
function-only expressions `use-when` really carries are exactly what
`count-compared-to-zero`, `string-length-compared-to-zero` and
`redundant-boolean-call` are looking for. And the literal-result spelling cannot
be reached through `selectorOf` at all — `//xsl:*/@name` cannot see an attribute
whose owner is not an XSLT element, so it needs the `getAttributeNS(XSLT, ...)`
that `expands` and `shadow` already do. Same shape as #627, wider surface, and
it belongs in whatever closes that one rather than being found again in six
months.

### Smaller

`select="child::  "` becomes `select=""`. The validator rejects both, and master
left `select="  "`, so nothing regresses — but it is the one input where eating
the gap is not plainly an improvement, since an empty `select` is a different
error from the one the user wrote.

Agreed on the `SHORT` docblock: the value has not come from the token since
`spans` landed, and the sentence claiming it does is now the opposite of true
for the gap behind the colons.

Agreed on `abbreviated-axis.yaml`, and on the counts — I get 20 files in
`axis-packs`, 19 declaring the check, and 16 in `using-namespace-axis-packs`.

### On the two flags

Split them.

`using-namespace-axis` I could not break. Every version, both roots, simplified,
non-`xsl` prefix, a raised and a lowered subtree, buried three times in one
expression, and the four lookalikes — all correct. Report-only is by nature
here, not deferred: the motive itself shows the rewrite depends on what the
namespace node was read for, so no deterministic fix exists to be written. That
flag has earned its place.

`unabbreviated-axis` has to wait for the pattern question, and I would not treat
the fix as merely mis-tiered. Demoting it to `suggestion: true` still emits
`match="y|."` under `--fix-suggestions`, and that stylesheet does not compile —
a suggestion is allowed to change behaviour, not to break the build. The narrow
answer in the review is the right one: withhold `.` where the expression came
from a pattern attribute, full stop.

The code is ready. I would land it with the two `mature:` lines dropped, file
the pattern defect against #583's class, and bring the flags back in their own
commit once that is settled — which is also what keeps the freeze meaning
something.

maxonfjvipon:
The pattern defect is fixed here rather than filed, because the narrow answer
you both landed on turned out to be small and it is the only thing keeping the
flag off.

`src/attributes.js` now says which of the names it reads hold a pattern —
nothing could ask it before, which is why no fixer could withhold on that basis
— and the axis linter withholds `.` and `..` there:

```text
match="y|self::node()"       unchanged
match="(self::node())"       unchanged
match="y|self::node()[1]"    unchanged
match="self::node()"         unchanged
match="R/child::qwer"        R/qwer     still fixed
match="attribute::id"        @id        still fixed
```

The two that drop a step keep their fixes, both being axes a pattern takes. I
took the grammar rather than the priority argument: §5.5.2 makes `.` a
`PredicatePattern`, an alternative to the whole union grammar rather than a term
inside it, so `y|.` and `(.)` are syntax errors by construction — which is the
version that reproduces without a processor to hand, and matches the correction
in the second review. The motive says it too now; it was teaching "precisely the
same nodes as the longhand", true of a `select` and not of a `match`.

### The rest

**`values:`** — you were right that nothing pinned what this branch changed.
Reverting `spans` left all 20 packs green. There is a `values:` list beside
`fixes:` in both axis harnesses now, and a pack that exercises a gap after the
colons, a tab, and both gaps at once.

**The pack naming no check** — fixed, and the gate extended to every
`*-packs` directory. It found nine more: two in `node-set-packs`, two in
`xpath-format-packs`, five in `xpath-validator-packs`. All renamed to the check
that owns them, so the coverage gate can see them.

**`use-when`** — added. As you say it costs these two checks nothing and the
function checks the whole attribute. The literal-result spelling still needs the
`getAttributeNS` path, which is not reachable through `selectorOf`; that is left
for #627's successor rather than bolted on here.

**The wrapped gap** — I tried holding the gap back when it carries a newline and
it cannot work: attribute-value normalisation turns the wrap into a space before
the lexer sees it, so there is no newline in the token to test. The decline is
pinned in the fixture pair instead, and it clears itself when #629 lands.

**Nits** — the `SHORT` docblock says what `spans` does now. Your counts are
right and mine were not: 21 files in `axis-packs` after the new one, 16 in
`using-namespace-axis-packs`.

### The flags

Both are still set, which is me disagreeing with the recommendation to split —
the reason to hold `unabbreviated-axis` was the pattern defect, and it is closed
above rather than deferred. If the withholding is wrong or incomplete, that is
the thing to shoot at.

970 tests, 100% branches (916/916).

maxonfjvipon:
Re-reviewed at 5fd4dee, running each claim rather than reading the commits for
it. 722 tests and the coverage gate pass here, 100% of 916 branches, CI 18/18.

**The withholding is right.** Every shape through the CLI:

```text
match="y|self::node()"      unchanged
match="(self::node())"      unchanged
_match="y|self::node()"     unchanged      <- the shadow spelling too
count="self::node()"        unchanged
from="parent::node()"       unchanged
select="self::node()"       .              still fixed
match="R/child::qwer"       R/qwer         still fixed
match="attribute::id"       @id            still fixed
```

All eight defects are still reported and only the fix is withheld. `PATTERNS`
covers what a pattern attribute can be in 3.0, and stripping the leading `_` in
`carried` catches the shadow spelling, which is the case I expected to slip.

**`use-when` is clean.** Read on an XSLT element, left alone on a literal result
element's plain `use-when` — result-tree text, not the shadow spelling — and
`system-property(...)` and a static `$param` produce nothing spurious.

**The `values:` list does its job.** Reverting `spans` to `token.value` now
fails four tests where it failed none.

Two things.

### Nothing in the suite reaches the withholding

The same hole you closed for `spans`, one commit later. Deleting `pattern ||`
from the ternary leaves all 722 tests green, and it is not an assertion being
too weak — no test reaches that line holding a pattern at all:

```js
} else if (step !== null) {
  if (pattern) { throw new Error('PATTERN-STEP-REACHED') }   // never thrown, 722 passing
```

No fixture in the tree pairs a `match`, `count`, `from` or grouping boundary
with a `self::` or `parent::` step. c8 puts the file at 100% of branches either
way, because V8 block coverage does not split the operands of
`pattern || (step.predicated && !modern)` inside a ternary condition, so the
gate will not catch this one for you. A pack with the four withheld shapes and
`fixes: [ null, null, null, null ]` closes it, and that is what I would fold in
before merging — the flag now rests on that line.

### The priority change does reproduce

Worth correcting, since the recommendation was to drop it from the issue. The
exact pair, on every Saxon-HE here — 12.3, 12.4, 12.5, 12.9 and 13.0:

```text
<xsl:template match="node()">NODE</xsl:template>          before   SELF
<xsl:template match="self::node()">SELF</xsl:template>    after    NODE
```

12.9 shows the mechanism. Before the fix it warns
`XTDE0540 Ambiguous rule match ... Matches both "node()" ... and "node()"` — a
tie at -0.5 broken by declaration order. After, the warning is gone and `node()`
wins outright, because `.` no longer sits at -0.5. Bisecting against an explicit
priority on the same processor gives both numbers, and shows priority being
honoured rather than bypassed:

```text
match="."             vs node() priority=-1.01   -> dot      -1   confirmed
match="."             vs node() priority=-0.99   -> node
match="self::node()"  vs node() priority=-0.51   -> self     -0.5 confirmed
match="self::node()"  vs node() priority=-0.49   -> node
match="." priority=-3 vs node()                  -> node     (not "dot wins")
```

I cannot run SaxonJS, so a divergence there stands. But on Saxon-HE it
reproduces, and it decides what the issue should say: the three static errors
need a union or a parenthesis, while a bare `match="self::node()"` — the
ordinary spelling, and one this branch also withholds — is covered by the
priority change alone. An issue that leads with the errors and drops the
priority argues for a narrower withholding than the one you shipped.

### Smaller

`CLAUDE.md` still describes the old `expressionsOf` contract in both places it
names it, the key-files row and the "Adding a rule" paragraph. Neither says an
expression now reports whether it is a pattern, or that a fixer has to withhold
`.` and `..` there. That is the general remedy for the #583 class, and the file
is where the next linter's author goes looking for it.

### The flags

Agreed on keeping `using-namespace-axis` marked — I could not break it either,
and report-only is by nature there rather than deferred. `unabbreviated-axis` I
would still hold, but now only for the pack above: the defect is closed, and a
freeze resting on an untested line is the argument the `values:` list was added
to answer.

maxonfjvipon (commented):
Re-checked at `5fd4dee`. 722 tests, 100% of 916 branches, CI 18/18.

First, the correction I owe.

### The priority change is real and my objection was an artifact

You are right on both numbers, and the demotion happens. Bisecting against an
explicit priority, with the two rules actually competing for the same node:

```text
Saxon-HE 12.9 and 13.0, both patterns applied to element r
  match="."             vs match="*" priority=-1.01   -> dot     -1   confirmed
  match="."             vs match="*" priority=-0.99   -> star
  match="self::node()"  vs match="*" priority=-0.51   -> self    -0.5 confirmed
  match="self::node()"  vs match="*" priority=-0.49   -> star
```

My earlier run never put them in a contest. `match="."` matches the document
node; `match="node()"` and `match="self::node()"` do not — Saxon normalises both
to `node()`, which is `child::node()`, and nothing is a child of the document
node:

```text
match="."             -> matched: document-node
match="self::node()"  -> matched: r
match="node()"        -> matched: r
```

So the `.` rule captured the root of the tree before the traversal ever reached
`r`, and it kept doing that at `priority="-3"` because there was no competitor
there at all. That is what I read as "not resolved by priority", and it is not.
Withdrawn.

It also explains the one divergence left between us. Your exact snippet still
prints `SELF` both before and after here, on 12.3, 12.4, 12.5, 12.9 and 13.0 —
after the rewrite the `.` template takes the document node and prints `SELF`
there, never reaching the element the ambiguity was about. The finding is right;
that particular pair just hides it behind the document node. Adding a driver —
`<xsl:template match="/"><xsl:apply-templates select="r"/></xsl:template>` —
makes it print `NODE` after the fix and `SELF` before, on every version above.
Worth using that shape in the issue, since the confounder cost one review round
already.

Which also settles the scope question you raised: the priority change does cover
a bare `match="self::node()"`, so withholding it is right and not over-broad.

### The withholding is right

Every shape through the CLI, and the two that must survive:

```text
match="y|self::node()"                    reported, no fix
match="(self::node())"                    reported, no fix
match="self::node()"                      reported, no fix
_match="y|self::node()"                   reported, no fix
count="self::node()" from="parent::node()"  both reported, no fix
group-starting-with="self::node()"        reported, no fix
select="self::node()"                     ->  .
match="R/child::qwer"                     ->  R/qwer
match="attribute::id"                     ->  @id
```

`PATTERNS` is the right five: no XSLT element spells a non-pattern `match`,
`count`, `from` or grouping boundary, so the list cannot over-reach. The braced
branch hard-coding `pattern: false` is right too — a shadow `_match="{...}"`
holds static expressions whose *result* is the pattern, not a pattern itself.

`use-when` reads on an XSLT element and is left alone on a literal result
element, where a plain `use-when` really is result-tree text.

### Your two points, both confirmed

**Nothing reaches the withholding.** Deleting `pattern ||` leaves 722 tests
passing, and the coverage gate still reports 100% of 916 branches — so c8 will
not catch it for you either, exactly as you say. The pack closes it: with
`fixes: [ null, ... ]` the harness compares `fix.replacement` against `null`, so
a restored fix fails the row.

**`CLAUDE.md`** is still stale in all three places that name the contract —
lines 114 and 178 describe what `expressionsOf` hands a linter, and the
key-files row at 365. None mentions that an expression now says whether it is a
pattern, or that a fixer must withhold `.` and `..` there. That last part is the
general remedy for the #583 class and the file is where it belongs.

### One thing neither of us has raised

The check still *reports* the shapes it now refuses to fix, with the message it
had before:

```text
match="self::node()"   (3:24) Verbose axis specifiers ... are used.
                              Replace them with the node name alone, @, .., or . respectively.
```

The motive, one commit later, says the opposite: "Leave a longhand step alone in
a `match`, a `count`, a `from`, or an `xsl:for-each-group` boundary." So the
warning tells the reader to make an edit the motive tells them not to make, and
for `match="y|self::node()"` that edit does not compile. A defect whose message
contradicts its own motive is a false positive under the bar, not a fix-tier
question.

The check's own design already answers it. `abbreviable` reports a step only
where a shorter form exists — `parent::n` and `self::text()` sit in the fixture
untouched and unreported, because `afterNode` returns null for them. In a
pattern, `.` and `..` are not available either, so `self::node()` there is the
same case: no shorter form, no defect. Skipping the `push` when `pattern` is
true, rather than only blanking the `fix`, keeps the check saying one thing.

It also makes the pack you are about to write stronger: `amount: 0` on four
pattern shapes cannot pass with `pattern ||` deleted *or* with the report left
in, whereas `fixes: [ null, ... ]` only catches the first.

### Smaller

The motive's new paragraph opens "A pattern takes the child and attribute axes".
That is narrower than the truth, and the sentence is doing load-bearing work —
a reader takes it to mean `match="self::x"` is illegal. It is not:

```text
match="self::r"                 HIT
match="descendant::c"           HIT
match="descendant-or-self::c"   HIT
```

on Saxon-HE 12.9 under both `version="2.0"` and `version="3.0"`. What makes the
two abbreviations safe is not that a pattern takes only those axes, but that it
takes *those two* — worth saying that way, since #552 is about motives that
overstate.

### The flags

`using-namespace-axis` — agreed, keep it.

`unabbreviated-axis` — with the pack, and with the report suppressed rather than
just the fix, I would let it stand too. The defect is closed rather than
deferred, the withholding is correct, and the two rewrites that survive in a
pattern are the two that provably do not move a rule.

That leaves the pack, the reporting question, and `CLAUDE.md` between this and
merge. Everything else I could throw at it holds.

maxonfjvipon:
Taking the reporting point — it goes further than my fix did, and it is right.

Withholding the fix left the check saying two things at once: the warning still
told the reader to write `.` where the motive, one commit later, told them not
to, and for `match="y|self::node()"` that edit does not compile. The check
already answers this shape, as you say — `parent::n` and `self::text()` are
unreported because nothing shorter exists for them — and a pattern is the same
case. So it is no longer reported there at all:

```text
match="self::node()"                          0 defects
match="y|self::node()"                        0
match="(self::node())"                        0
match="y|self::node()[1]"                     0
match="parent::node()"                        0
_match="y|self::node()"                       0
count="self::node()" from="parent::node()"    0
group-starting-with="self::node()"            0
match="R/child::qwer"                         1 -> R/qwer
match="attribute::id"                         1 -> @id
select="self::node()"                         1 -> .
```

`abbreviated-step-inside-a-pattern.yaml` covers all eight withheld shapes at
`amount: 0`, and it fails on either mistake — restoring the fix or restoring the
report. I checked it catches the revert rather than assuming: deleting the guard
takes the file from 22 passing to 21 and 1 failing.

Your point about c8 is the part worth keeping in view. V8 block coverage does
not split the operands of a ternary condition, so the gate reported 100% of 916
branches with the guard deleted. That is the second time on this branch a
100%-branches figure has been consistent with an untested line, and both times
it took someone deleting the line to find out.

### The correction

Noted, and it changes what the issue should say rather than what the code does.
The confounder is a good one to have written down: `match="."` matches the
document node while `match="node()"` and `match="self::node()"` do not, so the
rewritten rule fires above the element the contest was about and the contest
never happens. With a driver template the demotion shows on every Saxon here.

It also settles the scope question in my favour rather than yours, which I would
not have argued for on my own: the priority change reaches a bare
`match="self::node()"`, so withholding the plain spelling is not over-broad.

### The rest

`CLAUDE.md` carries the contract in all three places now — that an expression
says whether it is a pattern, that `PATTERNS` names the five, and that a fixer
withholds a step there. It is the general remedy for the class, so it belongs in
the file rather than only in this linter.

The motive's "a pattern takes the child and attribute axes" is fixed. You are
right that it read as though `match="self::x"` were illegal; the point is that a
pattern takes those two *among others*, which is what makes the step-dropping
abbreviations safe there and says nothing about the rest.

972 tests, 100% branches (917/917). Both flags still set.

maxonfjvipon:
Re-reviewed at 8dd2489. 723 tests, 100% of 917 branches, CI 18/18.

**Not reporting is the right call, and it holds.** All eleven rows reproduce
through the CLI — eight pattern shapes silent, and the three that must survive:

```text
match="R/child::qwer"     ->  match="R/qwer"
select="attribute::id"    ->  select="@id"
select="self::node()"     ->  select="."
```

**The pack pins it.** Deleting `&& !pattern` takes the suite from 723 passing to
722 and 1 failing, so the revert is caught rather than assumed. c8 still reports
100% of branches with the guard gone, which is worth keeping in mind for the next
one of these — the pack is the only thing standing there, not the gate.

`CLAUDE.md` carries the contract in all three places now, and the general rule
rather than this linter's version of it, which is what the class needed.

One thing left, and it is one sentence.

### The corrected motive sentence is right for 3.0 and wrong for 1.0 and 2.0

> A pattern is free with axes — `match="self::r"` and `match="descendant::c"` are
> both legal

Both of its examples fail to compile on a strict 1.0 processor:

```text
xsltproc, match=
  child::c                 HIT
  @a                       compiles
  self::r                  failed to compile 'self::r'
  descendant::c            failed to compile 'descendant::c'
  descendant-or-self::c    failed to compile 'descendant-or-self::c'
  self::node()             failed to compile 'self::node()'
```

1.0's `ChildOrAttributeAxisSpecifier` and 2.0's `PatternAxis` both admit exactly
`child::`, `attribute::` and `@`; it is 3.0's `ForwardAxisP` that widens the step
to `descendant`, `self`, `descendant-or-self` and `namespace`. Saxon reports HIT
under `version="2.0"` because it applies the 3.0 grammar whatever the sheet
declares, so it cannot tell the two apart — which is presumably where the "both
legal" reading came from.

The awkward part is that the sentence it replaced was *true*, just for a reason
it did not give: before 3.0 a pattern takes only the child and attribute axes,
and those are exactly the two whose abbreviations this check still applies inside
a pattern. That is not a coincidence to write around, it is the argument for why
the two surviving rewrites are safe in every version, which the paragraph now has
no way to make. `match="y|self::node()"` a few lines down is 3.0-only for the
same reason — legal there, not a pattern at all before it.

Worth carrying the version, then, rather than choosing a side: a pre-3.0 pattern
takes those two axes and no others, 3.0 widened it, and either way those two are
the ones the abbreviations touch. Line 63 also runs to 90 characters where the
rest of the file wraps at 80 — under the 100 the config allows, but the seam
shows.

### Two for later, neither in this PR's way

The silence now reaches `match="descendant::c"` in a 2.0 sheet, which no
conforming 2.0 processor accepts and nothing here reports. A pattern using an
axis its declared version forbids is a check in its own right, and this branch is
what makes it cheap — `expressionsOf` says which expressions are patterns and
`versionOf` says which version is in force, which is the whole of what it needs.

`using-namespace-axis` fires on `match="namespace::*"`, legal in 3.0, with a
message naming `in-scope-prefixes()` and `namespace-uri-for-prefix()` — neither
of which can stand as a pattern step. Milder than the case just fixed, since the
deprecation it reports is true wherever it fires and the check offers no fix to
misapply, and the motive does send the reader to `xsl:namespace` rather than
leaving them stuck. Still the same shape: advice that does not apply where it is
given.

### The flags

Both, now. `unabbreviated-axis` has the defect closed rather than deferred, the
withholding is provably right, the revert is caught by a pack, and the two
rewrites that survive in a pattern are the two that provably cannot move a rule.
`using-namespace-axis` I could not break across three rounds. Fix the sentence
and this is done.

maxonfjvipon (commented):
Re-checked at `8dd2489`. 723 tests, CI 18/18.

### You are right about the sentence, and my nit was version-blind

I tested "a pattern is free with axes" on Saxon alone, which applies the 3.0
pattern grammar whatever the sheet declares — the confounder I had just spent
two paragraphs warning about, one round later. `xsltproc` settles it:

```text
libxslt, match=
  child::c                 HIT
  @a                       compiles
  self::r                  failed to compile 'self::r'
  descendant::c            failed to compile 'descendant::c'
  descendant-or-self::c    failed to compile 'descendant-or-self::c'
  self::node()             failed to compile 'self::node()'
  namespace::*             failed to compile 'namespace::*'
```

So the sentence I asked you to widen was true before 3.0, and the replacement is
false there. Carrying the version is the right resolution, and the paragraph gets
a better argument out of it than either of us had: it is not a coincidence that
the two axes a pre-3.0 pattern admits are the two whose abbreviations survive
inside one.

That argument now has evidence for every version rather than for 3.0. The two
surviving rewrites cannot move a rule, tested by declaration order — a tie means
whichever is declared second wins, and both spellings tie in both directions:

```text
                            xsltproc (1.0)   Saxon-HE 12.9 (3.0)
child::c  then  c              second              second
c  then  child::c              second              second
attribute::a  then  @a         second              second
@a  then  attribute::a         second              second
```

Equal priority on two independent processors, at both ends of the version range.
That is the claim the freeze rests on, and it holds.

### `8dd2489` checks out

Deleting `&& !pattern` takes it from 723 passing to 722 and 1 failing, so the
pack catches the revert. Coverage still reports 100% of branches with the guard
gone — third time on this branch, and worth a line somewhere that a
100%-branches figure says nothing about a ternary's operands.

Beyond the eleven rows, the ones I went looking for:

```text
match="self::node()"                     silent on 1.0 and on 3.0
xsl:accumulator-rule/@match="self::node()"   silent
xsl:key/@match="self::node()"                silent
xsl:key/@use="self::node()"                  ->  .        still fixed
match="child::a" on a 1.0 sheet              ->  a        still fixed
```

`@use` beside a silent `@match` on the same element is the case I most expected
to leak, and it does not.

### The same sentence, one file over

`abbreviable`'s docblock now carries the 3.0-false half of what the motive is
about to lose:

> A pattern has no abbreviated step to offer at all — `.` is a pattern in its own
> right rather than a step inside one — so a longhand step there goes unreported,
> the way `parent::n` does: both are steps with nothing shorter to become.

True before 3.0. In 3.0 there *is* something shorter — `match="."` compiles, as
the priority work established — it simply is not a synonym, sitting at -1 where
the longhand sits at -0.5. So the reason to stay silent on 3.0 is that the short
form changes which template wins, not that none exists, and `parent::n` is only
half a parallel: nothing shorter exists there in any version.

Worth fixing in the same pass, since it is the same sentence's logic and the
docblock is where the next reader of this linter looks for why the branch is
there.

### On deferring the `namespace::` message

Your call, and I would not hold the PR for it. One thing to weigh, though: the
flag freezes the message along with the rest, so moving it later costs an
unfreeze — which is the churn `mature` exists to prevent, spent on something both
of us can already see. And it is a little worse than the summary suggests, since
by the same grammar you cite `match="namespace::*"` is not a legal pattern before
3.0 either — `xsltproc` refuses it above. On a 2.0 sheet the check reports a
deprecation, correctly, on a pattern no conforming 2.0 processor will compile,
and points at two functions that cannot stand there.

Settling it now costs a sentence in the YAML; settling it later costs taking the
flag off. Either is defensible — it is the sequencing I would think about rather
than the finding.

### Verdict

The motive sentence, the docblock sentence beside it, and this is done. Both
flags stand as far as I can tell: four rounds, and what I have left to say about
either check is prose.

maxonfjvipon:
Both sentences fixed in `65a1c7a`, and the version-blind one was mine. I tested
it on Saxon alone — the exact confounder the round before had just documented:

```text
xsltproc (XSLT 1.0), match=
  child::c              compiles        self::r                REJECTED
  @a                    compiles        descendant::c          REJECTED
                                        self::node()           REJECTED
```

So the sentence I replaced was true and my replacement was not. Carrying the
version gives the paragraph the argument neither wording had, which is the part
worth keeping: the two axes a pre-3.0 pattern admits are exactly the two whose
abbreviations are safe inside one, and they stay safe in 3.0 because they drop a
step without touching how the rule is weighed.

`abbreviable`'s docblock had the same half-truth — "no abbreviated step to offer
at all" — and now says the reason that holds in every version: from 3.0
`match="."` compiles, it is simply not a synonym, so there is no shorter form to
*recommend* rather than none to exist. `parent::n` is dropped as a parallel,
since nothing shorter exists there in any version.

Line 65 wraps at 80 with the rest of the file.

### The equal-priority evidence

Worth having in the thread: the two surviving rewrites cannot move a rule, on
both processors at both ends of the range, tested by declaration order — a tie
means the later declaration wins, and both spellings tie in both directions. That
is the claim the freeze actually rests on, and it is now checked rather than
argued.

### The sequencing question

Filed rather than settled, and here is the reasoning rather than a shrug. #632.
Hedging a one-sentence message for a rare context makes it read worse in the
common one, and the honest fix for `match="namespace::*"` is not a softer message
but the version-scoped pattern check — on a 2.0 sheet that pattern does not
compile at all, which is the real defect and is nobody's message. #631 carries
that, and it is cheap now precisely because this branch made `pattern` and
`versionOf` both available at the node.

So yes, acting on #632 later costs an unfreeze. I would rather pay that
deliberately, with the reason written down, than freeze a message I softened to
avoid it.

973 tests, 100% branches (917/917). Both flags set.

maxonfjvipon:
Green again at `52eed22`, 18/18.

Two things happened after the last comment, and the first is a process failure
worth stating plainly.

**A red run read as green.** The docblock sentence I reflowed grew one line past
80, and I committed and pushed anyway, because I run the gates as
`npx grunt 2>&1 | gsed ... | head -4 && git commit`. A pipeline reports the exit
status of its last command, so `head` returning 0 masked grunt returning 3, and
the `&&` let the commit through. That is the third time on this branch I have
mistaken a filtered view of a gate for the gate; the previous two were greps that
did not match a failure line. I am running them unpiped and reading the exit code
now.

**Coverage exits 1 on this machine, and did before the branch.**
`should test default directory` times out at 10s under instrumentation — it lints
the whole repository — and master fails the identical test the same way here, so
it is local timing rather than anything this branch did. CI's coverage job is
green on every commit above.

Chasing it did turn up a real improvement, which is in: positioning walked the
raw text one character at a time, and nothing in a span can be wider than one
character unless the span holds an `&` or a `\r`. Where it holds neither the
answer is the count itself, so the walk is skipped. The CRLF, CR-only and
entity-shifted cases still answer exactly what they answered before:

```text
lf 8:30    crlf 8:30    cr-only 5:27
```

972 tests, 100% of 5228 statements and 919 branches. The two sentences from your
last round are the commit before this one; nothing else is outstanding that I
know of.

maxonfjvipon:
Re-reviewed at 52eed22. 723 tests here, CI 18/18.

**The version-carrying rewrite is right.** Every claim in the new paragraph
checks out — the pre-3.0 axis pair, a 1.0 processor refusing `match="self::r"`,
3.0 widening the step, and the two surviving abbreviations leaving the weighing
alone. The argument it now makes is the one the check rests on, which neither
earlier wording had.

**The `skip` change is correct, and it is real.** I ran the two implementations
against each other over 200,000 generated spans — `&`, `\r`, `\r\n`, an
unterminated `&`, an unknown entity, offsets and counts past the end — with no
mismatch, and the reasoning closes in the direction that matters: `character`
advances by more than one only at an `&` or a `\r`, so a span holding neither is
one-for-one and the induction carries. On a real stylesheet it is 62.9 ms to
6.6 ms over 80,000 calls. Both operands and the fast path are pinned by the
`WALKS` rows in `test/source.test.js`, so this one is not resting on a coverage
figure.

Two things.

### Both new sentences have the priority the wrong way round

> and standing alone it outranks the longhand it replaced, quietly handing the
> node to a template that used to lose

and the docblock beside it — "illegal in a union and outranking the longhand
where it stands alone".

`.` sits at -1 and `self::node()` at -0.5, so the longhand outranks the
abbreviation rather than the reverse. Order-independent, on Saxon-HE 12.9:

```text
self::node() first, .            second  ->  self::node() wins
.            first, self::node() second  ->  self::node() wins
node()       first, .            second  ->  node() wins
.            first, node()       second  ->  node() wins
self::node() vs node()                   ->  tie, declaration order decides
```

The rewrite *demotes* the template, which is what makes the second clause right:
the node does go to a template that used to lose, because the one that used to
win got weaker. Flipping "outranks" to "is outranked by" fixes both sentences,
and nothing else in either needs touching.

Worth flipping rather than cutting, because the direction is the reason the
withholding is not over-broad. A reader who takes the abbreviation for a
promotion reaches for it to *win* a conflict and gets the opposite — and this is
the fact the last two rounds turned on, so it is the one sentence in the motive
that should not be loose.

### Smaller

"`match="c"`, `match="child::c"` and `match="@a"` are the whole of it" reads as
three exhaustive forms, where `match="*"`, `match="text()"` and `match="a/b"` are
legal 1.0 patterns too. The antecedent is the axes and the paragraph is about
axes, so it is only a reading — but it is the reading #552 exists for, and the
clause before it already does the work.

The `skip` commit is a `bug(#611)` on a `bug(#615)` branch, and the description
does not mention it. Not worth splitting out now that it is measured and pinned,
but the description is what the next reader of this branch has, and a change to
how every defect is positioned is not something to find only in the log.

### The flags

Both, still. Flip the two sentences and I have nothing left.

maxonfjvipon (commented):
Re-checked at `52eed22`. 723 tests, 100% of 5228 statements and 919 branches,
`npx grunt` unpiped exits 0, CI 18/18. Coverage passes here too — the timeout you
saw did not reproduce, which fits it being local timing rather than the branch.

**Both flips are right, and order-independent.** Competing at an element, so
neither spelling can capture the document node ahead of the contest:

```text
self::node() first, .            second  ->  self::node()
.            first, self::node() second  ->  self::node()
node()       first, .            second  ->  node()
.            first, node()       second  ->  node()
self::node() first, node()       second  ->  node()        tie, order decides
node()       first, self::node() second  ->  self::node()  tie, order decides
```

The longhand wins whichever way round it is declared. "is outranked by" in both
places, and the clause after it stands as written.

**The `skip` change is sound.** I ran my own differential against the walk rather
than trusting the reasoning — 11,223,074 spans over generated texts built from
`&amp;`, `&lt;`, `&nope;`, a bare `&`, `&#65;`, `\r`, `\r\n`, `\n`, a bare `;`,
and multi-byte characters, at every offset and every count including past the
end. No mismatch.

The induction closes for a reason worth writing in the docblock, because it is
what makes the window the right one: everything before the first wide character
is one-for-one, so the first `&` or `\r` the walk meets is always at an offset
below `at + count` — inside the slice being tested. There is no way for a wide
character to hide beyond the window and still be reached.

One edge the window does change. A negative `count` answers `at + count` on the
fast path and `at` on the walk, since the loop simply does not run:

```text
skip('abcdef', 3, -2)   fast = 1   walk = 3
```

Unreachable today — `checks.js` is the only caller and its `offset` is a position
inside a value — so this is not a defect, but it is a shared helper that `fixer`
sits next to, and the walk's answer was the safe one. A `count > 0` in the
condition, or a line in the docblock saying the count is non-negative, keeps the
next caller out of it.

### One more sentence in the same paragraph

> Before 3.0 there is nothing to shorten to: a pattern has no `.` step, so
> `match="self::node()"` does not compile at all.

Both halves are true and the `so` does not connect them. `match="self::node()"`
fails before 3.0 because the *self axis* is not a pattern axis there — which the
paragraph above has just established — not because there is no `.` to shorten it
to. The two facts are independent, and the sentence picks the weaker of the two
reasons and then derives the stronger one from it.

It reads cleanly the other way round: before 3.0 the longhand is not a legal
pattern in the first place, so the question of shortening it never arises; from
3.0 it is legal, and then the shorter form is not a synonym. That is also the
same two-era shape the paragraph before it now has.

Agreed on "are the whole of it" — same reading, same fix as the sentence it
follows.

### The description

You named the `bug(#611)` commit. The body is stale in three more ways, and it is
what the merge carries: it describes a branch that closes #615's half and #627,
where the bulk of it is now the #583 pattern defect — detection, withholding,
`PATTERNS`, the motive, `CLAUDE.md` — none of which is mentioned. `968 tests` is
now 973, and "22 packs for `unabbreviated-axis` and 12 for `using-namespace-axis`"
is now 22 and 16 — the first is right again by accident, the second never was.

Worth a rewrite before merging rather than after. Five commits in, the body is the
only thing a reader gets before the diff.

### The flags

Both. Four rounds, and the last two have been about prose.

Flip the two, fix the `so`, refresh the body, and I have nothing left.

maxonfjvipon:
All four in `9878689`, and the priority direction was the one that mattered.

**Backwards, as you both say.** `.` carries -1 and `self::node()` carries -0.5,
so the rewrite demotes rather than promotes — which is exactly what makes the
clause after it true: the node goes to a template that used to lose because the
one that used to win got weaker. I had the mechanism right and the arrow wrong,
which is the worst way to be wrong about it. Flipped in both places rather than
cut, for the reason you give: a reader who takes the abbreviation for a promotion
reaches for it to win a conflict and gets the opposite. The numbers are in the
sentence now, so the direction cannot drift again.

**The `so`** joined two independent facts and derived the stronger from the
weaker. Reordered: illegal before 3.0 because the self axis is not a pattern
axis, so shortening never arises; legal from 3.0, and then the short form's not
being a synonym is the whole of it.

**"are the whole of it"** is gone — the clause before it already said what was
meant, and `match="*"` and `match="a/b"` made the reading available.

**The negative count.** Good catch, and the walk's answer was the safe one. It
takes the walk now. I ran my own differential after fixing it — 1,730,594 spans
over texts built from `&amp;`, `&lt;`, `&nope;`, a bare `&`, `&#65;`, `\r`,
`\r\n`, `\n`, a bare `;` and multi-byte characters, every offset and every count
from -2 past the end — no mismatch. The induction reason is in the docblock, in
your words, because it is what makes that span the right one to test, and the
entity-at-the-end and negative-count rows are pinned in the source test.

### The description

Rewritten, and the title with it — it is `bug(#583)` now, since that is where the
weight ended up. The body leads with the pattern defect, then #615, #627, and the
positioning change from the last round, which as you say had no business being
discoverable only from the log. Counts corrected: 974 tests, 22 packs and 16.

That the body was three commits stale is the same failure as the piped gates —
trusting a thing I wrote earlier instead of re-reading it against what is there
now.

maxonfjvipon:
Re-reviewed at 9878689. 725 tests here, 100% of 920 branches, CI 18/18.

All four are in and each one checks out. Both flips carry the direction and the
numbers, so it cannot drift again. The reordered `so` puts the stronger reason
first in each era. The negative count takes the walk, and the row pins it —
deleting `count < 0 ||` fails `goes nowhere for a count below zero`, where
nothing failed before. The entity-at-the-end row is the right one to have added,
since it is the case the induction argument turns on. The body is accurate now;
I recounted the packs and 22 and 16 match the tree.

One thing left, and like the last two it is the paragraph rather than the code.

### `..` has one era, not two

> The abbreviated steps never survive the trip, for a different reason in each
> era. [...] From 3.0 it is legal, and then the short form is not a synonym

True of `self::node()`. Not of `parent::node()`, which never becomes legal:
XSLT 3.0 widened the pattern step to the *forward* axes, so the parent axis is
outside it in 3.0 exactly as it was in 1.0.

```text
Saxon-HE 12.9, version="3.0"
  match="self::node()"     HIT
  match="parent::node()"   The parent is not allowed in a pattern
  match=".."               compilation error

xsltproc
  match="parent::node()"   failed to compile 'parent::node()'
  match=".."               failed to compile '..'
```

So a reader with `match="parent::node()"` in a 3.0 sheet is told their pattern is
legal and that the reason to leave it alone is a priority shift, when in fact it
does not compile at all. The paragraph opens on "the abbreviated steps", plural,
and then tells one of the two's story. `descendant`, `self` "and the rest" a few
lines up invites the same reading — the rest of the *forward* axes is what 3.0
admitted, which is precisely why `parent::` is not among them.

A clause covers it, and it simplifies rather than complicates: the parent
longhand was never a legal pattern in any version, so for `..` the question never
arises at all and only `.` needs the two-era treatment. The docblock's "from 3.0
it is legal" wants the same qualifier.

### Two smaller

The title and all seven commits read `bug(#583)`, but #583 is
`starts-with-double-slash` fix shifts template priority in `src/fixers.js` — open,
and untouched here. The body is careful about this ("the #583-class defect"); the
title and the log are not, and I do not see an issue filed for this instance. That
leaves someone triaging #583 with a merged PR carrying its number and the bug
still in `fixers.js`. Filing the instance and pointing the commits at it costs one
issue and keeps #583 meaning what it says.

Line 123 of `src/xpath-axis-linter.js` is 25 characters where the paragraph
around it runs to 80 — the reflow left a seam, the same one you closed in the
motive last round.

### The flags

Both. Add the clause and this is done.

maxonfjvipon (commented):
Re-checked at `9878689`. 725 tests, 100% of 5232 statements and 920 branches,
`npx grunt` unpiped exits 0, CI 18/18. All four from last round are in, and the
negative-count row pins itself — deleting `count < 0 ||` fails
`goes nowhere for a count below zero`, where nothing failed before.

**The `skip` change moves nothing, end to end.** Rather than trust my fuzz alone,
I linted every `.xsl` under `test/resources` on `8dd2489` and on `9878689` and
diffed the reports — check name, line, column, `from`, and the whole `fix` object:

```text
87 stylesheets, 247 defects, byte-identical
```

Nothing to say about that one any more.

### `..` is right, and it is wider than `..`

Confirmed, and the sentence that invites the misreading is the one above it.
It is not only the parent axis that never becomes a pattern step — no *reverse*
axis does, in any version:

```text
                          Saxon-HE 12.9 (3.0)   xsltproc (1.0)
self::node()              HIT                   rejected
descendant::c             HIT                   rejected
parent::node()            XTSE0340              rejected
..                        XTSE0340              rejected
parent::r                 XTSE0340              rejected
ancestor::r               XTSE0340              rejected
preceding-sibling::c      XTSE0340              rejected
```

So "XSLT 3.0 widened the step to `descendant`, `self` and the rest" is where the
error starts: what 3.0 admitted is the rest of the *forward* axes, and the whole
of the reverse half stayed out. Naming that once — 3.0 widened the step to the
forward axes — makes the `..` clause below it follow rather than patch, since a
reader can then see for themselves that `parent::node()` was never in scope.
It also states the rule the check depends on in the form the check depends on it,
which is worth more than an exception for the one abbreviation that happens to
appear in the table.

The docblock wants the same word for the same reason.

### On the `bug(#583)` number

Agreed on the substance, and one thing that changes its urgency rather than its
merit: merging will not close #583. GitHub's closing keyword has to be followed
by the reference, and the body reads "fixes the #583-class defect", so the
linkage is not made — the API bears it out:

```text
closingIssuesReferences: [ #627 ]
```

#615 and #583 both stay open, correctly, since half of one and all of the other
are still live. So this is triage legibility, not a live risk of closing a bug
that is still in `fixers.js` — which is the argument for fixing it in the title
and log rather than urgently.

For what it is worth I would not file an issue for an instance that is already
fixed in the same branch. Retitling to `bug(#615)`, which the branch is named
for and which it genuinely advances, and leaving the body to say the weight
landed on the #583 class, costs nothing and leaves #583 meaning exactly what it
says. Your call — you own the convention.

Line 123 agreed.

### The flags

Both. Five rounds, and everything I have left is one clause.

maxonfjvipon:
In at `085d36f`, and your framing is better than the clause I had written.

I had already patched the `..` case as an exception; naming the forward axes once
makes it follow instead. Verified the rule rather than the examples — xsltproc
rejects every axis but child and attribute, and 3.0's `ForwardAxisP` is forward
only, so the reverse half never enters a pattern in any version:

```text
                       pre-3.0    3.0
child::c / @a          legal      legal
self::node()           illegal    legal
descendant::c          illegal    legal
parent::node() / ..    illegal    illegal
ancestor:: / preceding-sibling::   illegal in both
```

So the motive now says 3.0 widened the step to the forward axes and that the
reverse half stayed out, and the `..` sentence reads off that rather than
asserting a second fact. The docblock takes the same word. The line-123 seam is
gone, and I checked `npx grunt` unpiped rather than through a filter — it exits 0.

### The number

Retitled to `bug(#615)`, and taking your reasoning: an instance fixed in the same
branch does not want an issue of its own, and #583 keeps meaning what it says.
The body now states outright that the commit subjects reference the class rather
than that instance, that #583 is untouched here, and that
`closingIssuesReferences` is `#627` alone — which I confirmed on the API rather
than taking it from your comment.

The commit subjects themselves still read `bug(#583)`. Rewriting seven pushed
subjects means a force-push, which I will not do to a branch under review for a
cosmetic gain; the body carries the correction instead, and it is what a reader
gets first.

### On the end-to-end diff

That is the check I should have run myself for `skip` rather than reasoning about
the induction and then fuzzing the helper in isolation. 87 stylesheets, 247
defects, byte-identical including `from` and the whole `fix` object — that is the
evidence that the change is invisible, and the fuzz only says the helper agrees
with itself.

974 tests, 100% of 920 branches. Both flags set.

maxonfjvipon:
Re-reviewed at 085d36f. 725 tests, 100% of 920 branches, CI 18/18.

Everything from last round is in. Splitting `..` from `.` into their own
paragraphs reads better than the clause I asked for — `..` now needs no exception
because the rule above it already rules the axis out. The title is `bug(#615)`,
`closingIssuesReferences` is `#627` alone (checked on the API), and the line-123
seam is gone. Not rewriting seven pushed subjects for a cosmetic gain is the right
call.

One word, and it is the last thing I have.

### "Forward" is two axes wider than the rule

> XSLT 3.0 widened the step to the forward axes — `descendant`, `self`,
> `descendant-or-self`, `namespace` — [...] The reverse axes stayed out, which is
> why `parent::` never became a pattern step in any version.

The list is exactly right and the class name around it is not. `following::` and
`following-sibling::` are forward axes, and a 3.0 pattern refuses them the same
way it refuses the reverse half:

```text
Saxon-HE 12.9, version="3.0"
  self::y                  compiles      following::y             XTSE0340
  descendant::y            compiles      following-sibling::y     XTSE0340
  descendant-or-self::y    compiles      a/following-sibling::y   XTSE0340
  namespace::*             compiles      parent::a                XTSE0340
  attribute::i             compiles      ancestor::r              XTSE0340
                                         preceding-sibling::x     XTSE0340
```

So the dichotomy the paragraph rests on — forward in, reverse out — is not the
rule, and the conclusion about `parent::` is reached for a reason that does not
hold. The docblock's "the parent axis being reverse where a pattern step takes
forward axes only" has it in the same shape.

Saxon words the real rule in its own diagnostic: *only downwards axes are allowed
in a pattern*. The six a pattern step admits — child, descendant,
descendant-or-self, self, attribute, namespace — are exactly the downward ones,
and `following::` is the case that tells the two words apart. Swapping "forward"
for "downward" in both places makes both sentences true and leaves the argument
standing as written.

Worth the swap rather than a footnote, because #631 is a check about pattern axes
and this paragraph is where whoever writes it will go for the rule.

### The flags

Both. Change the word and I have nothing left — the code has been done for two
rounds.

maxonfjvipon:
Swapped in `5d55344`, and the mistake has a source worth naming: the grammar
production is called `ForwardAxisP` while enumerating only those six, so I took
the production's name for the rule it states. `following::` is exactly the case
that separates the two words, and it is a forward axis a pattern refuses.

Both places now say downward, which is how Saxon words it in its own diagnostic,
and "the reverse axes stayed out" becomes "every axis that reaches elsewhere
stayed out, `following::` as much as `parent::`" — so the `parent::` conclusion
follows from the rule rather than from a dichotomy that is not one.

You are right that this paragraph is where #631's author will go for the rule,
which is the reason it was worth a swap and not a footnote. I have added the
enumeration to that issue so it does not depend on the motive being read
correctly.

974 tests, 100% of 920 branches, `npx grunt` unpiped exits 0.

Five rounds on prose after the code settled, and the pattern in what you have
caught is consistent: every one has been a sentence that was true of the example
in front of me and false of the class I named it with — "processors drop the
axis", "a pattern is free with axes", "outranks", "forward". The examples were
checked each time; the generalisation was not. That is the habit to fix, not the
four sentences.

maxonfjvipon:
Checked at 5d55344. 725 tests, 100% of 920 branches, CI 18/18. No "forward" left
in either place, and "every axis that reaches elsewhere stayed out, `following::`
as much as `parent::`" states the rule the `parent::` conclusion needs. Nothing
left from my side.

Two things I had not run before, both worth having in the thread.

**Nothing unintended moved.** Linting all 87 committed `.xsl` fixtures under
`test/resources` on master and on this branch, comparing check name, line,
column, `from` and the whole `fix` object:

```text
247 defects on each, byte-identical
```

So the positioning change, the `PATTERNS` plumbing and the widened `ATTRIBUTES`
leave every existing report exactly where it was.

**Everything intended did.** One stylesheet holding all four changes, `--fix`
run on each side:

```text
                                          master              branch
match="self::node()"                      .                   untouched
match="y|self::node()"                    y|.                 untouched
count="self::node()" from="parent::node()"  . / ..            untouched
match="child::  gapped"                     "  gapped"        gapped
match="attribute::  id"                     "@  id"           @id
select="attribute::  spaced"                "@  spaced"       @spaced
use-when="child::flag"                    unread              flag
for-each-item="child::doc"                unread              doc
```

Four of those master rows do not compile or silently move a rule, and none of the
branch rows does either. That is the whole of the diff, and it is the diff the
description claims.

Good to merge, both flags included. `unabbreviated-axis` earned it the hard way —
the defect that blocked it was closed rather than deferred, the withholding and
the silence are each pinned by a pack that fails on the revert, and the two
rewrites that survive in a pattern are the two shown not to move a rule on two
processors. `using-namespace-axis` I could not break in six rounds.

On the habit: worth writing down somewhere durable, because the four sentences
had one shape — a claim checked against the example in front of you and then
stated about the class the example belongs to. It is also why the last five rounds
were cheap to review and expensive to have written: the examples always held, so
nothing failed, and only the generalisation was wrong.

PR #634, opened 2026-08-03
bug(#633): derive a stylesheet's expressions once, cheaply, and by walking

### LINKED ISSUE #633: Linting one stylesheet is quadratic in its size
Linting one stylesheet is quadratic in its size, and the constant is large enough
that an editor cannot use it. Measured through `lint()`, the exported entry point
#336's language server is meant to call:

```text
   85 lines    0.18 s
  405 lines    1.55 s
 1205 lines   13.62 s
```

DocBook-XSL and DITA-OT — the stylesheets the README cites — run to thousands of
lines each. At a keystroke, an editor needs a hundred milliseconds or so; this is
two orders out on a file of ordinary size, and the curve is the problem rather
than the constant.

Two causes, one behind the other.

## `expressionsOf` runs once per linter, not once per document

Eight code-based linters call `expressionsOf(xsl)` on the same document, so the
whole expression list is derived eight times per run. At 485 lines that is 118 ms
each and 981 ms of the 1004 ms the thirteen code-based linters take between them:

```text
xpath-axis-linter              125.8 ms
count-linter                   123.7
name-linter                    123.3
using-namespace-axis-linter     122.9
redundant-double-negation       122.2
string-length-linter            121.2
predicate-position-linter       120.8
translate-linter               120.8
-- expressionsOf alone          118.0
```

The five linters that do not call it — `node-set`, `redundant-boolean-call`,
`import`, `result-namespace`, `namespace` — cost between 0.1 ms and 7 ms.

`src/xslint.js` already computes the expression stream once for the expression
linters and hands it over. The same shape fits here: derive it once beside the
corpus and pass it to the document linters, which is one change at the wiring
and a signature change per linter.

## `SELECTOR` is a nineteen-way union of `//` scans

`expressionsOf` opens with `nodes(xsl, SELECTOR)`, where `SELECTOR` is
`ATTRIBUTES.map(selectorOf).join(' | ')` — nineteen separate `//xsl:*/@name`
descendant scans, unioned. That union is where the quadratic term lives:

```text
lines   19-way union   one scan + a filter
   85          26 ms          2 ms
  325         114              7
  805         618             38
 2005        4379            273
```

One `//xsl:*/@*` scan filtered by name in JavaScript returns the identical node
set — checked at every size above — and is sixteen times faster at 2000 lines.
`selectorOf` is still needed for the linters that narrow to a single attribute;
it is only the union of all nineteen that costs.

Together the two are multiplicative: eight times the same nineteen-way union.
Fixing both should bring 1205 lines from 13.6 s to well under half a second,
which is the range an editor can debounce against.

Worth measuring again after: at 2000 lines the single scan is still 273 ms
against 38 ms at 805, so something superlinear remains in fontoxpath's `//`
evaluation, and it may need `getElementsByTagName` plus a namespace test rather
than XPath at all. That is a second step, and the two above are worth having
first.

maxonfjvipon:
Closes #633 and #635.

```text
lines    before      now
   85    0.18 s    0.06 s
  405    1.55 s    0.09 s
 1205   13.62 s    0.32 s
 4005       --     1.86 s
```

Three changes. The first two are #633 and sit in `src/attributes.js`; the third
is #635 and adds `src/tree.js`.

### Eight linters, one list

`expressionsOf` walks every attribute and text node of a stylesheet, and eight
code-based linters asked it for the same list in a single run. At 485 lines that
was 118 ms each and 981 ms of the 1004 ms those linters took between them.

It is remembered against the document, in a `WeakMap` rather than a `Map` keyed by
anything the corpus outlives — the mistake #611's review caught in the line-offset
memo, and worth not repeating in the same file's neighbour. The list is frozen,
since eight callers now share one array.

### Nineteen scans, one scan

`SELECTOR` was `ATTRIBUTES.map(selectorOf).join(' | ')` — nineteen `//` descendant
scans unioned — and that union is where the quadratic term lived:

```text
lines   19-way union   one scan + a filter
   85          26 ms          2 ms
  325         114              7
  805         618             38
 2005        4379            273
```

One scan of every XSLT attribute, filtered by name in JavaScript, answers the
identical set. `selectorOf` stays for the linters that narrow to a single
attribute; it was only asking for all nineteen at once that cost.

### Unchanged output

Every `.xsl` under `test/resources` linted before and after and the reports
diffed — check name, line, column, `from`, and the whole `fix` object:

```text
87 stylesheets, 247 defects, byte-identical
```

That is the evidence the change is invisible; the suite passing is the weaker
claim, and this is the check the last branch taught me to run rather than reason
about.

### A gate that had been failing quietly

`npm run coverage` now exits 0 on my machine. `should test default directory`
lints the whole repository and had been over mocha's ten-second limit under
instrumentation — on master too, so it was not this branch's doing, but it means
the coverage gate had been red locally for anyone whose machine is not as fast as
CI's.

974 tests, 100% branch coverage (922/922), ESLint, xcop, markdownlint and the
fixtures gate clean locally, all read from exit codes rather than through a
filter.

### #635: walk the tree rather than ask XPath to descend it

fontoxpath evaluates a descendant step over an xmldom tree quadratically, and
that was the whole of what remained:

```text
lines    descendant scan    DOM walk    nodes
  165              15 ms      0.2 ms      366
  645              49         0.5        1446
 2005             486         1.7        4506
 4805            3060         2.7       10806
```

Both internal scans walk now — `expressionsOf` and the validator's, whose name
test moves from an XPath predicate to a filter. `expressionsOf` drops from the
dominant stage to 8 ms on a 2405-line stylesheet.

Matching the old node sequence exactly took two details, and I found both by
diffing the sequences across every fixture rather than by reading the spec:
XPath does not count a namespace declaration among an element's attributes
though the DOM lists it, and fontoxpath yields attributes sorted by name where
the DOM gives them as written — `zeta alpha mid` comes back `alpha mid zeta`.
XPath leaves that order to the implementation, so matching it is what keeps this
invisible. Reporting in written order may read better; that is a change to make
deliberately rather than as a side effect of a performance fix.

### What is left, and why it is not this PR

End to end the run improves by less than the scan does. At 2405 lines:

```text
parse                            18 ms
expressionsOf (walk)              8
xpath-validator (walk + isValid) 150
xpath-linter (67 checks' XPath)  509
corpus-linter (checks' XPath)    491
13 code-based linters            570
```

The two largest are the declarative checks' own XPath — that is the check, not a
scan this project chose, so it stays XPath. #635 named `corpus-linter` among the
places to change and that was wrong of me: its scans are `check.declaration` and
`check.usage` straight out of the YAML. Making those cheaper means compiling the
checks differently or a faster engine, which is its own question.

So: 405 lines is 93 ms and comfortable for an editor, 1205 is 320 ms and wants
debouncing, 4005 is 1.9 s and is still felt. Better than 13.6 s, not yet finished.

### Unchanged output, three times over

Every `.xsl` under `test/resources` linted before and after each change, reports
diffed on check name, line, column, `from` and the whole `fix` object:

```text
87 stylesheets, 247 defects, byte-identical
```

974 tests, 100% branch coverage (941/941), and every gate read from its exit code
rather than through a filter.

### DIFF
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -3,9 +3,9 @@
  * SPDX-License-Identifier: MIT
  */
 
-const {nodes} = require('./xpath')
 const {enclosed} = require('./expressions')
 const {XSLT, since, versionOf} = require('./xsl-version')
+const {walked} = require('./tree')
 
 /**
  * Attributes that hold an XPath expression or a pattern — every place a
@@ -154,10 +154,12 @@ const carried = function(node, bare, three) {
  */
 const expressionsOf = function(xsl) {
   if (!DERIVED.has(xsl)) {
-    const bare = new Set(
-      nodes(xsl, '//xsl:*/@*').filter((one) => NAMED.has(one.nodeName)),
-    )
-    DERIVED.set(xsl, Object.freeze(nodes(xsl, '//*/@* | //text()').flatMap(
+    const held = walked(xsl)
+    const bare = new Set(held.filter(
+      (one) => one.nodeType === 2 && NAMED.has(one.nodeName) &&
+        one.ownerElement.namespaceURI === XSLT,
+    ))
+    DERIVED.set(xsl, Object.freeze(held.flatMap(
       (node) => carried(node, bare, since(versionOf(node), '3.0')),
     )))
   }
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -48,10 +48,25 @@ const selectorOf = function(name) {
 }
 
 /**
- * XPath selecting every attribute that holds a bare XPath, across the document.
- * @type {string}
+ * The names that hold a bare XPath, for testing one attribute at a time. Asking
+ * XPath for all of them at once — a union of nineteen descendant scans, one per
+ * name — costs more than a single scan of every XSLT attribute filtered by name
+ * here, and costs it quadratically: 4.4 s against 0.3 s on a 2000-line
+ * stylesheet (#633).
+ * @type {Set.<string>}
  */
-const SELECTOR = ATTRIBUTES.map(selectorOf).join(' | ')
+const NAMED = new Set(ATTRIBUTES)
+
+/**
+ * The expression list already derived for a document. Eight code-based linters
+ * ask for the same one in a single run, and deriving it walks every attribute
+ * and text node of the stylesheet, so it is derived once and remembered against
+ * the document itself — which a `WeakMap` releases when the corpus does, rather
+ * than holding every stylesheet ever linted (#633). The list is frozen, since
+ * eight callers now share one array.
+ * @type {WeakMap}
+ */
+const DERIVED = new WeakMap()
 
 /**
  * The spellings that switch an XSLT boolean attribute on, whitespace trimmed —
@@ -136,10 +151,15 @@ const carried = function(node, bare, three) {
  *  expressions found, each saying whether it is a pattern
  */
 const expressionsOf = function(xsl) {
-  const bare = new Set(nodes(xsl, SELECTOR))
-  return nodes(xsl, '//*/@* | //text()').flatMap(
-    (node) => carried(node, bare, since(versionOf(node), '3.0')),
-  )
+  if (!DERIVED.has(xsl)) {
+    const bare = new Set(
+      nodes(xsl, '//xsl:*/@*').filter((one) => NAMED.has(one.nodeName)),
+    )
+    DERIVED.set(xsl, Object.freeze(nodes(xsl, '//*/@* | //text()').flatMap(
+      (node) => carried(node, bare, since(versionOf(node), '3.0')),
+    )))
+  }
+  return DERIVED.get(xsl)
 }
 
 module.exports = {
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -62,8 +62,10 @@ const NAMED = new Set(ATTRIBUTES)
  * ask for the same one in a single run, and deriving it walks every attribute
  * and text node of the stylesheet, so it is derived once and remembered against
  * the document itself — which a `WeakMap` releases when the corpus does, rather
- * than holding every stylesheet ever linted (#633). The list is frozen, since
- * eight callers now share one array.
+ * than holding every stylesheet ever linted (#633). Eight callers share one
+ * array, so the array is frozen against having its entries swapped and each
+ * entry is frozen where `carried` builds it — freezing the array alone leaves
+ * `held.expression = ...` free to poison the other seven.
  * @type {WeakMap}
  */
 const DERIVED = new WeakMap()
@@ -130,11 +132,11 @@ const carried = function(node, bare, three) {
     (bare.has(node) || (three && shadow(node)))
   const braced = node.nodeType === 2 || (three && expands(node))
   return whole ?
-    [{
+    [Object.freeze({
       node: node, start: 0, expression: node.nodeValue,
       pattern: PATTERNS.includes(node.nodeName.replace(/^_/, '')),
-    }] :
-    braced ? enclosed(node.nodeValue).map((found) => ({
+    })] :
+    braced ? enclosed(node.nodeValue).map((found) => Object.freeze({
       node: node, start: found.offset, expression: found.value, pattern: false,
     })) : []
 }
--- a/src/tree.js
+++ b/src/tree.js
@@ -1,0 +1,80 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+ * SPDX-License-Identifier: MIT
+ */
+
+/**
+ * Node kinds a walk collects: an attribute, a text node, and the CDATA section
+ * that is a text node written another way.
+ * @type {Array.<number>}
+ */
+const CARRIED = [2, 3, 4]
+
+/**
+ * Namespace declarations, which XPath does not count among an element's
+ * attributes however the DOM lists them — they are namespace nodes, reached by
+ * an axis of their own.
+ * @type {RegExp}
+ */
+const DECLARED = /^xmlns(:|$)/
+
+/**
+ * The walk already taken over a document. Every stage that needs the attributes
+ * or the text of a stylesheet wants the same sequence, so it is walked once and
+ * remembered against the document — released with it, a `WeakMap` holding no
+ * stylesheet the corpus has finished with.
+ * @type {WeakMap}
+ */
+const WALKED = new WeakMap()
+
+/**
+ * Every attribute, text node and CDATA section of a document, in document
+ * order: each element's attributes in the order they were written, then its
+ * children in theirs. Two details keep the sequence the one a descendant scan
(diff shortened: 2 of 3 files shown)


maxonfjvipon:
Force-pushed `477de2f` over the first push, which had a mistake worth naming
rather than quietly correcting.

My `git add -A` swept five files into the commit that are not mine: three
`.claude/commands/pr-*.md`, `scripts/pr-loop.sh`, and 42 lines of `CLAUDE.md`
describing them. They were untracked in the working tree, and the commit made
them tracked — which is what turned `markdown-lint` red, on `MD041` in files I
had never opened. The branch now carries one file:

```text
src/attributes.js | 34 +++++++++++++++++++++++++++-------
```

Two things about that. Reaching for a revert commit would have been worse than a
force-push here: it would have left the files tracked in history and *deleted*
them from the working tree, when the intent is plainly that they stay untracked —
there are `.gitignore` entries for exactly those paths. Force-pushing an
unreviewed branch of my own, minutes old, is the narrow way to put it back. I
would not have done it to a branch under review.

And the CI failure is the only reason I found it. `markdown-lint` named three
files, and my first instinct was that it was pre-existing, because my local run
reported 41,228 errors on this branch and the same 41,228 on master — my glob was
sweeping `node_modules` under stale worktrees, so it could not have told me
anything. Reading the CI job log rather than trusting a local number is what
found it. That is the fourth time on this stretch that a filtered local view read
green or read equal when the real gate did not.

17/17 green, 974 tests, and `npm run coverage` exits 0 for the first time on this
machine — which is the change itself: `should test default directory` lints the
whole repository and had been over mocha's ten-second limit under instrumentation.

maxonfjvipon:
Reviewed on the branch at 477de2f. 725 tests, 100% of 922 branches, CI 18/18.
Both changes are right and the win is real. It does not close #633.

**The output is unchanged, and I checked it rather than taking it.** Every `.xsl`
under `test/resources` linted on master and on the branch, comparing check name,
line, column, `from` and the whole `fix` object: 87 stylesheets, 247 defects,
identical. The `nodeName` filter is set-identical to the union it replaces — a
prefixed `xsl:select`, a shadow `_select` and an `xmlns:` declaration are excluded
by both, for the same reason in each case.

**The `WeakMap` is safe, and for a reason worth writing down.** It rests on the
document not changing after it is parsed, which holds: the only mutation in `src/`
is `expand` in `helpers.js`, and that runs inside the parser before any linter
sees the tree. Nothing states the invariant, though, and it is exactly what an
in-place `--fix` or #336's incremental re-lint would break — a line in the
docblock costs nothing and is where the next person will look.

### It is a constant, not the curve

#633 says it in as many words: "the curve is the problem rather than the
constant." The curve is where it was. Through `lint()`, same generated
stylesheets, same defect counts on both sides at every size:

```text
lines   defects    master    branch   speedup
   76        38     0.12s     0.07s     1.7x
  364       182     0.56      0.09      6.2
 1204       602     5.35      0.44     12.2
 2404      1202    22.73      1.42     16.0
 4804      2402   100.47      5.42     18.5
```

Fitting 1204 to 4804 — four times the lines — master takes 18.8x the time and the
branch 12.3x. That is n^2.1 against n^1.8. The speedup grows with size because one
of two quadratic factors is gone; the one left is still quadratic.

It also misses the target #633 set for itself. The issue predicted "well under
half a second" at 1205 lines and wants "a hundred milliseconds or so" at a
keystroke. This is 0.44s at 1204 lines here, 0.90s on your machine, and 5.42s at
4804 — which is the DocBook-XSL and DITA-OT scale the issue cites as the case that
matters. So `Closes #633` will shut an issue whose problem statement is still
true on merge. Worth a partial reference and a second issue, or reopening #633
scoped to what is left.

None of which argues against merging. A 1200-line stylesheet going from 5.4s to
0.44s is the difference between unusable and usable, and it is one file.

### Where the rest of it is

Measured on the branch, so the 8x and the union are already out:

```text
lines   //xsl:*/@*   //*/@* | //text()   expressionsOf
 1204        31ms                75ms          107ms
 2404       129                 306            432
 4804       552                1323           1906
 9604      2281                5740           8086
```

Both fontoxpath scans quadruple as the lines double — textbook quadratic — and
together they are 8021 ms of the 8086. So `expressionsOf` is now nothing but two
`//` scans, and the remaining curve is entirely fontoxpath's.

A plain `firstChild`/`nextSibling` traversal of the same tree, collecting the same
node kinds, is linear and takes 0.7 ms at 11,204 lines. So the ceiling on a
handwritten replacement is about three orders, not the sixteen-fold the union
gave.

Two differences to reconcile before anyone reaches for it, which is why it wants
its own change and its own equivalence check rather than a footnote here. I ran
the comparison and the sets are the same size but not the same set:

```text
in the XPath, not the walk:   the document-level text node after the root
in the walk, not the XPath:   @xmlns:xsl
```

`//text()` reaches text outside the document element, and `@*` excludes namespace
declarations where `element.attributes` includes them. Both are one line each —
start at the document rather than `documentElement`, skip `xmlns` and `xmlns:*`.
Worth knowing that identity is shared: fontoxpath hands back the real DOM nodes,
so `bare.has(node)` keeps working across a mixed pair, which is what makes the
replacement possible at all.

### Smaller

`Object.freeze` is shallow. The array cannot be reassigned but its entries can —
one caller writing to `held.expression` would poison the other seven, and the
comment ("frozen, since eight callers now share one array") reads as though that
were covered. Freezing the entries in `carried` costs a call per expression, or
the comment can say which half is guarded.

Nothing else. Merge it, and split the issue.

maxonfjvipon:
Yes, it is a problem — and the comment was the worse half of it, since it told a
reader not to look again.

Demonstrated before touching anything:

```text
array frozen?  true
entry frozen?  false
first[0] = {...}            silently ignored, value stays "/"
first[0].expression = 'X'   the next caller reads 'X'
same array object?          true
```

So the shallow freeze guarded the one thing nobody was going to do — swapping an
entry — and left open the one a linter might: writing through to a field. Eight
callers, one array, no barrier.

Fixed in `5b5c39d`, at `carried`, where the entries are built rather than after
the fact:

```text
array frozen? true   every entry frozen? true
after a write, another caller reads: "/"
```

Measured rather than assumed, since it is a per-expression call: 0.048 ms for 400
expressions and 0.331 ms for 1600, against derivations of 47 ms and 537 ms. End
to end the timings do not move — 1205 lines is 0.93 s against 0.90 s, inside the
noise of the run.

No caller mutates today, and I checked all eight rather than reasoning about it,
so this closes a door rather than fixing a live defect. Which is the reason to
take the fix over the alternative you offered: a comment saying only the array is
guarded would be true and would leave the hazard, and the cost is 0.06% of the
work it protects.

974 tests, 100% of 922 branches, and `npm run coverage` exits 0.

Unrelated, and worth flagging: something moved this checkout to `master` and
pulled mid-session — the reflog shows `checkout: moving from
bug/#633/derive-expressions-once to master` followed by a fast-forward, which I
did not run. Nothing was lost, `477de2f` was already on the remote, but the first
measurements I took afterwards were against master's un-memoised code and read as
1452 ms for eight warm reads until I noticed. If `scripts/pr-loop.sh` is driving
this PR, it and I are sharing one working tree.

Issue #589, opened 2026-07-30
The validator keeps its own list of attribute names, so 165 of 451 expressions are never validated


maxonfjvipon:
Two modules answer the question "what expressions does this stylesheet carry", and
they answer it differently. `src/attributes.js`'s `expressionsOf` walks the tree and
yields a record per expression — a bare attribute of an XSLT element, each expression
an attribute value template holds in its braces, a 3.0 text value template, a shadow
attribute — each saying whether it is a `pattern`. `src/validators/xpath-validator.js`
derives a list of attribute *names* and walks for those:

```js
const EXPRESSIONS = ATTRIBUTES.filter((name) => !PATTERNS.includes(name))
```

Since #647 that list is `ATTRIBUTES` less `PATTERNS` rather than a rival list, so the
twelve-against-twenty drift this issue was filed about is gone. What is left is the
subtraction itself, and it is the whole gap: over this repository's own 118 stylesheet
fixtures `expressionsOf` yields **451 expressions and the validator walks 286**. The
other 165 are validated by nothing, so on a default run:

```text
match="foo["            no defect at all
title="{1 +}"           no defect at all
match="following::c"    no defect at all, in a version="1.0" sheet
match="a[count(b) = 0"  count-compared-to-zero, with its fix withheld
```

The third of those is #631, whose detection has existed since #723 — `matched`
refuses an axis no 1.0 or 2.0 pattern may name, and names the version in the
message — with nobody to report it. The fourth is the partition rule from CLAUDE.md
read from the other side: a code-based linter takes the whole corpus and reports what
it finds in text the validator never looked at, and only #636's gate in `defect`
keeps the fix off it. Measured on `2bbe24a`.

The pattern half of the original filing is answered, and by something better than
the interim it proposed: `matched(pattern, version)` in `src/grammar.js` judges a
pattern by its own productions (#678, #679), and `isValid` reaches for it whenever
the record says `pattern`. So nothing is missing underneath any more. What is
missing is that the validator does not consume the derivation the linters do.

The repair is one derivation and one parse, which is properties 1 and 2 of the
pipeline in #644. `expressionsOf` becomes the single source of truth the validator
reads too, and each record carries its tree or its syntax error, computed once
rather than recomputed as a boolean behind `isValid`'s cache: today the parser is
asked for a verdict and the tree it built is thrown away, so nothing above can walk
it. Then every expression a stylesheet holds is validated, the fault is reported at
the offset the refusal already carries rather than at the attribute's own start,
[...]

PR #646, opened 2026-08-04
bug(#636): offer no fix on an expression the validator refused

### LINKED ISSUE #636: A linter fixes an expression `invalid-xpath-expression` rejected, in `src/xslint.js`

maxonfjvipon:
Closes #636.

`lint` partitions the expressions and then hands the whole corpus to every
code-based linter, so an expression `invalid-xpath-expression` has already
refused still reaches `expressionsOf` and still collects a fix. On
`select="child::"` the run reported the malformed expression and offered a
plain `--fix` that wrote `select=""`, replacing text the user wrote with an
empty attribute and leaving the syntax error in place for the next run.
`test="count(alpha) = 0 ("` behaved the same way.

The fix is withheld rather than the report, which is the smaller of the two
corrections the ticket weighs: a linter that read a broken expression may well
have found something real in it, and saying so costs nothing, while rewriting it
cannot be right. `defect` in `src/checks.js` is the one place every code-based
linter builds a defect through, so the gate goes there:

```js
const anchored = fix === undefined || !isValid(expression) ?
  undefined :
  {line: line, col: pos, ...fix}
```

The expression is now a parameter of `defect` rather than something it could
infer. It cannot be inferred: for a bare attribute the node's value is the
expression, but for one an attribute value template encloses in braces, or a
3.0 text value template, the node's value is not an expression at all and
would never parse. Every call site already had the text in scope.

Making it a required parameter is also what keeps the convention from drifting.
The project's own `no-missing-argument` rule weighs `Function.length`, so a
linter added later that calls `defect` without its expression is a lint error at
that call, not a silent return to the old behaviour. That is the
machine-enforcement CLAUDE.md asks for, and it costs nothing beyond the
parameter.

`isValid` runs only when a fix is actually offered — `fix === undefined`
short-circuits ahead of it — so the cost is one compile per fixable defect, not
one per expression. Measured rather than assumed: 0.082 ms a call, so 82 ms at a
thousand fixable defects, which no real run reaches.

End to end, over a fixture holding both kinds beside each other:

```text
<xsl:value-of select="child::"/>              refused, no fix, file untouched
<xsl:if test="count(alpha) = 0 (">            refused, no fix, file untouched
<xsl:value-of select="child::beta"/>          fixed to select="beta"
<xsl:value-of select="child ::gamma"/>        fixed to select="gamma"
```

The last row is the one #636 waited on #615 for. A spaced axis was refused by
`isValid` until #640 landed, so gating on validity before that would have
silenced a correct fix; it now parses, so it keeps its fix and only the genuinely
broken expressions lose theirs.

Two pieces of prose came with the behaviour. `validate`'s docblock promised that
"no linter ever reasons over broken input", which was never true of the
code-based linters and is not what this change makes true either — they still
read a refused expression and still report on it. It now says what the staging
actually does and where the fix is withheld. `README.md` gains the user-facing
half: an expression xslint cannot parse is never rewritten, and the corrections
appear once the syntax is fixed.

A pattern is judged by `isValid` too, and that is where this change is not
free. Patterns are almost a subset of expressions, so a pattern the engine
cannot parse loses its fix rather than gaining a wrong one — the safe direction —
but the loss is real and I measured its edge rather than assuming it away. Of
twenty-seven legal `match` patterns, exactly two are refused, both the XSLT 3.0
item-type form:

```text
~xs:string      refused
~element(a)     refused
```

Everything else — `/`, `a//b`, `@id`, `document-node(element(a))`,
`key('k','v')`, `root()`, the axes and the kind tests — parses and keeps its fix.
So on a 3.0 sheet a fixable defect inside something like
`match="~element(a)[count(b) = 0]"` is now reported without its correction. That
is a narrow and quiet regression, not a wrong rewrite, and #589 is where patterns
get a grammar of their own; naming it here so the next reader does not have to
rediscover it.

Left out deliberately: the staging itself. Handing code-based linters only the
validated expressions would drop the reports as well as the fixes, and would
mean `expressionsOf` — which covers patterns, attribute value templates and text
value templates that the validator never looks at — answering to a validator
that does not cover them. That is #589's and Phase 2's business, not this
ticket's.

`npm test` 771 passing, `npm run coverage` 100% of statements, branches,
functions and lines, `test/xcop.test.js` 250 passing with the tool on PATH, and
the new fixture passes `xcop` on its own.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -342,7 +342,12 @@ the harness asserts too.
   `xslint-disable-line`, `xslint-disable-file`, each with optional space-separated
   rule names (`src/directives.js`); an unused directive is reported.
 - **Fix tiers**: a defect is fixable when it carries
-  `fix: {line, col, value, replacement, suggestion?}`. A *safe* fix
+  `fix: {line, col, value, replacement, suggestion?}`, and only when the
+  expression it stands in parses. A code-based linter still reads an expression
+  the XPath validator refused — it takes the whole corpus, not the validated
+  part — so it still reports what it finds there, but `defect` withholds the
+  fix, because rewriting text no processor parses is how `select="child::"`
+  became `select=""` (#636). A *safe* fix
   (deterministic, semantics-preserving) is applied by `--fix`; a
   `suggestion: true` fix (changes behavior, or is one of several corrections) is
   applied only by `--fix-suggestions`. `--fix-dry-run` writes nothing.
@@ -368,7 +373,7 @@ the harness asserts too.
 | `src/xpath-linter.js` | Loads `checks/xpath/*.yaml`; attaches any `src/fixers.js` fix |
 | `src/corpus-linter.js` | Loads `checks/corpus/*.yaml`; cross-file rules |
 | `src/*-linter.js` | Code-based `checks/format/*.yaml`, one construct each (axis, namespace, count, name, ...); see the flow diagram |
-| `src/checks.js` | Shared for code-based linters: `metaOf`, `suppressed`, `defect(check, meta, source, node, offset, fix)` — walks the raw text so a wrapped or entity-shifted value reports where it truly stands (#611) |
+| `src/checks.js` | Shared for code-based linters: `metaOf`, `suppressed`, `defect(check, meta, source, node, offset, expression, fix)` — walks the raw text so a wrapped or entity-shifted value reports where it truly stands (#611), and drops the `fix` when `expression` does not parse (#636) |
 | `src/source.js` | Raw-text walking shared by `checks` and `fixer`: `offsetAt`, `placeAt`, `character`, `skip` |
 | `src/attributes.js` | `expressionsOf(xsl)` — every expression a stylesheet carries: a bare/AVT attribute, a 3.0 text value template, or a shadow attribute, each saying whether it is a `pattern`; `PATTERNS` names the five attributes that hold one; `selectorOf(name)` for a linter that narrows |
 | `src/xsl-version.js` | `versionOf(node)` — the version in force at a node, from the nearest ancestor's `@version` (XSLT element) or `@xsl:version` (literal result element), canonicalised as a decimal; `since(version, floor)` for a lower-bound gate; shared `MODERN`/`KNOWN`/`DECIMAL` |
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -342,12 +342,14 @@ the harness asserts too.
   `xslint-disable-line`, `xslint-disable-file`, each with optional space-separated
   rule names (`src/directives.js`); an unused directive is reported.
 - **Fix tiers**: a defect is fixable when it carries
-  `fix: {line, col, value, replacement, suggestion?}`, and only when the
-  expression it stands in parses. A code-based linter still reads an expression
-  the XPath validator refused — it takes the whole corpus, not the validated
-  part — so it still reports what it finds there, but `defect` withholds the
-  fix, because rewriting text no processor parses is how `select="child::"`
-  became `select=""` (#636). A *safe* fix
+  `fix: {line, col, value, replacement, suggestion?}`. A code-based linter still
+  reads an expression the XPath validator refused — it takes the whole corpus,
+  not the validated part — so it still reports what it finds there, but `defect`
+  withholds the fix, because rewriting text no processor parses is how
+  `select="child::"` became `select=""` (#636). That gate lives in `defect`, so
+  it covers the code-based linters only: a declarative fix from `src/fixers.js`
+  is attached in `src/xpath-linter.js`, never passes through `defect`, and is
+  still offered on an expression that does not parse (#651). A *safe* fix
   (deterministic, semantics-preserving) is applied by `--fix`; a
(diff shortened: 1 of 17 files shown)


maxonfjvipon:
### Round 1 — fixed 2, deferred 2, rebutted 0

`378a07e`.

**Blocker — the README claim is false.** Confirmed, and it is worse than a wording
slip: I reproduced both halves on the head before touching anything.

```text
$ node src/index.mjs d.xsl                     # both reported
[ERROR]   d.xsl(5:26) An XPath expression is malformed ... (invalid-xpath-expression)
[WARNING] d.xsl(5:5)  ... (select-starts-with-double-slash)
[WARNING] d.xsl(4:26) ... (unabbreviated-axis)     <- no fix, #636 working

$ node src/index.mjs d.xsl --fix                # plain, no suggestions
-  <xsl:template match="//child::">
+  <xsl:template match="child::">

$ node src/index.mjs d.xsl --fix-suggestions
-    <xsl:value-of select="//child::"/>
+    <xsl:value-of select=".//child::"/>
```

The second edit rewrites text the same run called malformed. So the sentence
claimed a guarantee the tree does not give, which is the one thing prose in this
repo may not do.

Narrowed rather than extended, and filed as **#651** with that reproduction. The
gate genuinely does not port as a one-liner: a declarative fixer takes the node
its selector matched, and that node need not carry an expression at all, so it
has to ask whether the node is an expression-bearing attribute *before* it asks
whether the text parses — and attribute value templates fail the parse test for
the wrong reason. That is its own change with its own tests, not a line in this
one. `README.md` now says which checks withhold and points at #651; `CLAUDE.md`
says the gate lives in `defect` and therefore reaches the code-based linters
only.

**The pattern regression.** Agreed on all counts, including that gating only
where `pattern` is false would be worse — it would keep bogus fixes on broken
patterns to buy back two spellings. Filed as **#649** so it survives outside this
pull request's body, with the two refused forms and the `~element(a)[count(b) =
0]` case named.

**The unresolved-entity path.** Thank you for checking it; I had not. Recording
your finding here so it is not re-derived: the fix on `count(&nodes;) = 0` is now
withheld up front instead of being built, offered, and then refused by
`src/fixer.js` with a *the source no longer matches* warning. Same outcome, one
less confusing message.

**Nit 1, seven parameters.** Agreed, and it is the better shape. It is not
uniform today, which is why I did not do it here: `node-set-linter` and
`redundant-boolean-call-linter` do not read `expressionsOf` at all — they select
through `selectorOf` and hold a bare attribute — so they have no entry to pass
and would have to synthesise one. Folded into **#648**, which is the same
refactor area: `expression` already names a node in one module and its text in
another, and both are worth settling in one pass rather than two.

**Nit 2, the inert call.** Confirmed inert, not merely redundant:
`xpath-format-linter` consumes the `expressions` that `validate` already kept, so
`isValid` there can only answer true. One compile per redundant-whitespace
defect, and the gate on that path is decoration. It reads as coverage, which is
the fair complaint. Left as is rather than special-cased, since the parameter is
what makes the rule enforce itself at the next call site, and noted here so the
next reader is not misled — Phase 3 of #644 removes the path entirely when
`expressionsOf` starts yielding parsed expressions.

**Nit 3, the shared constants.** Fixed. `REFUSED` and `KEPT` are gone and each
test carries its own literals, which also let the test names say what they check
rather than which array index they read.

Issue #606, opened 2026-07-31
Text value templates hide XPath from every linter in `src/attributes.js`


maxonfjvipon:
`expressionsOf` in `src/attributes.js:51` gathers every expression a stylesheet
carries from `//*/@*` and from nowhere else. In XSLT 3.0 that is no longer the
whole set: a text value template puts an XPath expression in a *text node*, so
with `expand-text="yes"` in scope, `<a>{namespace::*}</a>` evaluates its braces
exactly as a `@select` would. `grep -rn expand-text src/` returns nothing, so no
linter has ever been handed those expressions.

```js
const expressionsOf = function(xsl) {
  const bare = new Set(nodes(xsl, SELECTOR))
  return nodes(xsl, '//*/@*').flatMap((attribute) =>
```

Verified against `using-namespace-axis`, which fires on 2.0 and 3.0 alike. This
stylesheet yields zero defects:

```xsl
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
  version="3.0" expand-text="yes">
  <xsl:template match="/"><a>{namespace::*}</a></xsl:template>
</xsl:stylesheet>
```

Move the same axis into an `<xsl:value-of select="namespace::*"/>` in the same
file and it reports one. A valid XSLT 3.0 stylesheet using a deprecated
construct goes unreported, and the same hole swallows every other code-based
format linter, since `count-linter`, `name-linter`, `node-set-linter`,
`translate-linter`, `redundant-boolean-call-linter`,
`redundant-double-negation-linter`, `predicate-position-linter`,
`string-length-linter`, `xpath-axis-linter`, and
`using-namespace-axis-linter` all read their input through `expressionsOf`.
Text value templates are the 3.0 idiom that displaces `xsl:value-of`, so the
more modern the stylesheet, the less of it xslint can see.

The 3.0 shadow attribute is the same miss in a smaller corner: `_select` is the
static-expression spelling of `@select` on an XSLT element, and it is dropped
twice over — `SELECTOR` names `select`, not `_select`, so `bare` never matches
it, and with no braces in the value `enclosed` yields nothing either.

Teach `expressionsOf` about text nodes: in a 3.0 stylesheet, yield each
`enclosed` expression of a text node whose nearest expansion setting is on —
`expand-text` on an ancestor XSLT element, `xsl:expand-text` on a literal result
element, nearest wins. The `enclosed` lexer in `src/expressions.js:83` already
[...]

PR #653, opened 2026-08-04
bug(#651): offer no declarative fix on a refused expression

### LINKED ISSUE #651: A declarative fixer rewrites an expression the validator refused, bypassing #636's gate
#636 stops a code-based linter offering a fix inside an expression the engine
cannot parse, and it does it in `defect` at `src/checks.js`, which is the one
place those linters build a defect through. The declarative half does not go
through `defect`. `src/xpath-linter.js` loads `checks/xpath/*.yaml` and attaches
whatever `src/fixers.js` maps the check name to, so a declarative fix is offered
on a node whether its expression parses or not.

One stylesheet shows both halves disagreeing inside a single run:

```xsl
<xsl:template match="//child::">
  <xsl:value-of select="//child::"/>
</xsl:template>
```

The run reports `invalid-xpath-expression` on the `select`, and reports
`unabbreviated-axis` on both without a fix, which is #636 working. In the same
run, plain `--fix` rewrites the pattern:

```text
-  <xsl:template match="//child::">
+  <xsl:template match="child::">
```

and `--fix-suggestions` rewrites the expression the run has just called
malformed:

```text
-    <xsl:value-of select="//child::"/>
+    <xsl:value-of select=".//child::"/>
```

So `unabbreviated-axis` refuses to touch the text while `starts-with-double-slash`
and `select-starts-with-double-slash` rewrite it, and the second of those edits
text reported as broken in the same breath. The file is left holding the original
syntax error with a different shape around it, which is the failure #636 set out
to end.

The same one-line gate does not port over. A declarative fixer takes the node the
selector matched, and that node need not carry an expression at all — a fix on a
stylesheet root, or on an element rather than an attribute, has nothing to hand
`isValid`. The gate therefore has to ask first whether the node holds an
expression — an attribute named in `ATTRIBUTES` on an XSLT element — and only
then whether it parses, leaving every other fixer alone. Attribute value
templates need care in the same breath, since their whole value is not an
expression and would fail the test for the wrong reason.

Until that lands, the honest statement of the behaviour is that the withholding
covers the code-based checks, which is how `README.md` and `CLAUDE.md` now word
it.

maxonfjvipon:
Closes #651, which is the half of #636 that #636 did not reach.

#636 put the gate in `defect` at `src/checks.js`, the one place every code-based
linter builds a defect through. A declarative fix never goes there:
`src/xpath-linter.js` attaches whatever `src/fixers.js` maps the check name to,
so it was still offered on text the same run had reported malformed. Both halves
of that showed on one stylesheet:

```text
$ node src/index.mjs d.xsl
[ERROR]   d.xsl(9:26) An XPath expression is malformed ... (invalid-xpath-expression)
[WARNING] d.xsl(9:5)  ... (select-starts-with-double-slash)
[WARNING] d.xsl(8:26) ... (unabbreviated-axis)      <- no fix, #636 working

$ node src/index.mjs d.xsl --fix                    # plain, no suggestions
-  <xsl:template match="//child::">
+  <xsl:template match="child::">

$ node src/index.mjs d.xsl --fix-suggestions
-    <xsl:value-of select="//child::"/>
+    <xsl:value-of select=".//child::"/>
```

So within one run `unabbreviated-axis` refused to touch the text while
`starts-with-double-slash` rewrote it, and the second edit rewrote an expression
the run had just called malformed.

The gate now sits where those fixes are attached, and asks the same question
against the same derivation — `expressionsOf`, so there is one answer to what a
stylesheet carries rather than a second opinion:

```js
const fix = FIXERS[pack.name] && !refused(xsl).has(node) &&
  FIXERS[pack.name](node)
```

`refused` collects every attribute whose expression the engine cannot parse
**and the element carrying it**. The element is there because a declarative rule
selects either one — `starts-with-double-slash` matches the `xsl:template` and
its fixer then reaches inside for the `@match` — so the node the gate sees is
often not the node the fix will touch, and a fixer names its own target where no
gate can read it.

That is deliberate over-reach, and worth saying plainly: an element with two
expressions, one broken, loses the fix on the sound one too. An element holding
an expression no processor parses is not worth tidying, and the cost of being
wrong in that direction is a correction that waits for the syntax to be fixed,
against a rewrite of text the run reported malformed. The alternative — matching
the fix's own target back to an attribute — couples the gate to each fixer's
column arithmetic, which is worse.

The set is derived once per document and held in a `WeakMap`, the way
`expressionsOf` holds its own, so a stylesheet with many declarative defects
pays for it once and releases it with the corpus.

The fixture carries both node kinds the scan yields, because 100% branch
coverage counts a branch one kind reaches as covered for all (#606): the
attribute case, and a 3.0 text value template whose `{1 +}` puts the *text*
node's parent element out of bounds. Without the second, `held.node.parentNode`
never runs.

End to end, on a stylesheet mixing the broken with the sound:

```text
match="//child::"        refused   untouched
select="count(a) = 0 ("  refused   untouched
select="not(not(b))"     parses    rewritten to boolean(b)
xmlns:unused="urn:zz"    n/a       removed
```

The last row is the point of the wording in `README.md`. Three fixers —
`namespace-linter`, `result-namespace-linter` and `import-linter` — attach a fix
outside `defect` and outside this gate, but none of them rewrites expression
text; they delete a namespace declaration, edit `exclude-result-prefixes`, or
drop an import. So the claim the README can now make is that an expression
xslint cannot parse is never rewritten, which #636's round of review correctly
refused to let it make on the code-based half alone. `CLAUDE.md` records where
the gate lives and why it takes the element with the attribute.

`npm test` 775 passing, `npm run coverage` 100% of statements, branches,
functions and lines, `test/xcop.test.js` 250 passing with the tool on PATH, and
the new fixture passes `xcop` on its own.

---

**Measured**, since #633 makes this the one area where a number belongs in the record. A generated sheet of 3203 lines, 2400 expressions and 800 declarative fixable defects, best of four runs:

```text
                              master      branch
800 fixable declarative        2972 ms     3071 ms     +3.3%
none fixable                   2960 ms     2932 ms     parity
```

The second row is the short-circuit: `refused` is never built unless a declarative fix is about to be attached, so a stylesheet with none pays nothing.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -346,10 +346,14 @@ the harness asserts too.
   reads an expression the XPath validator refused — it takes the whole corpus,
   not the validated part — so it still reports what it finds there, but `defect`
   withholds the fix, because rewriting text no processor parses is how
-  `select="child::"` became `select=""` (#636). That gate lives in `defect`, so
-  it covers the code-based linters only: a declarative fix from `src/fixers.js`
-  is attached in `src/xpath-linter.js`, never passes through `defect`, and is
-  still offered on an expression that does not parse (#651). A *safe* fix
+  `select="child::"` became `select=""` (#636). A declarative fix never passes
+  through `defect` — `src/xpath-linter.js` attaches it from `src/fixers.js` — so
+  it is gated there instead, against the same `expressionsOf` derivation: no fix
+  is offered on an attribute whose expression the engine refuses, nor on the
+  element carrying it, since a fixer names the attribute it wants inside itself
+  where no gate can see it (#651). Withholding every fix on such an element is
+  deliberate over-reach: an element holding an expression no processor parses is
+  not worth tidying. A *safe* fix
   (deterministic, semantics-preserving) is applied by `--fix`; a
   `suggestion: true` fix (changes behavior, or is one of several corrections) is
   applied only by `--fix-suggestions`. `--fix-dry-run` writes nothing.
@@ -372,7 +376,7 @@ the harness asserts too.
 | `src/reporters.js` | `reporterOf(format)` — `text`, `json`, `sarif`, or `github` output |
 | `src/xsl-validator.js` | Builds the corpus; reports each non-well-formed stylesheet |
 | `src/xpath-validator.js` | Splits corpus expressions into valid (kept) and malformed (reported) via `isValid` |
-| `src/xpath-linter.js` | Loads `checks/xpath/*.yaml`; attaches any `src/fixers.js` fix |
+| `src/xpath-linter.js` | Loads `checks/xpath/*.yaml`; attaches any `src/fixers.js` fix, unless the node — or the element carrying it — holds an expression the engine refuses (#651) |
 | `src/corpus-linter.js` | Loads `checks/corpus/*.yaml`; cross-file rules |
 | `src/*-linter.js` | Code-based `checks/format/*.yaml`, one construct each (axis, namespace, count, name, ...); see the flow diagram |
 | `src/checks.js` | Shared for code-based linters: `metaOf`, `suppressed`, `defect(check, meta, source, node, offset, expression, fix)` — walks the raw text so a wrapped or entity-shifted value reports where it truly stands (#611), and drops the `fix` when `expression` does not parse (#636) |
--- a/README.md
+++ b/README.md
@@ -345,12 +345,20 @@ Checks whose correction needs real judgment (a fresh name, a more specific
 path) stay report-only. A run without `--fix` reports how many defects each
 option would fix.
 
-An expression xslint cannot parse is never rewritten. It is reported as
-`invalid-xpath-expression`, and the other defects standing in it are reported
-too, with no correction offered — including the ones that would rewrite the
-attribute from the outside rather than the expression within it. A stylesheet
-whose real fault is a missing bracket comes back with that fault untouched, so
-fix the syntax and the corrections appear on the next run.
+An expression xslint cannot parse is never rewritten. The other defects standing
+in it are still reported, with no correction offered — including the ones that
+would rewrite the attribute from the outside rather than the expression within
+it. A stylesheet whose real fault is a missing bracket comes back with that
+fault untouched, so fix the syntax and the corrections appear on the next run.
+
+Where the expression sits decides whether you are told why. A bare one — a
+`select`, a `test`, and the rest listed under `invalid-xpath-expression` — is
+reported as malformed. A **pattern** such as `match`, and the braces of an
+attribute or text value template, are not yet parsed by that check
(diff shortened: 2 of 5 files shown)


maxonfjvipon:
### Round 1 — fixed 2, rebutted 1

`add6311`.

**Finding 1 — the over-reach is wider than the docblock's reason.** Half right,
and the half that is right is the half that matters: the docblock explained
`ownerElement` and left `parentNode` unexplained. Fixed.

The premise underneath it is wrong, though, and worth correcting because it
argues for deleting a line that is load-bearing. `text-outside-xsl-text` does
not target the text node. Its selector matches the **element**:

```text
//xsl:*[text()[normalize-space()] and not(local-name() = (...))]
```

and `textOutsideXslText` then reaches *down* into the child text node and
rewrites it whole. So the node the gate sees is the element, and having the text
node in the set does nothing for it. Removing `parentNode` and re-running the
fixture shows what comes back:

```text
starts-with-double-slash   FIX -> match="gamma"
text-outside-xsl-text      FIX -> <xsl:text>delta {1 +} epsilon</xsl:text>
```

That second line is the fix rewriting the very text that holds the unparsable
brace. `parentNode` is what stops it, and it is the exact mirror of the attribute
case: a rule selects an element, the fix lands somewhere inside, and the fixer
names its own target where no gate can read it — sideways into an attribute for
`starts-with-double-slash`, downward into text for `text-outside-xsl-text`. The
docblock now says both, and names this reproduction.

The cost you point at is real and stays: line 14's `starts-with-double-slash` on
the sound `match="//gamma"` is withheld because a sibling text node holds
`{1 +}`. Same trade as the attribute case, same direction, pinned by the same
test.

**Finding 2 — a broken pattern or TVT withholds silently.** Agreed, and this was
the better of the two findings, because it is the user-visible one. `README.md`
now says where the explanation comes from and where it does not: a bare
expression is reported as malformed, while a pattern or a template's braces are
not parsed by that check at all (#589), so a broken one is silent and the
corrections around it simply disappear. It closes with the practical tell — a
defect reporting without a correction where you expected one means to check that
expression's syntax by hand. Two sentences, and the paragraph is now as true as
its first line.

**Finding 3 — the missing measurement.** Fair, and taken. My own numbers, on a
generated sheet of the same shape (3203 lines, 2400 expressions, 800 declarative
fixable defects), best of four runs to keep the noise out:

```text
                              master      branch
800 fixable declarative        2972 ms     3071 ms     +3.3%
none fixable                   2960 ms     2932 ms     parity
```

Smaller than your +9%, same direction and the same shape — different machines,
and my first two-run pass swung between +5% and −1.5%, which is why I went to
best-of-four. The second row is the short-circuit you predicted: `refused` is
never built unless a declarative fix is actually about to be attached, so a
stylesheet with none pays nothing. Both rows are in the pull-request body now,
since you are right that #633 makes this the one area where a number belongs in
the record rather than an adjective.

Issue #654, opened 2026-08-04
`xsl:use-when` on a literal result element is read by neither the validator nor any linter


maxonfjvipon:
`use-when` is a static attribute, and XSLT spells it twice: unprefixed on an
XSLT element, and namespaced as `xsl:use-when` on a literal result element,
which is the only way a simplified stylesheet can carry one. #647 added the
unprefixed spelling to the validator. The namespaced one is read by nothing.

Both gates key on the bare name and demand an XSLT owner. `held` at
`src/xpath-validator.js:56`:

```js
return node.nodeType === 2 && EXPRESSIONS.includes(node.localName) &&
  node.ownerElement.namespaceURI === XSLT
```

and the filter behind `expressionsOf` at `src/attributes.js:159` asks
`NAMED.has(one.nodeName)` with the same owner test. On `xsl:use-when` the
`localName` is `use-when` but the owner is the literal result element, whose
namespace is not the XSLT one, so the first refuses it; `nodeName` is
`xsl:use-when`, which is in neither list, so the second refuses it too.

```xsl
<xsl:template match="/">
  <html xsl:use-when="1 +">text</html>
</xsl:template>
```

reports nothing. The expression is malformed and evaluated at compile time, so a
processor rejects the stylesheet before it transforms anything — the same fault
`invalid-xpath-expression` reports one line up when the attribute sits on an
`xsl:variable` instead.

It is worse than a missing report: because `expressionsOf` does not yield it
either, no code-based linter reads it at all, so an axis or a `count(...) = 0`
inside an `xsl:use-when` is invisible to every check that would otherwise find
it. That makes this a false negative across the whole format tier, not only in
the validator.

The fix is one rule rather than one name: an attribute in the XSLT namespace on
a *non*-XSLT element is the shadow of the same attribute unprefixed, so both
lists should test `localName` against an XSLT-namespaced attribute as well as an
unprefixed attribute of an XSLT element. `xsl:version` and `xsl:expand-text` are
already read that way in `src/xsl-version.js` and `src/attributes.js`, so the
shape exists in the tree; it is the expression lists that never learned it.
Worth checking `xsl:exclude-result-prefixes` and the other static attributes in
the same pass.

maxonfjvipon:
The mirror case belongs here too, found while reviewing #655: `held` keys on
`node.localName` and never tests the **attribute's own** namespace, so a vendor
extension attribute on an XSLT element is read as XPath.

```text
ext:select      reported as invalid-xpath-expression
[...]

PR #655, opened 2026-08-04
bug(#647): validate every attribute that holds an expression

### LINKED ISSUE #647: `use-when`, `for-each-item` and `for-each-source` are never validated in `src/xpath-validator.js`
`EXPRESSIONS` at `src/xpath-validator.js:40` names the attributes whose XPath is
validated, and it names twelve:

```js
const EXPRESSIONS = [
  'select', 'test', 'use', 'value', 'group-by', 'group-adjacent',
  'key', 'initial-value', 'xpath', 'context-item', 'with-params',
  'namespace-context',
]
```

`ATTRIBUTES` at `src/attributes.js:19`, which is what every code-based linter
reads its expressions from, names twenty. Five of the extra eight are the
patterns #589 owns. The other three — `for-each-item`, `for-each-source` and
`use-when` — are ordinary XPath expressions, and nothing validates them:

```text
malformed expression "1 +" placed in each attribute a linter reads:
   select        REPORTED
   test          REPORTED
   use           REPORTED
   group-by      REPORTED
   for-each-item   SILENT
   for-each-source SILENT
   use-when        SILENT
```

`<xsl:for-each use-when="1 +"/>` therefore passes a default run with no defect
at all, while the identical text one attribute over is an error. `use-when` is
the one that bites in practice: it is evaluated at compile time, so a processor
rejects the stylesheet before it transforms anything, which is exactly the class
of fault `invalid-xpath-expression` exists to catch early.

The gap is not only a missing report. Every code-based linter reads these three
through `expressionsOf` and reasons over them whether they parse or not, so a
malformed `use-when` is unreported text that the checks still scan.

Adding the three names to `EXPRESSIONS` is the whole fix, and it is separate
from #589: those five are patterns and need a pattern grammar before anything
can judge them, while these three are expressions the existing `isValid` already
knows how to read. The motive for `invalid-xpath-expression` lists the covered
attributes by name and would gain them in the same change.

maxonfjvipon:
Closes #647.

`EXPRESSIONS` at `src/xpath-validator.js:40` named twelve attributes.
`ATTRIBUTES` at `src/attributes.js:19`, which is what every code-based linter
reads its expressions from, names twenty. Five of the extra eight are the
patterns #589 owns. The other three hold ordinary XPath and nothing looked at
them:

```text
malformed expression placed in each attribute a linter reads
                          before      after
select                    REPORTED    REPORTED
test                      REPORTED    REPORTED
use, group-by, ...        REPORTED    REPORTED
for-each-item             SILENT      REPORTED
for-each-source           SILENT      REPORTED
use-when                  SILENT      REPORTED
```

Adding the three names is the whole diff, but not the whole effect: the
validated set also feeds `EXPRESSION_LINTERS`, so `redundant-whitespace` now
reports and fixes inside all three as well —
`use-when="count( a )  =  1"` draws two defects where the branch previously
found none. That is consistent with the code-based tier, which already read
these three through `expressionsOf`.

The list is derived rather than written out, so the drift cannot recur: ```js
const EXPRESSIONS = ATTRIBUTES.filter((name) => !PATTERNS.includes(name))
```

`ATTRIBUTES` is "expressions and patterns" and `PATTERNS` is "the patterns", so
the remainder is exactly what this validator is for — verified as the same
fifteen names in the same order, not a numeric coincidence. #627 added
`for-each-item` and `for-each-source` to one list only, and #647 is what came of
it; there is now no second list to drift.

`for-each-item` and `for-each-source` are the two an `xsl:merge-source` carries beside its `select`,
which #627 already recognised as expressions on the linter side; `use-when` is
static, so a malformed one stops a processor before it transforms anything,
which is precisely the fault this check exists to surface early.

The fixture puts each in the element that really carries it — an
`xsl:merge-source` inside an `xsl:merge` for the first two — rather than
inventing a host, so it reads as XSLT someone might write.

Two things deliberately left out.

**The literal result element spelling.** `use-when` is written `xsl:use-when` on
a literal result element, which is the only way a simplified stylesheet can carry
one, and both lists key on the bare name with an XSLT owner, so the namespaced
form is refused by each. It is read by no linter either, since `expressionsOf`
filters the same way — so it is a false negative across the whole format tier,
not just a missing report. Filed as **#654** with the one-rule fix, because it is
a different mechanism from a missing name and touches `src/attributes.js` too.

**A pattern is still unvalidated**, which is #589 and unchanged here.

`npm test` 776 passing, `npm run coverage` 100% of statements, branches,
functions and lines, and the new fixture passes `xcop` on its own. The motive and
`README.md` both list the covered attributes by name, so both gained the three.

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,8 +401,8 @@ Validators:
 - **XPath syntax** — every bare XPath expression (in `select`, `test`,
   `use`, `value`, `group-by`, `group-adjacent`, and the XSLT 3.0 `key`,
   `initial-value`, `xpath`, `context-item`, `with-params`,
-  `namespace-context`) is parsed; the ones the processor cannot parse are
-  reported.
+  `namespace-context`, `for-each-item`, `for-each-source` and `use-when`) is
+  parsed; the ones the processor cannot parse are reported.
 
 Linters:
 
--- a/src/attributes.js
+++ b/src/attributes.js
@@ -168,6 +168,7 @@ const expressionsOf = function(xsl) {
 
 module.exports = {
   ATTRIBUTES,
+  PATTERNS,
   selectorOf,
   expressionsOf,
 }
--- a/src/resources/motives/validation/invalid-xpath-expression.md
+++ b/src/resources/motives/validation/invalid-xpath-expression.md
@@ -4,8 +4,9 @@ An attribute carrying a bare XPath expression — `select`, `test`, `use`,
 `value`, `group-by`, `group-adjacent`, or the XSLT 3.0 `key`, `initial-value`,
 `xpath`, `context-item`, `with-params`, `namespace-context`, the
 `for-each-item` and `for-each-source` of an `xsl:merge-source`, and the static
-`use-when` — must hold an expression the processor can parse. A malformed expression breaks the
-transformation at runtime, so the sooner it surfaces the better. Only the
+`use-when` — must hold an expression the processor can parse. A malformed
+expression breaks the transformation at runtime, so the sooner it surfaces the
+better. Only the
 expression syntax is checked, not its formatting; pattern attributes such as
 `match` and attribute value templates are left to other checks.
 
--- a/src/resources/motives/validation/invalid-xpath-expression.md
+++ b/src/resources/motives/validation/invalid-xpath-expression.md
@@ -2,8 +2,9 @@
 
 An attribute carrying a bare XPath expression — `select`, `test`, `use`,
 `value`, `group-by`, `group-adjacent`, or the XSLT 3.0 `key`, `initial-value`,
-`xpath`, `context-item`, `with-params`, `namespace-context` — must hold an
-expression the processor can parse. A malformed expression breaks the
+`xpath`, `context-item`, `with-params`, `namespace-context`, the
+`for-each-item` and `for-each-source` of an `xsl:merge-source`, and the static
+`use-when` — must hold an expression the processor can parse. A malformed expression breaks the
 transformation at runtime, so the sooner it surfaces the better. Only the
 expression syntax is checked, not its formatting; pattern attributes such as
 `match` and attribute value templates are left to other checks.
--- a/src/xpath-validator.js
+++ b/src/xpath-validator.js
@@ -40,7 +40,7 @@ const names = [CHECK]
 const EXPRESSIONS = [
   'select', 'test', 'use', 'value', 'group-by', 'group-adjacent',
   'key', 'initial-value', 'xpath', 'context-item', 'with-params',
-  'namespace-context',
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by maxonfjvipon** [comment]
### Round 1 — fixed 3, deferred 1, rebutted 0

`2cf9cdc`.

**The invariant.** Taken, and it is the better shape. I checked the equality
before trusting it — `ATTRIBUTES` minus `PATTERNS` is not merely fifteen names,
it is the *same* fifteen in the same order, with an empty difference both ways.
So the derivation is not a numeric coincidence that would rot: `ATTRIBUTES` is
"expressions and patterns", `PATTERNS` is "the patterns", and the remainder is
exactly what this validator is for.

```js
const EXPRESSIONS = ATTRIBUTES.filter((name) => !PATTERNS.includes(name))
```

`PATTERNS` is exported, and `src/attributes.js` requires only `expressions`,
`xsl-version` and `tree`, so nothing cycles back. The docblock now carries the
reason rather than the list, and names #627 as the drift that produced #647 —
the point being that a second list is a second opinion about what a stylesheet
carries, which is the same argument #653 accepted for the declarative gate.
CLAUDE.md asks for the check in the same change as the fix, and you are right
that this is that case; the derivation is stronger than the assertion I would
otherwise have added, because there is no longer a second list to assert
*about*.

**The consequence I undersold.** Confirmed, and worth the correction: the
validated set feeds `EXPRESSION_LINTERS`, so `redundant-whitespace` now reports
and fixes inside all three attributes.

```text
-  <xsl:variable name="alpha" use-when="count( a )  =  1" select="b"/>
+  <xsl:variable name="alpha" use-when="count( a ) = 1" select="b"/>
```

Two defects, both with fixes, where the branch previously found none. That is
the behaviour I want — consistent with the code-based tier, which already read
these three through `expressionsOf` — but "adding three names is the whole
change" described the diff and not the effect. The body now carries both
columns. Agreed that `README.md` needs nothing: its `redundant-whitespace` entry
does not enumerate attributes.

**The namespace finding.** Reproduced, including your point that it predates
this branch:

```text
ext:select      reported as invalid-xpath-expression
ext:use-when    reported as invalid-xpath-expression
ext:whatever    ignored
```

Folded into #654 as you suggest, with the reasoning that the two are one bug:
this issue is an attribute the check should read and does not because the
*owner* is not XSLT, and that comment is one it should not read and does because
the *attribute's own* namespace is never asked about. One `namespaceURI` test
closes both directions, and it also settles the `localName`-versus-`nodeName`
split between the two lists that the derivation above has now made visible.
Deliberately not here: it changes `src/attributes.js` too, and this branch is a
list.

**The nit.** Fixed — line 7 was mine, and it now wraps with the rest of the file.

Issue #668, opened 2026-08-05
Nothing reports an `xsl:import`/`xsl:include` with no `@href`


maxonfjvipon:
XSLT makes `@href` mandatory on `xsl:import` and `xsl:include` — without it the
element names no module and there is nothing to pull in. xslint says nothing
about it. A sheet holding

```xsl
<xsl:stylesheet version="2.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
  <xsl:import/>
  <xsl:template match="/">
    <xsl:value-of select="."/>
  </xsl:template>
</xsl:stylesheet>
```

draws `not-using-output` and nothing else. No check mentions `href` at all —
`grep -rl href src/resources/checks/` comes back empty — so the one element
whose whole purpose is an attribute is unchecked for it, and the author learns
about the mistake from their processor rather than from the linter.

This is the gap #597 leaves behind. That ticket is about the crash the same
element caused: `importsOf` in `src/import-graph.js` joined the absent href onto
a directory, threw, and took the run's whole report down. Skipping an
href-less reference there is right — resolving a module nobody named is not that
module's job — but it settles the crash by making the malformed element
invisible, and nothing downstream picks it up.

#209 does not cover it either. Its two checks are a `href` that resolves to no
file and a `href` whose target is not a stylesheet; both start from an href that
exists. A missing one is a required-attribute fault, decidable from the document
alone, needing neither `fs` nor path resolution.

One declarative rule closes it, in the shape the repository already uses for a
required-attribute fault such as `mode-or-priority-without-match`: a
`checks/xpath/` entry selecting `(//xsl:import | //xsl:include)[not(@href)]`, at
`error`, with its motive and pack. It is report-only by nature — the module the
author meant is not recoverable from the document, so there is no correction to
offer.

**Rescoped in place, 2026-08-27: an empty `@href` belongs in the same rule.**
The fault is identical and so is the repair, and an empty one is not the milder
case: a relative reference of no characters resolves to the module itself, which
SaxonJ-HE 12.5 refuses as `XTSE0210` importing and `XTSE0180` including, where
an absent one is `XTSE0010`. It is also the spelling that does harm beyond going
unreported. `href=""` resolves outside the linted set, which #468's guardrail
reads as a module external to the run and therefore fine, so `not-using-output`
excuses the very file that wrote it: `<xsl:import href=""/>` in a stylesheet
with no `xsl:output` draws nothing at all, where the same file without the
[...]

Issue #667, opened 2026-08-05
`redundant-import` offers a safe fix that moves import precedence in `src/import-linter.js`


maxonfjvipon:
`redundant-import` hands a repeated `xsl:import` a **safe** deletion, and
`removal` in `src/import-linter.js:111` says why in its own words:

```js
 * duplicate is semantics-preserving — the module stays imported by the first
 * reference — so it is a safe fix, not a suggestion.
```

The module does stay imported. What does not stay is its import precedence,
which is positional: XSLT 1.0 §2.6.2 ranks an `xsl:import` above every import
declared before it, so the *second* reference to one module is not a copy of the
first — it sits at a different level, and the level is what decides which
definition wins.

Verified with xsltproc, on two libraries that both define
`<xsl:template match="zz">`, one writing `A` and the other `B`:

```xsl
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
  <xsl:import href="lib-a.xsl"/>
  <xsl:import href="lib-b.xsl"/>
  <xsl:import href="lib-a.xsl"/>
</xsl:stylesheet>
```

It outputs `A`, because the third import puts lib-a above lib-b. `--fix` deletes
that third import, and the same input then outputs `B`. A safe fix changed the
stylesheet's output, which is the one thing the safe tier promises it will not
do.

The defect needs a third module between the two references, so the packs in
`test/resources/import-packs/` do not reach it — both duplicate fixtures repeat
one module with nothing in between, where the deletion genuinely is
semantics-preserving.

The smallest correction is to keep the *earlier* reference's position: delete the
first of the two rather than the last, and the surviving copy of lib-a stays
where the higher-precedence one stood. Failing that, withhold the safe fix
whenever an import of a different module stands between the two references, the
way #597 withholds it for a module reached by both `xsl:import` and
`xsl:include`, whose two mechanisms differ in precedence the same way this one's
two positions do.

PR #669, opened 2026-08-05
Skip an href-less import, and withhold a fix that crosses import precedence

### LINKED ISSUE #597: `xsl:import`/`xsl:include` without `@href` crashes the import linter in `src/import-graph.js`
A stylesheet with an `xsl:import` or `xsl:include` that carries no `@href` crashes the whole lint run. Such an element is well-formed XML — invalid XSLT, but nothing rejects it before the import linter — so it flows into `importsOf` in `src/import-graph.js`, where `target` joins a null onto the directory:

```js
const target = function(file, href) {
  return path.normalize(path.join(path.dirname(file), href))
}
```

`node.getAttribute('href')` returns `null` for the missing attribute, and `path.join(dirname, null)` throws `The "path" argument must be of type string. Received null`. Running the linter over a sheet holding `<xsl:import/>` reproduces it. The fix is to treat an href-less import as external and skip it — default the attribute to `''`, or drop edges whose href is absent — so a malformed import is reported by whatever validates XSLT, not turned into an exception that takes the rest of the report down with it.

The same check also lumps `xsl:import` and `xsl:include` of one module together. `byRedundancy` in `src/import-linter.js` keys duplicates on `file|to` alone, so a sheet that imports `c.xsl` and also includes it is flagged as a redundant import and handed a delete fix. But import and include differ in precedence — an imported module sits below the importing sheet's own definitions while an included one sits at the same level — so deleting the second reference is not semantics-preserving, and the safe-fix tier is wrong for that pair. Either exclude a cross-mechanism repeat from `redundant-import`, or withhold the safe fix when the duplicate crosses import and include. Both faults must be settled before the check can be called mature.

maxonfjvipon:
Closes #597.

Two faults, both in the import pair.

**An href-less reference took the whole report down.** `<xsl:import/>` is
well-formed XML, so nothing stopped it before `importsOf`, where its absent
`@href` was joined onto the declaring file's directory:
`The "path" argument must be of type string. Received null`, and the run ended
with nothing reported about any file. Such a reference names no module, so it
now yields no import at all:

```js
element.namespaceURI === XSLT &&
(element.localName === 'import' || element.localName === 'include') &&
element.hasAttribute('href'),
```

Skipping is the whole fix, not a workaround: resolving a module nobody named is
not this module's job. Nothing reports that malformed element either, which is
a real gap but a separate one — filed as #668 rather than answered here, since a
required-attribute check is a rule with its own YAML, motive and pack.

The pack pins what the crash cost rather than only that it is gone: a file
carrying two href-less imports, an href-less include, and one real import into a
cycle. The cycle is still reported, twice, so the malformed element no longer
hides the feedback on everything around it. Two href-less imports rather than
one is deliberate — defaulting the attribute to `''`, the other fix #597
suggests, resolves both to the same directory and would have invented a
`redundant-import` between them.

**A module reached both ways was treated as one written twice.** `byRedundancy`
keyed a duplicate on `file|target` alone, so a stylesheet that imports `c.xsl`
and also includes it was handed a **safe** deletion. The two mechanisms differ
in import precedence — an included module's definitions stand at the level of
the including sheet's own, an imported module's below them — so deleting either
moves definitions between levels. Since XSLT requires every `xsl:import` to
come first, the duplicate reported is always the include, which makes the
offered edit always the worse of the two available corrections.

`crossed` now collects the modules a sheet reaches both ways, and those are
reported without a fix — not demoted to a suggestion, because which reference to
drop is a decision about precedence rather than one of several spellings of one
correction. Verified through the CLI: `--fix` and `--fix-suggestions` both leave
such a sheet byte-identical, while two `xsl:include` of one module still lose
their duplicate. The rule is per module rather than per pair, so a third
reference in a mixed group withholds too, and `repeated-include.yaml` guards the
same-mechanism fix that stays.

The motive gained the precedence teaching and lost a claim that was false for
this case — "the second reference brings in the same templates ... adds nothing"
is true of two imports, not of an import beside an include.

**One defect found while doing it, filed rather than folded in: #667.** The safe
tier is wrong for a shape this ticket does not name either. Import precedence is
positional, so with `lib-a`, `lib-b`, `lib-a` imported in that order, where both
libraries define `match="zz"`, xsltproc outputs `A`; after `--fix` deletes the
third import it outputs `B`. A safe fix changed the stylesheet's output. It
needs a third module between the two references, which no fixture has, so it
sits outside this change.

`npm test` 1050 passing and 0 failing, `npm run coverage` the same at 100% of
statements, branches, functions and lines, `npx mocha test/xcop.test.js` 256
passing with the tool present, `npx eslint .` clean, and yamllint, markdownlint
and typos clean over what changed. `docs/` is generated and untracked, so the
motive change reaches the site through `npx grunt docs`, which was run.


maxonfjvipon:
Three things before this goes in. The change itself is right — both faults reproduce and every decision in it is pinned (details at the bottom).

**1. It no longer merges: `CLAUDE.md` conflicts with master.**

#666 landed after this branch was cut and rewrote the `src/tokens.js` row; this PR rewrites the `src/import-graph.js` row directly below it. The eighteen green checks all ran against the pre-#666 base, so they say nothing about the merged state. The resolution is to keep both rows. Merged locally that way it is clean: 1052 passing, 100% of statements, branches, functions and lines, `npx eslint .` clean, `npx mocha test/xcop.test.js --forbid-pending` 256 passing.

**2. `removal`'s docblock still claims the fix is semantics-preserving, which #667 disproves.**

The sentence this PR edits reads:

> Deleting a duplicate is semantics-preserving — the module stays imported by the first reference — so it is a safe fix, not a suggestion. It is offered only where every reference to that module uses one mechanism, which is what `crossed` decides.

The appended qualifier is about mechanisms only, so the first clause is left standing. It is false in the positional case: with `liba`, `libb`, `liba` imported in that order and both libraries defining `match="zz"`, xsltproc outputs `A`, and after `--fix` deletes the third import it outputs `B`. Deferring that to #667 is the right call — it needs a third module no fixture has — but the claim should not read as unconditional while the counterexample is known. `importsOf` cites #668 inline for the gap deferred there; the same reference here would stop the next reader trusting it.

**3. The message still gives a one-step instruction for the shape that now needs a choice.**

`redundant-import.yaml` is unchanged and reads "Remove the duplicate xsl:import/xsl:include, keeping a single reference". For a module reached both ways the reported node is always the `xsl:include`, since every `xsl:import` must come first — and the motive this PR writes says that include may be the reference to keep. One message serves both shapes, so this is either a small rewording or nothing, but as it stands the message points at the deletion the motive argues against.

---

Everything else checks out, verified rather than read:

- **The crash costs the whole report.** On master a lone `<xsl:import/>` throws out of `importsOf` and nothing at all is reported about any file, including the unrelated defects in the sibling stylesheet. On this branch both files are reported.
- **The crossed safe fix really did change output.** A sheet importing and also including `common.xsl` gives `L` under xsltproc; after master's safe `--fix` deleted the include it gives `M`. This branch leaves the file byte-identical, so it stays `L`.
- **The tiering behaves as described through the CLI.** Crossed reports with no fix and survives `--fix --fix-suggestions` unchanged; two `xsl:include` still lose their duplicate; `import x` with two `include x` withholds on both, so the rule is per module and not per pair.
- **Every decision is pinned.** Four mutations, all four caught: defaulting `href` to `''` instead of skipping (1 failing), always attaching `removal` (1), never attaching it (5), and neutering `crossed` (1). The two href-less imports in the pack are load-bearing, not decorative — the harness asserts over every defect, so `amount: 2` is what forbids the phantom `redundant-import` an `href=''` default would invent.
- Neither `redundant-import` nor `circular-import` carries `mature: true`, so changing behaviour here breaks no freeze.

maxonfjvipon:
All three stand. Fixed in `0e8e24e` and `8168b1f`.

**1. The conflict was real, and the green checks were worthless.** `mergeable`
was `false`, `mergeable_state` `dirty` — #666 merged at 14:39, forty minutes
after this branch's checks ran. Both rows kept: master's `src/tokens.js` row,
which #666 rewrote, and this branch's `src/import-graph.js` row. The branch is
`mergeable=true` again and the suite re-run on the merged tree: 1052 passing,
100% of statements, branches, functions and lines, `npx mocha test/xcop.test.js
--forbid-pending` 256 passing, `npx eslint .` clean.

**2. The docblock was asserting what I had just filed a counterexample to.**
Appending the mechanism qualifier left the first clause reading unconditionally,
which is the worse of the two failure modes here — a reader who trusts it writes
the next safe fix on the same premise. It now separates where the claim holds
from where it does not:

```js
 * `crossed` decides — the module stays imported by the reference left standing,
 * so the deletion is safe rather than a suggestion. Safe there is not safe
 * everywhere: import precedence is positional, so where an import of another
 * module stands between two imports of this one, cutting the later of the two
 * drops this module below that one and can change the output (#667).
```

The tiering itself is untouched, so #667 stays deferred rather than
half-answered: it needs a third module between the two references, which no
fixture has, and giving it a real fix means deciding whether to cut the earlier
reference instead — a decision worth its own change.

**3. The message pointed at the node the motive may want kept.** It read "Remove
the duplicate xsl:import/xsl:include, keeping a single reference", and the
duplicate is always the `xsl:include`, since every `xsl:import` must come first.
Now: "Keep a single xsl:import/xsl:include of it and remove the other" — one
instruction that is true of both shapes, naming neither node. The precedence
reasoning stays in the motive, where an author has room to read it, rather than
being compressed into a warning line. Nothing asserts the message text, so the
change is the YAML and the regenerated site.

Issue #643, opened 2026-08-03
Six code-based linters read an XPath gap with JS `\s`


maxonfjvipon:
Eleven regexes across six code-based linters spell an XPath gap as JS `\s`, which matches far more
than XPath does. `ExprWhitespace` is `S ::= #x20 | #x9 | #xD | #xA` — four characters — while `\s`
also takes a no-break space, a vertical tab, a form feed, an em space, an ideographic space and
U+FEFF. So each of these reads a call where the source holds no call at all:

```js
const WRAPPER = /^\s*boolean\s*\(/                      // src/redundant-boolean-call-linter.js:35
const CALL = /(^|[^\w:.-])not\s*\(/g                    // src/redundant-double-negation-linter.js:33
const CALL = /(^|[^\w:.-])(name|local-name)\s*\(/g      // src/name-linter.js:35
const CALL = /(^|[^\w:.-])translate\s*\(/g              // src/translate-linter.js:34
const CALL = /[\w.-]+:node-set\s*\(/g                   // src/node-set-linter.js:35
const TAIL = /^\s*(!=|<=|>=|=|<|>)\s*([01])(?![\w.])/   // src/comparisons.js:15
```

`boolean (a)` is not a function call — the engine refuses the whole expression — yet
`redundant-boolean-call` matches it and offers to drop a wrapper that was never there. A stylesheet
whose `test` reads `boolean (a)` and `not (not (b))` draws two
`invalid-xpath-expression` errors and, on the same two expressions, a `redundant-boolean-call` and a
`redundant-double-negation` with fixes attached. Both fixes edit an expression that does not parse,
which is #636 reached by a second route: there the linter sees a rejected expression because the
staging hands it the whole corpus, and here it recognises a construct in it that is not there.

`src/tokens.js:69` already holds the right class as `WHITESPACE = ' \t\r\n'`, and `src/xpath.js`
started sharing it in #640, but only five files import it and none of the six above is among them.
`CLAUDE.md:377` says `src/tokens.js` "owns `WHITESPACE`, the XML `S` every reader of a gap shares",
which describes the intention rather than the tree — that line is being corrected on #640.

Each of these regexes should spell its gap from the shared class instead of `\w`-adjacent shorthand,
so a gap the grammar does not permit stops looking like one; the packs need a row per linter with a
no-break space where the ordinary space sits, since 100% branch coverage counts a branch an ASCII
gap reaches as covered and hides every one of these.

PR #670, opened 2026-08-05
bug(#642): see the gap the axis token swallowed

### LINKED ISSUE #642: `redundant-whitespace` misses the gap in front of an axis's `::` in `src/tokens.js`
`opensAxis` at `src/tokens.js:217` folds the gap in front of an axis's `::` into the axis token itself, and its length says so:

```js
const name = `${xpath.slice(at, end)}::`
return end === at || xpath[colons] !== ':' || ... ?
  null :
  {name: name, length: colons + 2 - at}
```

`colons` has already walked past the whitespace, so `ancestor  ::  b` lexes as one axis token spelling `ancestor::`, then a `whitespace` token for the run *behind* the colons, then the name test. `redundancies` at `src/xpath-format-linter.js:39` only ever inspects `whitespace` tokens, so of the two redundant runs in that expression it sees one.

The consequence is a file the linter fixes and then pronounces clean while a defect it defines is still in it. On `<xsl:value-of select="ancestor  ::  b"/>` the run reports one `redundant-whitespace` warning; `--fix` writes back `select="ancestor  :: b"`; the next run reports no whitespace defect at all. Two spaces in front of the colons are redundant by the check's own definition — a run longer than one space — so the check is under-reporting rather than tolerating. The abbreviable spelling hides it: `child  ::  a` comes out as `a`, because `unabbreviated-axis` owns a span that reaches past both gaps.

The expression only reaches the format linter now that `isValid` accepts a spaced axis (#615). Before that the validator called it `invalid-xpath-expression` and dropped it, which is why the under-report has been unreachable until now, and #628 and #565 record neither half of it.

Either let `redundancies` look inside an axis token for a run longer than one space, or stop folding the leading gap into that token and let the lexer emit it as the `whitespace` token it already is everywhere else.

maxonfjvipon:
Closes #642.

`redundant-whitespace` saw one of the two redundant runs in `ancestor  ::  b`,
and `--fix` then left the other in the file:

```text
before                    1 warning
--fix writes              select="ancestor  :: b"
run again                 no whitespace defect at all
```

Two spaces in front of the colons are redundant by the check's own definition, so
the tool pronounced clean a file holding a defect it defines. After:

```text
before                    2 warnings
--fix writes              select="ancestor :: b"
run again                 clean, and correctly so
```

**The roadmap predicted this would fix itself and it did not**, which is worth
recording: #644 had #642 as "falls out of name-aware axis lexing; verify and close
with a pack row" after #249. Verified against the #666 merge — it does not. #666
changed how a *name* is lexed, while `opensAxis` still folds the gap in front of
`::` into the axis token, so `ancestor  ::` remains one token and a scan reading
only `whitespace` tokens cannot see the run inside it. The line in #644 is
corrected.

**I took the ticket's first option, not the second, having changed my mind on
reading what the second costs.** The ticket offers either letting `redundancies`
look inside an axis token, or stopping the fold and letting the lexer emit the
gap as the `whitespace` token it emits everywhere else. I said I would take the
second. It needs a token type for a bare `::` — there is none — and every consumer
that reads an axis token's value and span to build a fix would move with it, on a
lexer merged hours ago. The first option is local to the check that owns
whitespace, and the suite proves the difference: **not one pack position shifted**,
because the token stream is untouched.

The rule is general rather than an axis list: a token that carries a run of more
than one gap character inside itself reports it, with `STRING` and `COMMENT`
excluded because the lexer keeps those whole on purpose, and a line-wrapping run
left alone as it is everywhere else. Only an axis can hold such a run today —
`instance of` is matched with exactly one space — so the rule needs no list to
maintain and gains nothing to drift.

The gap is spelled with `GAP` from `src/tokens.js`, as #643 requires.

New pack `gap-in-front-of-the-colons.yaml` pins both runs, at 4:35 and 4:39, with
the replacement each carries.

`npm test` 1047 passing and 0 failing, `npm run coverage` 100% of statements,
branches, functions and lines, `npx mocha test/xcop.test.js --forbid-pending` 253
passing, `yamllint` clean on the new pack.

### DIFF
--- a/src/xpath-format-linter.js
+++ b/src/xpath-format-linter.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: MIT
  */
 
-const {tokenized, TOKENS} = require('./tokens')
+const {tokenized, TOKENS, GAP} = require('./tokens')
 const {metaOf, suppressed, defect} = require('./checks')
 const {logger} = require('./logger')
 
@@ -26,6 +26,34 @@ const META = metaOf(CHECK)
 const names = [CHECK]
 
 /**
+ * A run of more than one gap character, for looking inside a token that carries
+ * one.
+ * @type {RegExp}
+ */
+const INTERNAL = new RegExp(`${GAP}{2,}`)
+
+/**
+ * The redundant run a token holds inside itself, or null. Only an axis holds
+ * one: the lexer folds the gap in front of its `::` into the axis token, so
+ * `ancestor  ::` is a single token and a scan reading only `whitespace` tokens
+ * saw the run behind the colons and not the one in front of them — one of the
+ * two runs in `ancestor  ::  b`, with `--fix` leaving the other in the file and
+ * the next run calling it clean (#642). A string or a comment is kept
+ * whole on purpose, so neither is ever looked into, and a run that wraps a line
+ * is left alone here as it is everywhere else.
+ * @param {{type: string, value: string, start: number}} token - The token
+ * @return {?{offset: number, value: string, replacement: string}} - The run
+ */
+const inside = function(token) {
+  const run = token.type === TOKENS.STRING || token.type === TOKENS.COMMENT ?
+    null :
+    INTERNAL.exec(token.value)
+  return run === null || /[\r\n]/.test(run[0]) ?
+    null :
+    {offset: token.start + run.index, value: run[0], replacement: ' '}
+}
+
+/**
  * Redundant whitespace runs in an expression. A run is redundant when it is
  * longer than one space, or leads or trails the expression; a run that wraps a
  * line is left alone, and runs inside string literals or comments are never
@@ -41,6 +69,7 @@ const redundancies = function(expression) {
     const edge =
       token.start === 0 ||
       token.start + token.value.length === expression.length
+    const held = token.type === TOKENS.WHITESPACE ? null : inside(token)
     if (
       token.type === TOKENS.WHITESPACE &&
       !/[\r\n]/.test(token.value) &&
@@ -51,6 +80,8 @@ const redundancies = function(expression) {
         value: token.value,
         replacement: edge ? '' : ' ',
       })
+    } else if (held !== null) {
+      runs.push(held)
     }
   }
   return runs
--- a/test/resources/xpath-format-packs/gap-in-front-of-the-colons.yaml
+++ b/test/resources/xpath-format-packs/gap-in-front-of-the-colons.yaml
@@ -1,0 +1,19 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 Max Trunnikov
+# SPDX-License-Identifier: MIT
+---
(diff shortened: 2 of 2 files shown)


maxonfjvipon:
Good to merge. Two things worth recording, neither blocking.

**1. The `COMMENT` half of the exclusion is unpinned.**

Mutating `|| token.type === TOKENS.COMMENT ?` to `|| false ?` leaves the suite green at 1047 passing. The `STRING` half is pinned — dropping it fails a test — but the comment half is not, and it is load-bearing: `a (: keep  this :)/b` is valid XPath, and without the exclusion `--fix` writes `a (: keep this :)/b`, editing the inside of a comment. A row in the new pack closes it.

**2. The line-wrap guard does not hold for an attribute value, and this widens a pre-existing bug.**

`/[\r\n]/.test(run[0])` never fires for an expression that came from an attribute, because XML normalizes a newline in an attribute value to a space before the linter ever sees it. So for

```xsl
<xsl:value-of select="ancestor
      ::  b"/>
```

the run in front of the colons is reported and announced fixable, but `src/fixer.js` decode-walks the raw source, where those characters are a newline and six spaces rather than seven spaces. Verify-before-apply skips it, and `--fix` converges with the defect still reported and still counted as fixable:

```text
before --fix        2 defects, "2 defects fixable with --fix"
after --fix twice   1 defect,  "1 defects fixable with --fix"   <- never goes away
```

This is not introduced here. Master does the same for a run between tokens — `select="aa and\n      bb"` is reported, announced fixable, and survives `--fix` unchanged. But this branch adds a second occurrence of it at a position that produced none before, so the axis case inherits it.

The guard is narrow rather than dead: where a newline genuinely reaches the linter — a 3.0 text value template, whose text node is not attribute-normalized — the same wrapped run correctly yields no defect. So the docblock's "a run that wraps a line is left alone here as it is everywhere else" is true only there, and reads as a general promise. An issue for the normalization mismatch seems better than folding it in here.

---

- **The bug reproduces exactly as described.** Master: 1 warning, `--fix` writes `ancestor  :: b`, re-run reports nothing — clean on a file still holding a run the check defines as redundant. This branch: 2 warnings, `--fix` writes `ancestor :: b`, re-run clean.
- **The generality argument holds.** Only an axis token can carry a gap at all: `NAME`, `NUMBER`, `OTHER` and `USER_FUNCTION` cannot contain whitespace by construction, and `instance of` is matched with exactly one space. So one run per token is the most that exists, and `INTERNAL.exec` returning only the first match cannot hide a second — no first-match bug is reachable, and the rule needs no axis list.
- Three spaced axes in one expression report and fix all six runs, in order.
- The `{2,}` boundary and the feature itself are both pinned — flagging a single gap, or never reporting an internal run, each fails a test.
- 1047 passing, 100% of statements, branches, functions and lines, `npx eslint .` clean, `npx mocha test/xcop.test.js --forbid-pending` 253 passing including the new pack.

Issue #681, opened 2026-08-06
`redundant-namespace-declarations` reports a spaced declaration two columns right of it


maxonfjvipon:
`redundant-namespace-declarations` works out the column it reports by arithmetic
over the attribute's name, at `src/namespace-linter.js:89`:

```js
line: attribute.lineNumber,
pos: attribute.columnNumber - attribute.name.length - 1,
```

xmldom reports an attribute at its opening delimiter, so subtracting the name
and one more character lands on the name's first letter only when the source
spells the declaration `name="value"` exactly. A gap around the `=` moves the
delimiter right while the name stays put, and the report follows the delimiter:

```xsl
<xsl:stylesheet xmlns:xsl="..." xmlns:dead = "urn:dead" version="2.0">
```

`xmlns:dead` begins at column 66. The defect is reported at column 68 — two
columns right of the declaration, which is where the two added gap characters
went. An editor integration puts the squiggle there, on the `ad` of the prefix
rather than on the declaration.

This is the invention #594 removed one layer down. `deletion` used to rebuild the
text it meant to cut and had the same three assumptions inside it; it now reads
the span from the source, so after that change the *fix* cuts the right
characters while the *defect beside it* still points two columns away. The two
disagree about where one declaration stands.

The deeper reason is that this linter builds its defect object by hand instead of
through `defect` in `src/checks.js`, which exists to walk the raw text and answer
exactly this question — it is what #611 added when a wrapped or entity-shifted
value reported the wrong place. Routing this check through the shared builder, or
taking the position from the span `deletion` already computes, would put one
answer in one place.

Issue #553, opened 2026-07-30
`redundant-namespace-declarations` false-positives on an exclude-result-prefixes-only prefix and its `--fix` corrupts the sheet


maxonfjvipon:
`redundant-namespace-declarations` (`src/namespace-linter.js`) reports a prefix as unused when its only reference is in `exclude-result-prefixes` (or `extension-element-prefixes`), and — worse — its safe `--fix` then deletes the declaration and leaves the reference dangling.

Verified on a 1.0 sheet whose `foo` prefix is declared and named only in `exclude-result-prefixes`:

```xsl
<xsl:stylesheet version="1.0" xmlns:xsl="..." xmlns:foo="urn:foo" exclude-result-prefixes="foo">
```

The check warns "prefix that is never used" because `used()` scans element/attribute QNames but not the `exclude-result-prefixes`/`extension-element-prefixes` token lists. Running `--fix` then removes `xmlns:foo="urn:foo"` while leaving `exclude-result-prefixes="foo"` in place — an unbound prefix a conformant processor rejects (XTSE0808). A "safe" fix that turns a valid stylesheet into an invalid one is the most serious kind of defect here.

Count a prefix named in `exclude-result-prefixes`/`extension-element-prefixes` (and honour `#all`) as used, so it is neither reported nor deleted.

PR #683, opened 2026-08-06
Read a deleted attribute's span from the source

### LINKED ISSUE #594: `deletion` declines a single-quoted or spaced attribute in `src/fixes.js`
The shared attribute-deleting fix in `src/fixes.js:15` does not read the source it is about to cut. It rebuilds the text from the DOM:

```js
const deletion = function(attribute) {
  return {
    line: attribute.lineNumber,
    col: attribute.columnNumber - attribute.name.length - 2,
    value: ` ${attribute.name}="${attribute.value}"`,
    replacement: '',
  }
}
```

One space, an `=`, and a double quote are assumed. `src/fixer.js` then decode-walks the raw source looking for exactly that string, so an attribute written any other way never matches and the fix is dropped. A single-quoted delimiter — the normal spelling when the XPath itself holds double quotes, as in `select='concat("a", b)'` — misses, and so does a stray space around the `=`.

I ran `--fix-suggestions` over a sheet holding `select='count(o)'`, `select="o[@n &lt; 3]"` and `select = "'a'"`, each beside a body. Only the middle one was rewritten, which shows the entity handling of #518 and #525 works as advertised; the other two came back as "Skipped fixing ..., the source no longer matches" and stayed in the report. The run counts them as fixable, announces a fix, then declines it, and the author is left to make the edit by hand for a reason that has nothing to do with their code.

Three fixers route through this helper — `using-disable-output-escaping` at `src/fixers.js:16`, `mode-or-priority-without-match` at `src/fixers.js:65`, and `variable-or-param-with-select-and-content` at `src/fixers.js:183` — so the gap is not one check's. The last of those is the only check in the tree marked `mature: true`, and its motive sells the deletion as the correction it offers.

Let `deletion` take the delimiter and the spacing from the source rather than inventing them: seek from the attribute's recorded line and column, accept whichever quote character follows the `=`, and end the span where that same character returns.

maxonfjvipon:
`deletion` rebuilt the text it meant to cut as ` name="value"`, which assumes three separate things about how the attribute is spelled: a double quote, one space in front of the name, and no gap around the `=`. The fixer applies a fix only where the source decodes to its `value`, so every other spelling was announced as fixable and then refused for a reason having nothing to do with the stylesheet. A single-quoted `select='...'` — the ordinary choice when the XPath itself holds double quotes — was the common casualty; so were `select = "x"` and a wider gap in front of the name. An attribute standing on its own line was not refused, and the first version of this paragraph said it was: master applies that one, cutting a single space of the indentation and leaving the `>` stranded on a line of its own. What changes there is the extent of the cut, not whether it happens.

Now the span is read from the source. xmldom reports an attribute at its opening delimiter, so the delimiter is whichever quote stands there and the value ends where that same quote returns, XML forbidding it inside:

```js
const quote = offsetAt(content, attribute.lineNumber, attribute.columnNumber)
const start = opens(
  content,
  opens(content, opens(content, quote) - 1) - attribute.name.length,
)
```

`opens` walks back over a run of gap characters, so the three nested calls cross the gap before the quote, the `=`, and the gap in front of the name whatever their width — a line ending among them, which is why a wrapped attribute now works too. The value is decoded with the same `character()` walk the fixer uses to verify, so the two agree by construction rather than by luck; a numeric entity decodes to nothing either walk can express, so that one case is still declined, and an `UNCHANGED` row pins it.

Four callers were threaded the raw source: the three fixers in `src/fixers.js` and `src/namespace-linter.js`, whose docblock had documented the very limitation this removes.

One consequence to weigh before landing. On master, a single-quoted `xmlns:foo='urn:foo'` was announced and declined, which accidentally spared the sheet from #553 — the false positive on a prefix used only by `exclude-result-prefixes`, whose `--fix` cuts a declaration the stylesheet needs. With the deletion working, that corruption reaches the single-quoted and spaced spellings too. The cure is #553, not a fix that refuses to do what it advertises, but the sequencing is yours.

Left alone deliberately: #681, the reported column of a spaced namespace declaration, which is the defect's position and not this fix's, and #667 on the import tier.

Closes #594


maxonfjvipon:
The own-line case is yours and the body was wrong. I reproduced it: master applies that fix, cuts a single space of the indentation, and leaves the `>` stranded on a line of its own — valid XML, and ugly. What this change does there is widen the cut to the whole gap run, which is a behaviour change to a case that already worked, not a refusal lifted. The body now says so.

On the sequencing I'd land as-is and take #553 next, and I'd not take the third option at all. The tier describes the edit, and deleting a declaration nothing references is deterministic and semantics-preserving; marking it a suggestion would be a false statement about the fix, used to shield a detection bug, and it would disarm a correct fix on every sheet that does not trip #553. It is also the opposite of how the neighbouring case was settled on #669 two days ago — there the docblock's safety claim was narrowed and the hazard filed as #667, rather than the fix retiered — and a tier change is not one line, since `README.md`'s fix lists and the docs site move with it.

#553 is cheaper than "fix it first" suggests, which is what makes that ordering the right one. Its cure is a token-list read in `used()` honouring `#all`, in this very file:

```js
const used = function(elements, prefix) {
  const qualifier = `${prefix}:`
  return elements.some(...)   // scans QNames and attribute values, never a prefix list
}
```

Two attributes hold prefixes as bare tokens — `exclude-result-prefixes` and `extension-element-prefixes` — and `foo` does not contain `foo:`, so the scan cannot see them. A few lines in the right place, against a stopgap that lies about the tier.

The stringified `undefined` is a fair smell and I checked how far it can travel: `src/reporters.js` never mentions `fix`, and the json report drops the key, so no format can print it. That is containment by accident rather than by design, which is your point.

Your last one deserves its own issue and I think it is the sharper form of the `UNCHANGED` row I added. A defect that advertises "1 more fixable with --fix-suggestions" and can never clear misinforms the user on every run, and `deletion` is in a position to know: a span holding an entity `character()` cannot decode will never match, so it could answer null and stop advertising. Out of scope here, and nothing tracks it — the report on #670 is buried in a merged pull request. I will file it as the class rather than the instance unless you would rather it hang off #629.

Tree state: master merged through #684 and #688, `test/fixer.test.js` renamed to `test/fixer.deep.test.js` on the way, 18 checks green, 1080 passing, 100% on all four counts.

Issue #694, opened 2026-08-07
One rejected fixture blanks the shared xcop verdict, failing 204 unrelated tests


maxonfjvipon:
Since #687 the xcop suite asks the tool once over a directory and every assertion reads its own line out of that one report:

```js
let verdict = ''
if (available) {
  verdict = xcopped(SCRATCH)
}
```

```js
assert.ok(verdict.includes(`${fixture.file} looks good`))
```

xcop stops at the first file it rejects, so the report holds no line for anything it never reached. Every one of those assertions then fails, all with the same message, and none of them names the file that actually broke.

Adding one non-canonical fixture to `namespace-packs` turned the run into 204 failures — and a different count on the next run, 237 passing once and 57 the next, since the failures follow wherever xcop stopped in a directory whose order is not fixed. The one real complaint was in xcop's own output, which the assertion never prints:

```text
Invalid XML formatting in .../all-result-prefixes-excluded-0.xsl
```

Finding it meant moving fixtures out of the tree by hand until the run went green. The suite spent 26 seconds per attempt to say nothing about which file it meant.

Read each fixture's verdict out of the report as now, but when a file has no line at all, fail with what xcop said instead of with a bare falsy assertion — and let a fixture xcop rejected fail its own test rather than everyone else's. Asking xcop per directory, rather than once for all of them, would bound the blast radius too.

Issue #687, opened 2026-08-06
Deep test suite spends 52 seconds starting processes


maxonfjvipon:
The xcop suite at `test/xcop.deep.test.js:70` starts one `xcop` process per fixture. There are 257 of them, each costing about 100ms of ruby startup, so the suite spends 25.7 of its 26 seconds waiting for interpreters to boot. xcop takes a directory and reports one line per file, so a single call over all 257 costs 0.148s here — the same 257 assertions for 1/170th of the wall clock.

```js
const xsl = path.resolve(
  __dirname, `temp-${path.basename(pack, '.yaml')}-${index}.xsl`,
)
fs.writeFileSync(xsl, `${xml.parsedFromString(input)}\n`)
const stdout = runXcop(xsl)
```

Those scratch files also land in `test/`, inside the working tree, which is a race rather than only a mess. `should test default directory` in `test/xslint.deep.test.js:96` runs xslint with no path, so it walks the repository — including `test/`, where the line above is creating and unlinking `temp-*.xsl`. A file that vanishes mid-walk takes the file count with it. Serially the two never overlap; under `mocha --parallel` one run in four fails on `Processed files: [1-9]\d*`, and four runs with the xcop file excluded pass. The project's own rules already ask for temporary directories, and every other deep test uses `mkdtempSync`.

The node side pays a similar toll for a different reason. `src/index.mjs:9` imports `xslint.js` at module scope, which pulls in fontoxpath, xmldom, commander and — through `src/xpath-linter.js` — reads and parses all 67 check YAMLs before `main` runs. That is 137ms before a byte of XSL is read, and `xslint --version` pays every millisecond of it. The 132 node spawns in the deep suite spend roughly 18 of their 25 seconds there.

Writing the fixtures under `mkdtempSync` and calling xcop once over that directory fixes the race and the bulk of the cost together; loading the pipeline behind the command action rather than at import trims what is left, and with xcop no longer dominating, running the `deep` target under `mocha --parallel` is finally worth the flags.

PR #697, opened 2026-08-07
Let a fix reach across a line wrap, and stop calling one redundant

### LINKED ISSUE #628: `redundant-whitespace` flags the indentation of a wrapped expression

### LINKED ISSUE #629: `decodes` cannot match a fix value across a line wrap in `src/fixer.js`

maxonfjvipon:
A parser turns every line ending inside an attribute value into a space, so a check reading `count(item) > 0` off an expression wrapped after `count(item)` builds a fix whose space stands where the source holds a newline. `decodes` compared them character for character, failed on that one, and declined the whole fix. Nothing could be fixed across a wrap, whatever offered it:

```text
[WARNING] Skipped fixing unabbreviated-axis at sheet.xsl:11, the source no longer matches
[WARNING] Skipped fixing count-compared-to-zero at sheet.xsl:9, the source no longer matches
```

That wording is for a span an earlier edit moved, and nothing had touched the file. Now a line ending answers to a space, the last normalisation between the two texts, and `count(item)\n > 0` becomes `exists(item)` while `parent::\n node()` becomes `..`.

The second half is why this closes #628 as well. Landing the first half alone would have armed a bug rather than fixed one: `redundant-whitespace` reads a wrapped expression's indentation as a run to collapse, and its fix was unreachable only because `decodes` refused it. With the refusal gone, plain `--fix` would have joined

```xsl
<xsl:value-of select="@a
                      or child::deep"/>
```

onto one line, under the safe tier, on every wrapped expression anyone had written.

The check already meant to leave those alone — `!/[\r\n]/.test(token.value)` sat right there in `redundancies`. It could never fire against a wrapped attribute, because the value it tests is the parsed one, where the wrap became a space before the check ever saw it. It was not useless, though: a character reference is exempt from §3.3.3, so `&#10;` puts a real line ending in the value that the source never shows. Each text sees a wrap the other cannot, so the guard now reads both:

```js
const wrapping = function(source, found, run) {
  const from = rawly(source, found, run.offset)
  return /[\r\n]/.test(run.value) || /[\r\n]/.test(
    source.content.slice(from, skip(source.content, from, run.value.length)),
  )
}
```

`rawly` is the raw-offset walk `defect` was already doing inline, lifted out of `src/checks.js` and exported, so the two agree by construction rather than by a second copy. A wrapped run is now neither reported nor fixed; the doubled spaces standing within one line still are, which is everything the check exists for.

The committed `unabbreviated-axis-in-a-wrapped-value.fixed.xsl` changes to record both halves: `child::\n alpha` collapses to `alpha`, which it could not before, and the `@a\n or deep` wrap beside it survives untouched, which it would not have. Three mutations, three caught — dropping the licence in `decodes` fails five tests, letting `wrapping` read the parsed run instead of the source fails the new pack, and narrowing the window it reads to one character fails the trailing-space fixture.

Closes #629
Closes #628


maxonfjvipon:
Good to merge. One thing worth fixing first, because it lands in the class this
closes; the rest is notes.

**1. A line ending spelled as a character reference regresses from silent to
announced-and-refused.**

Moving the test from the value to the source loses the one case the value test
got right: where the *value* holds a line ending that the *source* spells as a
reference.

```xsl
<xsl:value-of select="alpha &#10;  or beta"/>
```

`&#10;` is exempt from §3.3.3, so the parsed value holds a real newline, but the
raw source under the run holds no `[\r\n]` — `wrapping` answers no, the run is
reported, and `decodes` then declines it, because `character` reads a numeric
reference as `undefined`. Master is silent here. `&#13;` and `&#x0A;` behave the
same way.

The class is pre-existing — `&#9;` is announced and refused on master today, and
the corpus-wide `--fix` differential below leaves exactly one decline standing,
`variable-with-select-holding-a-numeric-entity.xsl`, of that same class. But this
widens it by precisely the spellings master's `/[\r\n]/.test(token.value)`
covered. It takes the reference used as inter-token whitespace; inside a string
literal, which is where anyone really writes one, the lexer keeps it whole and
nothing changes either way.

One condition restores master's silence and keeps both halves:

```js
return /[\r\n]/.test(run.value) || /[\r\n]/.test(
  source.content.slice(from, skip(source.content, from, run.value.length)),
)
```

Verified: `&#10;`, `&#13;` and `&#x0A;` go quiet, the `&#9;` row stays exactly as
master has it, both the #629 and the #628 fixtures behave unchanged, and the
suite stays green at 1103. The alternative — keep the report and withhold the
fix — is the bigger change and closer to #636's ground.

**2. The width of the slice is load-bearing but unpinned.**

Mutating `skip(source.content, from, run.value.length)` to `from + 1` leaves all
1103 tests passing, and makes a run whose newline is not its first character —
`select="alpha  ⏎      or beta"`, a trailing space before the wrap — reported
*and* collapsed onto one line by plain `--fix`, which is #628 again. The only
shapes that reach it carry trailing whitespace, which yamllint's
`trailing-spaces` forbids in a pack, so pinning it means a committed `.xsl`
rather than a pack row. Worth one either way, since the pack's three wraps all
begin with their newline.

The `char === ' '` side of the licence in `decodes` is unpinned too: dropping it,
so a source newline answers to any character at all, changes nothing measurable.
That looks unreachable rather than untested — a value built from a normalised
attribute can only hold a space there — but 100% branch coverage says nothing
about it either way.

---

- **Both faults reproduce exactly.** Master reports 6 defects on the fixture,
  announces 4 fixable, declines all 4 with "the source no longer matches", leaves
  the file byte-identical, and re-reports the same 6 — announced forever. The
  branch reports 4, fixes 2, and re-reports only the two unrelated ones.
- **The second half's premise holds, and more sharply than the body puts it.**
  With only the `decodes` licence applied to master, plain `--fix` collapsed a
  deliberately wrapped three-line `select="@aa ⏎ or child::deep/@bb ⏎ or another"`
  onto one line, under the safe tier. With both halves the wrap survives and only
  the axis is abbreviated.
- **Corpus-wide the shape is exactly right.** 369 stylesheets: 1074 defects to
  1063, and every one of the 11 removed is a `redundant-whitespace` on a wrapped
  value — nothing added, nothing moved by a column, no other check touched. Two of
  the eleven are `directives/wrapped.xsl` and `wrapped-tag.xsl`, the fixtures
  already excluded from the xcop sweep for keeping a wrapped value, so the repo's
  own fixtures were victims of #628.
- **`--fix` over 98 stylesheets differs in exactly the two files this is about**,
  each by the intended edit, and all 98 still parse. The `@a ⏎ or deep` wrap
  beside the fixed `child::` in `unabbreviated-axis-in-a-wrapped-value.xsl` is
  untouched, which is both halves in one fixture.
- **CRLF and a lone CR work**, since `character` already reads either as the
  newline a parser leaves: a CRLF copy of the whole fixture fixes cleanly and
  keeps its endings. A literal tab does not — §3.3.3 turns it into a space too,
  and the licence admits only `\n` — but that is master's behaviour unchanged.
- **A text value template and a CDATA section are unaffected**, on both trees: a
  text node is not normalised, so the fix value holds the real newline and matched
  before this change as it does after.
- **Nine fixes across four wraps in one attribute** all apply in one run, the file
  parses, and a second run converges.
- **Six of six new or changed tests fail against master's `src`**, including the
  pre-existing CRLF test riding the updated `.fixed.xsl`. Dropping the licence
  fails 5, not the three the body claims; letting `wrapping` read the parsed run
  fails 3; reporting every run fails 7; reporting none fails 3.
- `redundant-whitespace` carries no `mature: true`, so no freeze is broken, and
  `rawly` agrees with `defect` by construction because `defect` now calls it.
- 1103 passing, 100% of statements, branches, functions and lines, `npx eslint .`
  clean, `npx mocha test/xcop.deep.test.js --forbid-pending` 258 passing including
  the new pack — which passes because the harness re-serializes through xmldom
  first, joining the wrap before xcop sees it.

maxonfjvipon:
Point 1 is right and it is my reasoning that was wrong, not just the code. I moved the test to the source having convinced myself the source sees every line ending the value does, and wrote that into the docblock — "the value cannot answer it". A character reference is exempt from §3.3.3, so `&#10;` is precisely the case where the value holds a line ending the source never shows, and each text sees a wrap the other cannot. Fixed as you wrote it, and the docblocks that asserted otherwise now say why both are read.

Measured against master rather than reasoned about, on `select="alpha REF  or beta"`:

```text
&#10;   master=0  branch=0
&#13;   master=0  branch=0
&#x0A;  master=0  branch=0
&#9;    master=2  branch=2
```

So the three go quiet and the tab keeps master's announced-and-refused exactly, which is the pre-existing class and not this change's to widen or narrow.

Point 2 is fixed too. `test/resources/fix/a-wrap-behind-a-trailing-space.xsl` puts the newline behind a trailing space, in a committed `.xsl` for the reason you gave — yamllint's `trailing-spaces` would reject it in a pack — with an `UNCHANGED` row. Narrowing the window to one decoded character now fails exactly that row. It has to keep `skip` in the call to be measurable at all: the more obvious mutation, slicing `from, from + 1`, orphans the import and dies at lint before a test runs, which is worth knowing for the next person who probes this file.

Your count is right and mine was wrong: dropping the licence fails five, not three. I measured it mid-change, before the `.fixed.xsl` update and the last rows were in, and never re-measured. The body now says five, and says three mutations rather than two.

On the `char === ' '` side being unpinned — agreed, and I would rather leave it unpinned than manufacture a fixture for it. A fix value only reaches `decodes` from a normalised attribute or from a text node that was never normalised, so nothing that exists can put a non-space where the source holds a newline; the condition is there to keep the licence narrow if that ever stops being true. Calling it unreachable-rather-than-untested is the accurate reading.

Nothing else changed: 1104 passing, 100% of statements, branches, functions and lines, eslint clean, xcop 258.

maxonfjvipon:
You are right that it was untested, and I checked rather than took it: dropping `/[\r\n]/.test(run.value)` from `wrapping` left 1104 passing and branches 100% at 956/956. A licence with no test is the same thing as no licence a refactor from now.

Landed with your row, `test/resources/xpath-format-packs/a-wrap-written-as-a-reference.yaml`, verbatim. It bites — the same mutation now fails it with `AssertionError [ERR_ASSERTION]: 2 == 1`, the second line's typed-out double space reported alone.

It has to sit in `UNFORMATTED` in `test/xcop.deep.test.js`, which is the one thing your note did not cover. The fixture's `&#10;` is its whole point and xmldom serializes it back as a raw line ending, so what xcop reads is a wrapped attribute it refuses — and since #687 made that one run over one directory, a single refusal costs all 257 other fixtures their verdict (#694). The docblock says so.

Now 1105 passing, statements/branches/functions/lines all 100% (branches 957/957), xcop 258, eslint and yamllint and typos clean.

Issue #703, opened 2026-08-07
A name swallows the `::` behind it in `src/tokens.js`


maxonfjvipon:
`afterName` in `src/tokens.js:328` runs a name over every character the `NAMED` class at `src/tokens.js:290` matches, and that class holds `:`:

```js
const NAMED = /[\p{L}\p{N}\p{M}_.:-]/u

const afterName = function(xpath, start) {
  let at = start
  while (at < xpath.length && NAMED.test(xpath[at])) {
    at += 1
  }
  return at
}
```

A QName carries one colon, which is why it is there. It never carries two adjacent, because `::` is the axis separator and a token of its own — `TOKENS.COLONS` exists for it. So a `::` standing directly behind a name is eaten by the name instead of being lexed, and the same characters give two different streams depending on whether a gap stands in front:

```text
tokenized('a::b')    name<a::b>
tokenized('a ::b')   name<a> whitespace< > ::<::> name<b>
```

The first stream holds no `::` at all. It is not the axis names that are special, but the gap: `foo::a` lexes as `name<foo::a>`, `$v::a` as `$<$> name<v::a>`, `child::a::` as `child<child::> name<a::>`. Where no name precedes, the separator is lexed correctly — `1::a` gives `number<1> ::<::> name<a>` and `)::a` gives `)<)> ::<::> name<a>` — so the defect is confined to the one position a name ends in.

A second symptom rides on the same cause. `opensAxis` at `src/tokens.js:367` declines to open an axis when `spelling` reports a name already in progress, and `:` counts as a name character to that question too, so whether an axis name opens an axis flips on the gap:

```text
tokenized('child:: child::b')   child<child::> whitespace< > child<child::> name<b>
tokenized('child::child::b')    child<child::> name<child::b>
```

`src/xpath.js:265` needed exactly that exception when it asked the same question of a node test and carries it — `spelling(xpath, at) && xpath[at - 1] !== ':'`. `opensAxis` does not.

Nothing reports differently today. Every expression this reaches is malformed, `isValid` refuses it whichever way it lexes, and since #636 a refused expression carries no fix. The cost falls on #677: a recursive-descent grammar handed `name<a::b>` cannot say that `a` is not an axis, nor point at the `::` it did not expect, because the `::` is inside the name token. That is the wrong-stream hazard #685 was filed for rather than the unbuildable-stream one #676 was, and it is the last of the three in the same place.

[...]

Issue #701, opened 2026-08-07
markdown-lint passes on a violation, because the job runs with fix: true


maxonfjvipon:
The `markdown-lint` job runs the action with `fix: true`:

```yaml
      - uses: DavidAnson/markdownlint-cli2-action@v19
        with:
          fix: true
          globs: '**/*.md'
          config: '.markdownlint.json'
```

So it repairs what it can in the runner's checkout, exits `0`, and throws the
repair away — nothing commits it and nothing reports it. Every auto-fixable
violation is therefore invisible in perpetuity, and only an unfixable one can
ever turn the job red.

`CLAUDE.md:526` on `master` is one of them today, arrived with #683:

```text
CLAUDE.md:526:312 MD038/no-space-in-code Spaces inside code span elements
  [Context: "` name="value"`"]
```

Run the same action's tool over that file with `--fix` and it exits `0` having
rewritten the span to `` `name="value"` ``; run it again without `--fix` on the
untouched file and it exits `0` too, because the runner's copy was already
repaired. The violation has been on `master` since #683 merged and every run
since has passed.

This is the shape this repository calls a machine-enforcement hole: a checker
that rewrites instead of failing enforces nothing. It is also the one place a
convention was supposed to be held automatically, so the file that says "every
style or consistency convention must be machine-enforced" is the file the
un-enforcement is sitting in.

Drop `fix: true` from the workflow so the job fails on a violation, and clean up
whatever the first red run names.

maxonfjvipon:
Concrete evidence, found while measuring something else: the violations this
job cannot fail on are already in the tree.
`src/resources/motives/xpath/with-param-use-in-invalid-parent-node.md` breaks
MD031 four times over — fenced blocks with no blank line around them, at lines
34, 36 and 43 — on `c31b23a` with every CI job green. Running
`npx markdownlint-cli2 "**/*.md"` locally reports them; the workflow does not,
because `fix: true` rewrites what it finds and then exits on what is left.

So the job is not merely capable of passing on a violation, it is doing so
today, and the file it is quiet about was last touched by the very commit that
introduced the 100-character rule. Dropping `fix: true` turns the job red until
those four are fixed, which is the point of the ticket rather than an obstacle
to it.

Issue #677, opened 2026-08-06
Write the XPath 3.1 expression grammar as recursive descent over the token stream


maxonfjvipon:
Step 3 of #228 needs a grammar, and this is the expression half of it: recursive
descent over `tokenized`, one function per production, roughly forty of them for
the XPath 3.1 expression grammar. Every node's span is a token range, so a
position is never computed — it is carried. Whitespace and comments stay in the
stream as trivia, which is what makes the tree lossless and what keeps `--fix` a
span replacement over raw source rather than a re-serialisation.

No new dependency. The audit on #228 established that `fontoxpath.parseScript`
returns an XQueryX AST with no per-node positions, and positions are the whole
point, so the tree we need is not the tree our engine already builds.

**It takes the version in force as a parameter.** That is the property #652 turns
on: modern XPath syntax in a `version="1.0"` stylesheet is a parse failure against
the 1.0 grammar, rather than an entry on a curated list of spellings nobody can
keep current. `versionOf(node)` in `src/xsl-version.js` already answers what is in
force at the node the expression hangs off.

A syntax error then carries its own offset, so `invalid-xpath-expression` finally
points at the fault instead of at the attribute holding it.

Blocked by #676: the path punctuation has no token kind of its own, and `PathExpr`
is written against exactly those characters.

maxonfjvipon:
Blocked by #685 as well as #676. Both are the same lexer, but they fail this issue differently: #676 left the path punctuation in one undivided `other` token, so a `PathExpr` production had nothing to read; #685 leaves the 3.1 arrow lexing as `=` then `>`, so a production *does* read something — the wrong thing. `a => f()` arrives token for token identical to `a = (> f())`, and a recursive-descent parser built on that stream would return a comparison with no symptom to show for it. Land both before the first production is written.

Issue #664, opened 2026-08-05
The starts-with-double-slash selector duplicates PATTERNS with nothing keeping them in step


maxonfjvipon:
Two places now name the attributes that hold an XSLT pattern, and nothing keeps them in step.

`PATTERNS` in `src/attributes.js:34` is the list every code-based linter reads, so that a fixer knows it is holding a pattern rather than an expression:

```js
const PATTERNS = [
  'match', 'count', 'from', 'group-starting-with', 'group-ending-with',
]
```

`src/resources/checks/xpath/starts-with-double-slash.yaml` names the same five, plus the element each one may sit on, written out by hand:

```text
(//xsl:template | //xsl:key | //xsl:accumulator-rule)/@match
| //xsl:number/(@count | @from)
| //xsl:for-each-group/(@group-starting-with | @group-ending-with)
```

The two agree today. Nothing makes them agree tomorrow. A future XSLT version adding a pattern attribute, or a correction to the elements one may appear on, lands in whichever of the two the author happened to be reading — and the failure is silent in both directions: a name added only to `PATTERNS` leaves this check blind to it, and a name added only to the selector leaves the code-based tier treating it as an expression, where a rewrite legal in one language is a syntax error in the other.

That is the drift shape this repository has closed twice already. #647 was three attributes in `src/attributes.js` that the validator's rival list never learned, fixed by deriving one from the other rather than by adding three names. #607 was `found.fixes` declared in a pack that no harness read, fixed by two `conformance.test.js` gates that derive which packs and directories are in scope instead of trusting a list. Both replaced agreement-by-diligence with agreement-by-construction.

A `conformance.test.js` case asserting that the `starts-with-double-slash` selector mentions every entry of `PATTERNS` would fit that pattern and cost one test. Deriving the selector itself would be better still, though it needs somewhere for the element-per-attribute mapping to live, since `PATTERNS` carries names only and `xsl:number/@count` is a pattern where another element's `@count` would not be.

PR #704, opened 2026-08-07
Sweep the respelling retry instead of disclaiming it

### LINKED ISSUE #641: `squeezed` fabricates a comment delimiter in `src/xpath.js`
`squeezed` in `src/xpath.js:163` deletes the whitespace around an axis's `::` and between a node
test's name and its `(`, then hands the result to `compiles`, so `isValid` accepts an expression
whose respelled form parses. The premise is that squeezing can only ever widen what is accepted —
the docblock says it "cannot make a broken expression whole." Deleting a gap can put two characters
side by side that form a token the source never held, and `(:` and `:)` are the XPath 2.0 comment
delimiters, one token each. A deletion next to a `(` or a `)` fabricates one, and the comment then
swallows whatever wreckage surrounds it.

```js
const SPACED_AXIS = /\s*::\s*/g
const SPACED_TEST = new RegExp(`(^|[^\\w.-])(${TESTS.join('|')})\\s*\\(\\s*`, 'g')

const squeezed = function(xpath) {
  return xpath
    .replace(NAMESPACE_AXIS, 'child::')
    .replace(SPACED_AXIS, '::')
    .replace(SPACED_TEST, '$1$2(')
}
```

Every expression below is a parse error to the engine — it reports `1: <source>` rather than a W3C
code, so `compiles` is not forgiving it on the first pass — yet `isValid` answers true:

```js
isValid('a (: b :: )')             // squeezed to `a (: b::)`, the fabricated `:)` closes a comment
isValid('a ( :: b :)')             // squeezed to `a (::b :)`, the fabricated `(:` opens one
isValid('count(node ( :) 1 :) )')  // SPACED_TEST merges `( :` into `(:`
isValid('a[text ( :) + :) ]')      // the same, inside a predicate
```

There are two independent sites. `SPACED_AXIS` deletes the gap on both sides of the colons, so it
merges `::` with an adjacent `)` or `(`. `SPACED_TEST` ends in `\s*`, which deletes the gap behind
the bracket it just tightened. Neither is reachable on `master`, whose retry only ever ran when the
literal name `namespace` appeared; the retry became unconditional in #640, which is where this
enters. So `master` answers false to all four today, and will not once that branch lands.

Narrowing the whitespace class does not help, which is worth saying plainly because it is the
obvious first move: the gaps above are ordinary spaces, so spelling the class as XPath's
`ExprWhitespace` — `S ::= #x20|#x9|#xD|#xA`, the way `src/tokens.js:68` already spells it — leaves
every case standing. Dropping `SPACED_TEST`'s trailing `\s*` closes the second site and leaves the
first untouched.

The repair is to stop respelling with regular expressions, since a substitution cannot see the
tokens it is rearranging. Lexing the expression with the positioned lexer in `src/tokens.js` and
rebuilding it without the whitespace between tokens drops a gap only where a gap is what it is,
and can never join two characters into a token; failing that, every deletion has to check the
characters it is about to make adjacent. Until one of those is in place, a respelled retry cannot
be trusted to only widen what `isValid` accepts.

maxonfjvipon:
#641 does not reproduce. All four fabrications it reported are refused on `master`, so are the three cases it was reopened for, and every one of them is already pinned by a row in `test/xpath.test.js` — the ticket has been fixed twice over by #640 and #674 and its examples cannot regress silently.

What stood was its last paragraph, which is not about examples at all: a substitution cannot see the tokens it rearranges, so no character-level guard can *prove* the retry only ever widens what is accepted. That claim was written into the code as a disclaimer and left there. It is now swept instead:

```js
  it('cannot change the kinds an expression is made of', function() {
    assert.deepEqual(
      SWEPT.filter((one) => kinds(one) !== kinds(squeezed(one))),
      [],
      'a respelling retyped a token, so the engine reads something else',
    )
  })
```

`SWEPT` is every sequence of up to four of the pieces a fabrication is spelled from — the brackets, the colon, the separator, a gap, the names a squeeze opens on, a quote — which is 16,104 expressions in 54 ms. If no kind moves, the engine is handed the tokens it was always going to read and merely spelled the way it insists on. Against `squeezed` as it stood when #641 was filed, 120 of the sweep fail, the ticket's own `( ::` -> `(::` first, conjuring a comment out of nothing.

The first draft watched only the comment and string kinds, on the reasoning that a fabricated delimiter is the whole of the harm. Review showed it is not: a squeeze that merged two adjacent names would leave every delimiter where it stood and still change what compiles — `a b` as `ab`, `child a` as `childa`, both accepted — and that class caught 0 of 16,104 under the narrow property against 368 under this one. Fabricating `(:` or `:)` is the sharp end, since those are one token each and a comment swallows what surrounds it, but it is not the boundary.

`namespace` is an atom for a reason review had to point out: without it not one swept expression held a `NAMESPACE` token, so the line in `kinds` reading `namespace::` as the `child::` it becomes was unfalsifiable — deleting it left the suite green — and the `namespace::` rewrite, the oldest of the three squeezes and the one the engine leaves no choice about, was swept no times at all. With the atom in, 313 of the sweep hold the token, the guards still hold there, and dropping the normalisation fails with `namespace::` at the head of the list.

The sweep's *reach* is pinned as well as its verdict, which review had to ask for twice over. One row per squeeze — `namespace::a`, `child ::a`, `node (a`, `(a )` — each asserted to be a sequence the sweep generates and one the retry rewrites, so a squeeze cannot be swept zero times in silence; and a second guard that the sweep can hold a comment and a string at all, since those are the kinds it must see rather than rewrite. Nine of the eleven atoms now turn something red when they leave, against none before.

Four mutations, four caught: the pre-#640 squeeze fails the sweep, a name-merging squeeze fails it, deleting the namespace normalisation fails it, and a squeeze reduced to the identity fails the second test, which refuses to answer from a sweep no respelling touches.

The sweep stops at four atoms for a reason worth writing down rather than leaving to be rediscovered. A fifth breaks on eight out of a hundred and sixty thousand, every one a pair of axes like `child:: child::` or `namespace ::child::`, each respelling to a separator the lexer then reads as a name. That is #703 rather than anything the retry does.

One docblock claim was false and is corrected. `SPACED_CLOSE` said the deletion "leaves the token stream exactly as it stood" for a gap a string literal carries. It does not — `f('a )')` is retried as `f('a)')`. Nothing observable follows, since the grammar reads into neither a literal nor a comment and the respelling is discarded once it has answered, but the reason given was the wrong one.

The sweep turned up a lexer asymmetry on the way, filed as #703 rather than folded in: a `::` directly behind a name is swallowed into it, so `a::b` lexes as one `name` token and `a ::b` as three. Nothing reports differently today, because every expression it reaches is malformed either way, but #677 will read that stream.

678 fast and 444 deep tests pass, coverage holds at 100% of statements, branches, functions and lines, and xcop is green over all 258 fixtures. `CLAUDE.md:526` still carries the MD038 that #701 is about; it is on `master` and untouched here.

Closes #641


maxonfjvipon:
Good to merge. One sentence in the new docblock is false, which is the same
defect the `SPACED_CLOSE` correction in this PR exists to fix.

**`DELIMITED`'s reason for narrowing is not the true one.**

> Every other kind a squeeze could retype leaves the engine reading the same
> characters

A squeeze that merged two adjacent names would not. It would change what the
engine reads, and every one of these becomes a false accept:

```text
isValid('a b')         -> true     ('ab' compiles)
isValid('child a')     -> true     ('childa' compiles)
isValid('a and or b')  -> true     ('aandorb' compiles)
```

The sweep passes that regression. It does turn the suite red, but through three
unrelated accepting rows about sequence-type brackets — the general case this
test exists to provide is precisely what misses it, so the next merge to slip in
at a position no example covers goes uncaught.

The real reason for stopping at comments and strings is the one `src/xpath.js`
gives, and it is a good one, so it is worth having in both places. Measured over
the same `SWEPT`: a **non-whitespace-token-kind** invariant — the sequence of
kinds with the whitespace tokens dropped — is also **0 violations at depth four**,
and at depth five it breaks on exactly two expressions, `child:: child::` and
`child ::child::`, both respelling to `child::child::`. That is the #703 lexer
asymmetry the docblock already names, not anything the squeeze does wrong. So the
narrow property is what survives depth five and the stronger one is red on #703
alone. Comments and strings are the only retyping the sweep *can* police, which
is a different claim from their being the only one that could do harm.

---

- **The "does not reproduce" claim holds.** All four fabrications are refused, and
  the three cases the ticket was reopened for behave correctly: `1-namespace::x`
  and `1 - namespace::x` accepted, `$a-namespace::x` refused — rightly, since
  `$a-namespace` is one variable name and the `::x` dangles. Six of the seven are
  pinned by a row in `test/xpath.test.js`; `1 - namespace::x` is not there
  literally, though it is the easy one of the three and the hard
  `1-namespace::x` sits at line 60.
- **The sweep has teeth.** Reconstructing `squeezed` as #641 found it and running
  this branch's property against it gives 24 violations — 23 within the fidelity
  of my reconstruction of that era's name list — and the first is the ticket's own
  `( ::` -> `(::`, conjuring a comment out of nothing. The shipped squeeze: 0
  violations, 173 of 11,110 respelled, 57 ms. At depth five, 111,110 sequences and
  2,772 respelled, still 0.
- **The property is not vacuous**, which was the thing worth checking hardest: the
  lexer reads an unterminated `'` or `"` as a string and an unterminated `(:` as a
  comment, and it nests comments, so the fragments the sweep generates really are
  seen as delimited spans rather than falling through to `other`.
- **`node` standing in for the ten node-test names is sound.** All 40 fabrications
  built from each name in turn — `text`, `comment`, `processing-instruction`,
  `schema-element` and the rest — are refused. Double-quoted strings, which the
  atom set has no character for, behave symmetrically to the single-quoted ones.
- **The no-op guard bites.** Making `squeezed` the identity fails `cannot answer
  from a sweep no respelling reaches`, along with 39 accepting rows.
- **The corrected `SPACED_CLOSE` claim is accurate.** `f('a )')` is retried as
  `f('a)')` and compiles the same, so the deletion does reach inside a literal and
  the old "leaves the token stream exactly as it stood" was indeed the wrong
  reason.
- **#703 is exactly as described.** `a::b` lexes as one `name<a::b>` and `a ::b` as
  three tokens; filing it rather than folding it in is right, since every
  expression it reaches is malformed either way today.
- 678 fast and 444 deep passing, 100% of statements (6378), branches (952),
  functions (196) and lines, `npx eslint .` clean, xcop 258, 18/18 CI green on a
  current base, merges clean into master.

maxonfjvipon:
You are right, and the sentence was worse than loose — it asserted the thing the PR was written to stop asserting. I narrowed the property to comments and strings for a good reason and then wrote down a different, false one, which is the same substitution the `SPACED_CLOSE` correction here exists to undo.

Measured your regression rather than conceded it. A squeeze that also merged two names across a gap, over the same `SWEPT`:

```text
delimiter property catches:   0 of 11110
kind property catches:      189 of 11110
```

and `a b` -> `ab`, `child a` -> `childa`, `a and or b` -> `aandorb` all compile, so the class is real false accepts and the sweep was blind to it.

So the test now asserts the kind sequence, not the two delimiters. It is strictly the better net on the original bug too — against `squeezed` as #641 found it the offenders go from 23 to **105**, still led by the ticket's own `( ::` -> `(::`. Three mutations, three caught, and the middle one is yours: the pre-#640 squeeze fails it, a name-merging squeeze fails it, and the identity fails the no-op guard beside it. Under the old property your regression turned the suite red only through three unrelated accepting rows, exactly as you said; it now fails the general test directly.

Your depth-five measurement reproduces: two breaks out of a hundred thousand, `child:: child::` and `child ::child::`, both #703. The sweep stops at four for that reason and the `SWEPT` docblock now says so, rather than leaving a reader to wonder whether five was tried.

`DELIMITED` is gone rather than kept beside the new one — on this sweep a fabricated delimiter always moves a kind, so the two could never fail apart, and an assertion that cannot fail alone is the shape #607 and #660 are about. What it carried that was worth keeping is prose now, in both places: fabricating `(:` or `:)` is the sharp end, since those are one token each and a comment swallows what surrounds it, but it is not the whole of the harm.

Left `1 - namespace::x` unpinned on purpose. Its hard sibling `1-namespace::x` sits in `OPERATED` already, and the spaced form is the case the engine never had trouble with, so a row for it would lengthen the table without asking anything. Happy to add it if you would rather have all three written down.

678 fast and 444 deep passing, 100% of statements (6382), branches (952), functions (196) and lines, eslint clean, xcop 258.

maxonfjvipon:
Confirmed, every number of it, and it is the sharper of your two findings. Counting the token rather than reasoning about the atom list:

```text
                        as shipped   with namespace
swept                        11110            16104
holding a NAMESPACE token        0              313
violations                       0                0
violations, normalisation gone   0              297
```

Zero. So the normalisation was not merely unfired, it was unfalsifiable — and the `namespace::` rewrite, the one squeeze the engine leaves no choice about, had no expression to be tested on. Twice now the docblock has been ahead of what the sweep touches, which is worth naming as a pattern rather than fixing a third time by accident.

`namespace` is in `ATOMS`, and dropping the normalisation now fails the sweep with `namespace::` at the head of the list. Three figures moved with it and all three are corrected rather than left to rot: the sweep is 16,104 expressions at 54 ms, the pre-#640 squeeze fails 120 of them rather than 105, and the name-merging regression 368 rather than 189.

Your depth-five sentence needed the same correction, which is the part I would have missed by taking the atom on its own. It is eight breaks now rather than two, not because anything got worse but because `namespace` pairs with `child` both ways — `child:: namespace::`, `namespace ::child::`, and so on. Every one is still the #703 shape, so the reason for stopping at four is unchanged and the docblock now says eight.

Merge verified locally rather than read off the badge: `git merge-tree` against `60fd359` reports no conflict, so your reading of `mergeable: UNKNOWN` was right.

Leaving `1 - namespace::x` out, as agreed.

1122 passing, 100% of statements (6382), branches (952), functions (196) and lines, eslint clean, xcop 258.

maxonfjvipon:
Taken, and widened, because measuring your guard showed it cannot reach two of the four squeezes. Dropping one atom at a time and asking whether anything turns red:

```text
caught     (  child  '  namespace
not caught )  :  ::  space  a  node  *
```

`node` is the only atom that exercises the node test squeeze and `)` the only one that exercises the closing bracket, and neither is reachable from a kind: `child`, `node` and `namespace` all lex as a bare `name`, so no token type tells one from another until a `::` or a `(` follows. A kind list has a ceiling there rather than a gap — which is the same lesson as the `kinds()` attempt you described, one level up.

So the reach is pinned by what each squeeze rewrites, one row per squeeze, each asserted to be a sequence the sweep generates *and* one the retry changes:

```js
const REWRITTEN = [
  ['namespace::a', 'the axis the engine has no parse for'],
  ['child ::a', 'an axis spaced from its separator'],
  ['node (a', 'a node test spaced from its bracket'],
  ['(a )', 'a gap in front of the bracket it closes'],
]
```

with your kind guard kept beside it for `comment` and `string`, which are the kinds the sweep must *see* rather than rewrite, reading raw token types for exactly the reason you gave. Nine of the eleven atoms now turn something red when they leave, against four:

```text
without '('  both guards      without 'child'      rewrite guard
without ')'  rewrite guard    without 'node'       rewrite guard
without '::' rewrite guard    without 'namespace'  rewrite guard
without ' '  rewrite guard    without '\''         kind guard
without 'a'  rewrite guard
```

`:` and `*` are the two left, and deliberately: a comment still forms from `(` and `::` without the colon atom, and every row spells its node test `a` rather than the wildcard, so neither loss leaves a squeeze unswept. That is what the guard claims and all it claims — I would rather it fell short of the atom list than pad the table with rows that assert a squeeze twice.

Rows are spelled out rather than derived from the patterns, since what a squeeze matches is `src/xpath.js`'s to say and a copy of it in the test is the drift #664 is about.

1124 passing, 100% of statements (6382), branches (952), functions (196) and lines, eslint clean, xcop 258.

Issue #680, opened 2026-08-06
Gate the parser on a round-trip proof and an acceptance diff against fontoxpath


maxonfjvipon:
The tree must be proved lossless and proved no stricter than the engine before any
check reads it, or the migration in phase 4 of #644 ports checks onto a foundation
nobody has measured. Two gates, both merged before the first consumer:

**Round trip.** Concatenating a parse's tokens and trivia reproduces the input byte
for byte, and every node's span slices to exactly its own text. That is the whole
definition of full fidelity, and it is cheap to assert over every expression the
repository's fixtures already carry. A tree that cannot rebuild its input cannot be
trusted to position a fix in it.

**Acceptance diff.** A corpus test parses every expression in every fixture in the
repo and diffs the verdict against fontoxpath's. Each disagreement is then one of
two things: a known engine strictness bug of the #639 family — where the engine
refuses a spelling the grammar allows, which is what the respelling retry in
`src/xpath.js` exists to paper over — or a gap in our grammar. The parser ships when
that list is empty or annotated line by line, and the annotated ones become the
evidence for retiring the retry in phase 3.

Depends on #677.

Issue #652, opened 2026-08-04
Modern XPath syntax in a version="1.0" stylesheet is never reported


maxonfjvipon:
#533 set out to flag 2.0+ constructs in a stylesheet declared `version="1.0"`,
and listed three kinds: instructions, functions, and *syntax* — naming `||`,
`if/then/else` and the `*:name` wildcard. What landed,
`modern-construct-in-xslt-1`, selects elements and one attribute:

```text
//(xsl:for-each-group | xsl:sequence | xsl:analyze-string | xsl:next-match |
   xsl:perform-sort | xsl:namespace | xsl:character-map | xsl:result-document |
   xsl:import-schema | xsl:*[@as])
```

A declarative selector cannot see inside an expression, so the syntax half was
never covered, and nothing else covers it either. `invalid-xpath-expression`
compiles under fontoxpath, which is an XPath 3.1 engine — deliberately, so that
1.0 numeric coercion is not a false positive (#396) — so it accepts every
modern spelling regardless of the version in force. In a `version="1.0"`
stylesheet:

```text
select="a => count()"          3.1 arrow           NOTHING REPORTED
select="map{1:2}"              3.1 map             NOTHING REPORTED
select="let $x := 1 return $x" 3.0 let             NOTHING REPORTED
select="if (a) then b else c"  2.0 conditional     NOTHING REPORTED
select="a || b"                3.1 concatenation   NOTHING REPORTED
select="for $x in a return $x" 2.0 for             NOTHING REPORTED
```

Every one of those is rejected by a conformant 1.0 processor, which is the
portability bug #533 exists to catch, and each is the same severity of mistake
as the `xsl:sequence` the check does flag. A user moving a stylesheet to a 1.0
engine gets a clean xslint run and a failure at transform time.

Two ways to close it, and they differ in more than effort. A curated token list
— the operators and keywords 2.0 and 3.0 introduced — read off `src/tokens.js`
would catch these six today and would sit beside the element list as a second
partial answer. The complete answer is a version-aware grammar: once the parser
of #228 exists, "does this parse under the version in force at this node" is one
question with one answer, `modern-construct-in-xslt-1` becomes a fallback for
the element half only, and no list needs curating as XPath 4.0 lands. The
version is already available at the node through `versionOf`.

Worth deciding which before adding a third partial mechanism.

PR #707, opened 2026-08-08
Parse an XPath expression instead of matching its text

### LINKED ISSUE #677: Write the XPath 3.1 expression grammar as recursive descent over the token stream

maxonfjvipon:
Nothing parsed an XPath expression. `tokenized` produced a stream and every
check read that stream with a regex, which is the substrate #228 exists to
replace. This is the expression half of step 3: recursive descent over the
tokens, one function per production, roughly forty of them.

A node's span is a range of **token indexes**, never a pair of character
offsets. A position is therefore carried from the lexer, which had it, rather
than computed from text by a scan that can be wrong — and the text of a node is
the tokens of its span joined back together:

```js
const shaped = function(kind, from, cursor, children) {
  return Object.freeze({
    kind: kind, from: from, to: cursor.at, children: Object.freeze(children),
  })
}
```

The whole stream comes back with the tree, trivia and all, so joining it
reproduces the expression as written. That is what keeps `--fix` a span
replacement over raw source rather than a re-serialisation that loses the
author's spacing, and it is the property #680 will prove byte for byte.

**The version in force is a parameter**, which is what #652 turns on. The same
text is a different language under a different version — `a to b` is a range in
2.0 and two steps around a name in 1.0 — so modern syntax in a 1.0 stylesheet is
a parse failure against the 1.0 grammar rather than an entry on a curated list of
spellings nobody can keep current. Twenty-one constructs are gated, each row
proving both halves: refused below its floor, admitted at it. It was twenty-two
rows against twenty-one gates until review counted them: `array` and `reference`
each had two rows and `lookup` had none, so that gate was live and unpinned.
Lowering it to `1.0` left the suite green. Two rows — `?a` for the unary lookup
and `$m?a` for the postfix one — now fail when it is neutered.

A refusal is an answer rather than an exception, since a corpus asks about
thousands of expressions and most callers want a verdict. It names what the
grammar expected and the offset it stood at, so `invalid-xpath-expression` can
finally point at the fault instead of at the attribute holding it. Internally a
production throws, so no production has to carry a failure through the ones above
it; `parsed` turns that back into an answer at the one place a caller can see,
and rethrows anything that is not a refusal rather than reporting a bug as a
syntax error.

Review found one expression where it was an exception after all, and it is the
commonest one there is: an expression of nothing but trivia. `offsetOf` guarded
on `cursor.at` and then indexed with `significant(cursor)`, so where the cursor
sat on a gap the guard passed and the index was one past the end. `select=" "`
is real XSL, `isValid` refuses it today and it gets reported; once the parser is
the front door that file would have taken the whole run down, which is the shape
#700 fixed in `importsOf`. The index is hoisted and tested now, and two rows pin
it — ` ` and `(: c :)` both refuse at the end offset. The empty expression
survived only because `tokens.length` was 0.

Two things the lexer already does that the first draft got wrong, both found by
running it rather than reading it: an axis token carries its own `::` (and the
gap in front of it), so a step must not ask for one; and a prefixed wildcard
arrives as a name ending in `:` beside a `*`, not as three tokens.

Verified against the engine rather than against my own expectations, and over a
corpus wider than the first draft of this paragraph used. The repository's
`.xsl` files hold 177 distinct expressions and all 177 agree with fontoxpath —
but they are not a representative corpus, and review was right to say so.
Widening to the inline `input:` of the pack YAMLs brings it to 556, and there it
found 14 disagreements, 11 of them axis lookalikes (`grandchild::x`,
`a-child::b`) that the lexer folded into one `NAME` because a colon was a name
character, so the grammar never saw a `::` to refuse.

#703 has since landed and closed exactly that class — those lex as
`name :: name` now and the grammar refuses all of them. Against master as it
stands the sweep is **556 expressions, 556 round-tripping byte for byte, and one
disagreement**: `my:25l(3)`, a local name beginning with a digit, which the
engine refuses on NCName grounds and this grammar accepts because it asks the
lexer for a name rather than validating one.

That one, and three more a synthetic battery found — `item()` and `map(*)`
standing where a node test does, an unterminated `'unclosed` lexing as a
`string`, and `Q{http://x}a` having no production at all — are recorded in #708
as the acceptance diff's first entries. #680 should read the pack YAMLs as well
as the stylesheets: a corpus of 177 agreed completely and 556 did not, which is
the whole reason any of this surfaced. A 60-case battery of 3.1 constructs
agrees 60 out of 60.

`parsed` is therefore not yet a drop-in for `isValid`, which is exactly why
nothing reads it.

Nothing reads it yet, deliberately: `isValid` and `expressionsOf` move onto it in
Phase 3, behind #680.

Verified: 876 fast + 457 deep, coverage 1333 passing at 100% of statements
(7428), branches (1181), functions (240) and lines; ESLint and typos clean.

Closes #677.


maxonfjvipon:
Landed a lexer fix in #710 that your grammar is downstream of, so flagging it here rather than leaving it to a merge.

`src/tokens.js` was eating a `::` that stood directly behind a name, so `a::b` arrived as one `name<a::b>` token where `a ::b` arrived as three. `parsed` reads a step's axis by token kind, which means an axis hidden inside a name is just a name, and six malformed expressions parse as an ordinary step:

```text
                 engine   grammar on master   grammar with #710
foo::a           invalid  ACCEPTS             refuses
a::b             invalid  ACCEPTS             refuses
a:b::c           invalid  ACCEPTS             refuses
child::a::b      invalid  ACCEPTS             refuses
child:: a::b     invalid  ACCEPTS             refuses
$v::a            invalid  ACCEPTS             refuses
foo ::a          invalid  refuses             refuses
child::a         valid    ACCEPTS             ACCEPTS
```

Measured by dropping the fixed `src/tokens.js` onto your branch and running `parsed` — no change to `src/grammar.js` at all. The spaced `foo ::a` was refused all along, which is the tell that the grammar was right and the stream was lying to it.

Nothing of yours needs to change, and no test of yours goes red. If #710 lands first you inherit the six refusals; if yours lands first they arrive when #710 does.

Issue #711, opened 2026-08-08
A parenthesized step is refused after the first in `src/grammar.js`


maxonfjvipon:
The path production `walked` in `src/grammar.js:550` reads the first part of a path with `postfixed`, which admits a parenthesized expression, but reads every part after a `/` or `//` with `stepped`, which reads an axis step and nothing else. A parenthesized expression is therefore legal where it opens a path and refused everywhere after:

```js
  } else {
    parts.push(postfixed(cursor))
  }
  let rooted = opened
  while (sees(cursor, TOKENS.SLASH) || sees(cursor, TOKENS.DOUBLE_SLASH)) {
    rooted = true
    take(cursor)
    parts.push(stepped(cursor))
  }
```

Run it and the split shows plainly. `(a|b)` and `(a|b)/c` parse, while `//(a|b)` fails with `expected the end of the expression at 2, found "("` and `x/(a|b)` with `expected a name at 2, found "("`. fontoxpath accepts all four.

That asymmetry is XPath 1.0's rule, and only 1.0's. There a `FilterExpr` may open a path and stand nowhere else — `PathExpr ::= LocationPath | FilterExpr | FilterExpr '/' RelativeLocationPath | FilterExpr '//' RelativeLocationPath` — while 2.0 generalised the step itself to `StepExpr ::= PostfixExpr | AxisStep`, which puts a `ParenthesizedExpr` at every position a step may take. `parsed` already receives the version in force as a parameter, so the 1.0 shape is not wrong, only wrongly applied to every version above it.

The reach is not limited to what a user writes. Nine of this project's own check selectors are written in the idiom the parser refuses, `//(xsl:variable | xsl:param)[not(@name)]` and `(/xsl:stylesheet | /xsl:transform)//(xsl:stylesheet | xsl:transform)` among them, and `//xsl:number/(@count | @from)` buried inside the union `starts-with-double-slash` selects on. At Phase 3 of #644 the parser answers `isValid`, and each of those turns into an `invalid-xpath-expression` reported against valid XPath — nine checks disabled by their own selector. The acceptance diff in #680 is what surfaced it, and it is annotated in `test/grammar-corpus.test.js` until this closes, at which point the annotation has to come off or that gate turns red.

Reading the part after a `/` or `//` with `postfixed` rather than `stepped` from 2.0 onward, and leaving `stepped` in place below it, is the whole of the fix; `test/grammar.test.js` already has a `GATED` table for a construct that is legal only above a floor, so the pair of rows belongs there.

Issue #708, opened 2026-08-08
The expression grammar and the engine disagree about 14 expressions


maxonfjvipon:
The grammar in #707 agrees with fontoxpath on all 177 distinct expressions in
the repository's `.xsl` files. Widen the sweep to the inline `input:` of the
pack YAMLs — 556 distinct expressions — and 14 disagree, all one way: the parser
accepts what the engine refuses. They fall into four classes, two of them the
lexer's and two the grammar's, and they are the raw material #680's acceptance
diff exists to work through.

**An axis lookalike is one name.** `src/tokens.js` lexes a name whole and
greedily, and the colon is a name character, so an unknown axis arrives as a
single token with the `::` inside it:

```text
tokenized('grandchild::x')  ->  name:"grandchild::x"
tokenized('a-child::b')     ->  name:"a-child::b"
tokenized('my:namespace::x')->  name:"my:namespace::x"
```

The grammar never sees a `::` to refuse, so it reads a step and accepts. This is
#666's behaviour and predates #707 — the lexer is right that a name is greedy,
and wrong that a name may hold `::`. Eleven of the fourteen are this.

**An unterminated literal is a string.** `'unclosed` lexes as one `string`
token, so the grammar accepts an expression that has no closing quote. Also the
lexer's, and also older than #707.

**An item type stands where a node test does.** `item()` and `map(*)` are item
types, legal after `as` or `instance of` and nowhere else. The grammar's `KINDS`
list holds them beside the real kind tests, so `item()` parses as a step. That
one is #707's own, and the fix is #679's business: a pattern and a step admit a
restricted production set, and naming it is what that ticket is for.

**A braced URI literal is unsupported.** `Q{http://example.com}name` is XPath
3.0's way of writing a name with its namespace inline. The lexer splits it into
six tokens and the grammar has no production for it, so a valid expression is
refused — the only disagreement in the other direction. Also #707's own.

None of it can hurt anybody today: nothing reads `src/grammar.js`, and
`isValid` still answers from the engine. It matters at Phase 3, where the parser
becomes the front door and every one of these becomes a wrong verdict on a real
stylesheet. The two lexer classes want fixing in `src/tokens.js` before then —
an unknown axis and an unterminated string are both things the lexer should
refuse to fold into a name — and the two grammar classes are the acceptance
diff's first entries.

Worth recording the shape of the sweep too, since it is the reason this was
found at all: a corpus of 177 agreed completely, and the same parser against 556
[...]

PR #712, opened 2026-08-08
Prove the parser lossless over the whole fixture corpus, and diff its verdict against the engine

### LINKED ISSUE #680: Gate the parser on a round-trip proof and an acceptance diff against fontoxpath

maxonfjvipon:
#707 proved the grammar against examples — 177 expressions, a 60-case 3.1 battery, one round-trip test over one string. Examples are chosen by whoever writes them, and what nobody thought to write down is exactly what a parser gets wrong. These two gates ask the whole repository instead: 607 distinct expressions, from the committed stylesheets, from the ones the packs hold inline, and from the selectors the declarative checks are themselves written in.

The round trip holds everywhere. Tokens join back into the expression byte for byte, every node's span slices to its own text, and — the part slicing cannot see on its own — every child nests inside its parent, in order:

```js
let behind = node.from
for (const child of node.children || []) {
  if (child.from < behind || child.to > node.to) {
    loose.push(`${child.kind} of ${node.kind} in ${one}`)
  }
  behind = child.to
}
```

Shift a node and its children together by a token and the text still slices, because both sides of that comparison read the same tokens. Nesting is what pins a span to the construct it belongs to, which is the property a fix built from a child's span rests on. All three hold over 2278 nodes.

The acceptance diff asks the engine as `compiles`, not as `isValid`, and that choice is the whole of what this closes. `isValid` is `compiles(xpath) || compiles(squeezed(xpath))`, so the respelling retry stands on the engine's side of the comparison and an expression fontoxpath refuses but the squeeze rescues cannot surface as a disagreement at all. Forty do. They are the #639 family exactly — `child ::gamma`, `parent :: node ( )`, every `namespace::` axis — and the grammar accepts all forty:

```js
const RESCUED = CORPUS
  .filter((one) => !compiles(one) && compiles(squeezed(one)))
```

That set is read from the engine and `squeezed` alone, never from the grammar, so holding the grammar to it is a comparison rather than a restatement. It is also the evidence #680 asked for: the parser already accepts every spelling the retry exists to rescue, so retiring it at Phase 3 costs nothing. It is gated as a set rather than annotated forty times, and a corpus that stopped rescuing anything fails a guard of its own.

Fifty-one disagreements fall out, and they decompose exactly: forty the retry accounts for, and eleven annotated one to a line in `GAPS`, each naming the side that accepts and the ticket that closes it. Ten of the eleven are a check selector, though a selector is one expression in twelve here — the checks are written in an idiom the stylesheets never use, which is why the corpus in #708 agreed almost completely and proved less than it looked like it proved. Nine are a parenthesized expression standing as a step, now #711: `(a|b)/c` parses and `//(a|b)` does not, which is XPath 1.0's rule applied at every version. The other two, a node comparison and a name no NCName can spell, are recorded on #708 along with the re-measurement that its own largest class — eleven of its fourteen — was closed by #710. The eleventh is the single place we are the lenient side, and the engine is right there: `25l` opens with a digit and no NCName may.

A gate over a sweep can pass by finding nothing, so the sweep is gated too: each of the three sources must still yield what it did when this was written. Every gate was checked by breaking what it watches — dropping trivia from the stream, starting a node a token early, ending one a token late, dropping a row from `GAPS`, annotating one on the wrong side, cutting each source out of the sweep in turn, making the retry a no-op, taking the grammar's `namespace::` axis away, and taking the diff back through `isValid`. Eleven mutations, eleven caught, each by the gate that claims the property — including swapping the diff back to `isValid`, which the annotation gate catches, and refusing the `namespace::` axis, which the retirement gate does.

Closes #680


maxonfjvipon:
The round-trip half is solid and every structural number reproduces. One finding
against the other half, and it is the one that decides whether this closes #680.

**1. The acceptance diff runs against the workaround it was meant to indict.**

`isValid` is `compiles(xpath) || compiles(squeezed(xpath))`. The respelling
retry sits on the *engine* side of the comparison, so an expression fontoxpath
refuses and the squeeze rescues cannot surface as a disagreement. It is not
that the corpus holds none — it holds forty, and the comparison cancels them.

Measured over the same 607 expressions, with the repo's own `compiles`:

```text
isValid (as this PR does)   disagreements  11  ->  grammar accepts   1, engine accepts 10
compiles (no retry)         disagreements  51  ->  grammar accepts  41, engine accepts 10
```

The forty extra are the #639 family exactly — `child ::gamma`,
`parent :: node ( )`, `ancestor  ::  b`, `attribute::  spaced`, a `child::`
holding a tab, and every `namespace::` axis (`namespace::*`,
`count(namespace::*)`, `/o[namespace::*]`, ...). The grammar accepts all forty
and is right to; the engine refuses them. That is the Phase 3 evidence, and it
is already in the tree.

So the sentence

> It collects none, because no fixture holds an expression the engine is wrong
> about, so that evidence has to come from somewhere else

reads as a fact about the fixtures when it is a consequence of which side the
diff was taken against. Both gates seem worth keeping: the `isValid` diff
answers whether swapping the parser in would move a verdict today, and a
`compiles` diff answers what #680 actually asked.

**2. A docblock contradicted by the row eight lines below it.**

`GAPS` opens with `{xpath: 'my:25l(3)', accepts: 'grammar'}`, and the docblock
above it says "None is the engine refusing a spelling the grammar allows".
That row is precisely that. The intended sense is plainly "none is the engine
*wrongly* refusing" — but given the finding above, even that reading needs the
qualifier that the diff is taken against the retry.

**3. The corpus floors carry a third more slack than the prose claims.**

`REACHES` is 120/300/40 against actuals of 177/431/51, so a source can lose 57,
131 or 11 expressions and stay green. The docblock and the `CLAUDE.md` row both
say each source "must still yield what it did when the gates were written". A
floor with that much slack is not what it did.

---

- **607 distinct reproduces exactly**, and so does the split — 177 stylesheet,
  431 pack, 51 selector. 2278 nodes walked; the round trip, the span slicing and
  the child nesting all hold over every one of them.
- **Eleven disagreements, ten of them a selector, one expression in 11.9** — every
  figure exact.
- **The #711 characterisation is precisely right.** `(a|b)/c` parses at every
  version while `a/(b|c)`, `//(a|b)` and `/(a|b)` are refused at every version.
  That is XPath 1.0's rule, and a parenthesized step is legal from 2.0 on.
- **Eight mutations, eight caught**, each by the gate that claims the property:
  trivia dropped from the stream, a node started a token early, one ended a token
  late, a `GAPS` row dropped, a row annotated on the wrong side, and each of the
  three sources cut out in turn.
- 1353 passing (896 fast, 457 deep), coverage clean, `npx eslint .` clean,
  eighteen checks green, and both #708 and #711 are open where the annotations
  send a reader.

maxonfjvipon:
All three land, and the first one is not a wording fix — the diff was taken against the workaround, so the class #680 was written to collect was cancelled by construction. Fixed in `edfb15c`.

**1.** Reproduced exactly, including the split:

```text
                            disagreements   grammar accepts   engine accepts
isValid  (as it was)             11                 1               10
compiles (as it is now)          51                41               10
```

The forty in between decompose cleanly, which turned out to be worth more than forty annotation lines:

```js
const RESCUED = CORPUS
  .filter((one) => !compiles(one) && compiles(squeezed(one)))
```

Forty-one expressions the bare engine refuses and the grammar accepts; forty of them are exactly `RESCUED`, and the remainder is `my:25l(3)`. The grammar refuses **none** of the forty. So rather than list them, the gate holds the grammar to the set — the parser already accepts every spelling the retry exists to rescue, which is the Phase 3 evidence stated as a property instead of an inventory. `RESCUED` is read from the engine and `squeezed` alone and never from the grammar, so it is a comparison and not a restatement, and a guard fails if the retry ever has nothing to rescue and leaves that gate holding air.

`PARTED` now diffs against `compiles`, and the expected set is `GAPS` plus `RESCUED`. That also subsumes the `isValid` diff rather than dropping it: for an expression the retry does not rescue the two verdicts are identical, so `GAPS` *is* the isValid-level diff, and asserting it once avoids the pair of assertions that cannot fail independently.

**2.** Right, and literally so — `accepts: 'grammar'` on that row is the very shape the sentence denied. The paragraph is rewritten; the qualifier it needed was never "wrongly", it was that the comparison could not see the class it was denying. What is true of `my:25l(3)` is that the engine is *correct* to refuse it, and it says that now.

**3.** 120/300/40 against 177/431/51 was slack the prose did not admit to. The floors are the actuals now, so a fixture leaving the sweep is a visible edit rather than an erosion, and `CLAUDE.md` says the same thing.

Re-proved after the change: 1355 passing, 100% branch coverage, `npx eslint .` clean. Eleven mutations, eleven caught — the eight from before plus making the retry a no-op, taking the `namespace::` axis away from the grammar (caught by the new retirement gate), and putting the diff back through `isValid`, which the annotation gate catches. That last one is the finding, wired shut.

PR #713, opened 2026-08-08
Let a step be parenthesized, as 2.0 allows

### LINKED ISSUE #711: A parenthesized step is refused after the first in `src/grammar.js`

maxonfjvipon:
`walked` read the first part of a path with `postfixed` and every part after a
slash with `stepped`, so a parenthesized expression was legal where it opened a
path and refused everywhere after it. `(a|b)/c` parsed; `//(a|b)` and `x/(a|b)`
did not.

That asymmetry is XPath 1.0's rule and only 1.0's — there a `FilterExpr` may open
a path and stand nowhere else — while 2.0 generalised the step itself to
`StepExpr ::= PostfixExpr | AxisStep`. The version was already in hand, so the
older shape is kept where it belongs rather than imposed on the versions that
outgrew it:

```js
const parted = function(cursor) {
  let node = undefined
  if (since(cursor.version, MODERN)) {
    node = postfixed(cursor)
  } else {
    node = stepped(cursor)
  }
  return node
}
```

`//(a|b)` now parses at 2.0 and above and is still refused at 1.0, where
`(a|b)/c` still parses. The same generalisation admits `a/$v` and `a/"x"` from
2.0, which is what `PostfixExpr` means there.

One thing found on the way, and fixed with it because it is the same production:
`a/.` was refused at every version. `.` is an `AbbreviatedStep` in XPath 1.0 as
much as `..` is, and `stepped` handled only the second.

The tree it builds is *not* the same on both sides, and an earlier draft of this
paragraph claimed it was. 1.0 gives `path(step, step)` and 2.0 gives
`path(step, context)`, because 1.0 spells `.` as an `AbbreviatedStep` and 2.0
reaches it as a `ContextItemExpr` through `PostfixExpr`. That is spec-faithful
and stays, but it is worth stating rather than smoothing over: at Phase 3 a
reader asking whether a step is a dot has to accept both kinds, and which one
arrives depends on the stylesheet's version.

The reach is this project's own checks. Nine selectors are written in the idiom
that was refused — `//(xsl:variable | xsl:param)[not(@name)]` and
`//xsl:number/(@count | @from)` among them — and at Phase 3 of #644 each would
have become an `invalid-xpath-expression` reported against valid XPath, nine
checks disabled by their own selector. Sweeping every selector in
`checks.json`: **53 of 54 parse**. The one left is `$var << .`, the node
comparison, which has no token kind and is already recorded on #708 — a
different gap from this one, and a lexer change rather than a grammar change.

Eleven rows pin it. Seven were `ACCEPTS`, which only ever asks at 3.0, so review
was right that nothing failed if the 1.0 restriction were deleted — the whole
suite stayed green with `parted` ungated. Four `GATED` rows now ask both sides,
and `a/(b|c)` is the one that discriminates: `//(a|b)` does not, because at 1.0
that refusal comes from the leading-slash branch rather than from `parted`. The
slash has to be inside the path. Ungating `parted` fails 4.

The dot needed pinning in two places, not one, and only the second was obvious.
`stepped`'s `TOKENS.DOT` branch was dead in the suite for the same reason — it is
reachable only at 1.0, since from 2.0 `parted` routes through `postfixed` to the
context-item primary, and nothing asked for `a/.` below 2.0. Branch coverage read
100% throughout, which is the `||`-operand blind spot: the second operand is
evaluated on every dotless step and counts as covered without ever being taken.
One assertion at 1.0 closes it, and dropping the branch now fails 1.

The `TOKENS.DOT` in `steps` is a *different* clause and stayed unpinned after
that, because `steps` is consulted only right after a leading slash and the new
assertion reaches its dot through the while loop instead. It is load-bearing all
the same: `/.`, `//.`, `/./a` and `/.//a` are valid XPath the engine accepts, and
all four are refused at every version if the clause goes. Four `ACCEPTS` rows now
hold them, and dropping the clause fails 4.

The `since(...) &&` that guarded the openers in `steps` is gone. It changed no
verdict anywhere — only the wording of a fault, and `/(a|b)` at 1.0 now reads
`expected a name at 1` rather than `expected the end of the expression at 1`,
which is the more useful of the two for whoever reads a #680 report. The version
is tested at the one place it decides an answer.

**This and #712 cannot land independently.** Nine of #712's `GAPS` rows go stale
the moment this merges, so whichever lands second leaves master red until that
table is trimmed. I am happy to trim it in whichever order suits — say the word
and I will either add the trim here or send it as a follow-up to #712.

Verified: 903 fast + 457 deep, coverage 1360 passing at 100% of statements
(7490), branches (1188), functions (241) and lines; ESLint clean.

Closes #711.


maxonfjvipon:
The grammar change is right — the version table matches both specs and the
corpus confirms it (details at the bottom). What is missing is a test of the
distinction the change exists to draw.

**1. The version gate is unpinned.**

Making `parted` always take the modern shape — so `//(a|b)` and `a/$v` parse at
1.0 — leaves the whole suite at 1360 passing. All seven new rows go into
`ACCEPTS`, which only ever calls `parsed(xpath, '3.0')`, and no `GATED` row was
added, which is the one table that asks about both sides. So

> Seven rows pin it, covering both sides of the version line

is not what the rows do: nothing here fails if the 1.0 restriction is deleted.

One row closes it, though not the obvious one. A row spelled `//(a|b)` does not
discriminate, because at 1.0 that refusal comes from the leading-slash branch
rather than from `parted`. It has to be a slash inside the path:

```js
  {xpath: 'a/(b|c)', floor: '2.0', below: '1.0'},
```

Green as shipped, one failing with the gate ungated. `a/$v` and `a/1` serve as
well; `//(a|b)` and `//(a)` do not.

**2. The new `TOKENS.DOT` branch in `stepped` is dead in the suite.**

Made to `throw`, it leaves 1360 passing; dropped, the same. It is reachable only
at 1.0 — from 2.0 `parted` routes through `postfixed` to the context-item
primary — and nothing parses `a/.` below 2.0. Branch coverage still reads 100%
of 1188, which is the `||`-operand blind spot: the second operand is evaluated
on every dotless step and so counts as covered without ever being taken. One
assertion closes it, and fails when the branch goes:

```js
    assert.equal(parsed('a/.', '1.0').fault, '', '...')
```

**3. "The tree has one shape for it either way" is the opposite of what happens.**

```text
"a/."    1.0: path(step, step)    2.0: path(step, context)
```

The split is spec-faithful — 1.0 spells `.` as an `AbbreviatedStep`, 2.0 as a
`ContextItemExpr` reached through `PostfixExpr` — so the code is right and the
sentence is wrong. Worth correcting rather than ignoring: at Phase 3 a reader
asking whether a step is a dot has to accept both kinds, and which one arrives
depends on the stylesheet's version. That is the asymmetry the sentence says was
removed.

**4. Half of the version handling cannot change an answer.**

The `since(cursor.version, MODERN) &&` in `steps` changes no verdict anywhere.
Over 54240 generated expressions at three versions: zero verdict differences
with it removed, and 2960 differences in the fault text, every one at 1.0 —
`/(a|b)` reads `expected the end of the expression at 1` with the guard and
`expected a name at 1` without it. So no verdict test can pin that half, and of
the two messages the ungated one seems the more useful to whoever reads a #680
report. Worth either dropping or pinning by its message.

**5. This and #712 cannot land independently.**

#712's corpus gate run against this branch gives 6 passing, 1 failing — `cannot
part from the engine anywhere it is not annotated` — since nine `GAPS` rows go
stale at once. The note that the annotation "can come off when both land" is
right, but whichever merges second leaves master red until `GAPS` is trimmed.

---

- **The version table matches XPath 1.0 and 2.0 exactly.** `(a|b)/c` and
  `(a|b)//c` parse at every version; `//(a|b)`, `/(a|b)`, `a/(b|c)`,
  `a//(b|c)`, `a/$v`, `a/"x"`, `a/1` and `a/f(1)` are refused at 1.0 and
  admitted from 2.0; `a/[1]` and `a/?b` stay 3.0-only behind their own gates.
- **Nothing regressed after a slash.** `a/text()`, `a/node()`, `a/comment()`,
  `a/processing-instruction("x")`, `a/*`, `a/@b`, `//text()` and
  `a/child::text()` all still parse at 1.0, so no node test was mistaken for the
  function call the new opener admits.
- **53 of 54 selectors in `checks.json` parse**, and the one left is the `<<`
  node comparison already recorded on #708.
- **The corpus diff drops from eleven disagreements to two**, over the same 607
  expressions #712 sweeps, with no new one appearing — exactly the nine
  selectors this closes, leaving the two on #708.
- 903 fast, 457 deep, coverage 1360 at 100% of statements, branches, functions
  and lines, `npx eslint .` clean, eighteen checks green.

PR #714, opened 2026-08-08
Make the duplicate-reference deletion a suggestion, and cut the shadowed reference

### LINKED ISSUE #667: `redundant-import` offers a safe fix that moves import precedence in `src/import-linter.js`

maxonfjvipon:
`redundant-import` reported the second and later reference to a module and offered a safe deletion of it. Import precedence is positional, though — XSLT 1.0 §2.6.2 ranks each `xsl:import` above everything declared before it — so a module's level is set by its *last* reference, and that is the one reference that cannot go. With another module imported in between, cutting it drops this one below that one, and the stylesheet's output changes under a fix tier that promises it will not.

The report moves to the other end. Every reference but the last is a defect now, and the last is what stays:

```js
const last = new Map()
imports.forEach(({file, to}, at) => last.set(`${file}|${to}`, at))
const defects = []
imports.forEach(({file, node, to}, at) => {
  const key = `${file}|${to}`
  if (last.get(key) !== at) {
```

That much is only about which reference to cut. It does not make the deletion safe at either mechanism, and the fix is a suggestion whatever it cuts.

An `xsl:import` creates a precedence level, and the earlier reference is shadowed for template *selection* only: `xsl:apply-imports` walks down the chain and meets the module at every level it occupies. An `xsl:include` creates no level of its own — but it copies the includee's children in, and those children include its own `xsl:import` elements, so a module included twice brings its imports in twice. Measured with xsltproc, `gamma.xsl` importing `delta.xsl` and both rules ending in `xsl:apply-imports`:

```text
                        before   plain --fix   --fix-suggestions
import alpha twice      AA       AA            A
import alpha, beta, alpha ABA    ABA           AB
include gamma twice     GDD      GDD           GD
```

Plain `--fix` moves no output now.

Gating on the corpus was the tempting alternative and does not work in either direction. The permissive form — safe unless the corpus holds an `xsl:apply-imports` — produces a false safe, since `--fix` cuts a duplicate import of a module xslint never reads. The conservative form — safe only when the includee resolves and its closure imports nothing — is sound but empty: not one of this repository's own duplicate-reference fixtures resolves, so the safe tier would fire nowhere, and in the field it would depend on which files the caller passed rather than on the stylesheet.

`test/resources/fix/redundant-import-between.xsl` is the case the existing packs could not reach — `alpha`, `beta`, `alpha` — and `--fix` now cuts line 7 and leaves `beta`, `alpha`, so `alpha` still outranks `beta`. Before this it cut line 9 and reversed them. The three packs that pinned the old end shifted with it, including `mixed-import-and-include`, where a module reached both ways is still reported without a fix.

The motive taught the opposite of what the check flags — "the second `xsl:import` brings in the same templates the first already did" — so it now teaches both halves: the later reference decides which module wins, and the earlier one is still walked by `xsl:apply-imports`, so a duplicate import is a decision to make by hand. `README.md` said the module stays imported by the first reference, and listed the whole check under `--fix`; the include half stays there and the import half moves to the suggestions list.

Closes #667

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,9 @@ Today this covers ten checks:
   and `not($x)`.
 - `redundant-import` — a duplicate `xsl:import`/`xsl:include` of a module
   already imported in the same stylesheet is removed; the module stays imported
-  by the first reference. A module reached *both* ways — imported and also
+  by the *last* reference, which is the one that fixes its import precedence,
+  so the references cut are the earlier ones and nothing changes level. A
+  module reached *both* ways — imported and also
   included — is reported without a fix, since `xsl:import` and `xsl:include`
   differ in import precedence, so which of the two to drop is the author's
   decision rather than a deletion the tool can make.
--- a/README.md
+++ b/README.md
@@ -276,14 +276,13 @@ Today this covers ten checks:
   `exists($x)` and `count($x) = 0` becomes `empty($x)`; on 1.0 (where those
   functions don't exist), `boolean($x)` — or a bare `$x` in a whole `@test` —
   and `not($x)`.
-- `redundant-import` — a duplicate `xsl:import`/`xsl:include` of a module
-  already imported in the same stylesheet is removed; the module stays imported
-  by the *last* reference, which is the one that fixes its import precedence,
-  so the references cut are the earlier ones and nothing changes level. A
-  module reached *both* ways — imported and also
-  included — is reported without a fix, since `xsl:import` and `xsl:include`
-  differ in import precedence, so which of the two to drop is the author's
-  decision rather than a deletion the tool can make.
+- `redundant-import`, on an `xsl:include` — a repeated include of one module is
+  removed. An include creates no import-precedence level of its own, so the
+  repeat is pure noise and dropping it changes nothing. The `xsl:import` half is
+  a suggestion, below. A module reached *both* ways — imported and also
+  included — is reported without a fix at either tier, since `xsl:import` and
+  `xsl:include` differ in import precedence, so which of the two to drop is the
+  author's decision rather than a deletion the tool can make.
 - `starts-with-double-slash`, outside an `xsl:template` — the redundant leading
   `//` of a pattern is dropped: `match="//keyed"` on an `xsl:key` becomes
   `match="keyed"`. Only a template's pattern is ranked by priority, so on
@@ -322,6 +321,13 @@ Today this covers:
 - `incorrect-use-of-boolean-constants` — the string test `'true'`/`'false'`
   becomes `true()`/`false()`, which changes the truth value of a `'false'` test
   (a non-empty string is always true).
+- `redundant-import`, on an `xsl:import` — a repeated import of one module is
+  removed, and the reference cut is an earlier one, so the module keeps the
+  precedence its last reference gives it. A suggestion because no choice of
+  reference makes the deletion behaviour-preserving: importing a module twice
+  puts it at two precedence levels, and `xsl:apply-imports` (like 2.0's
+  `xsl:next-match`) walks that chain and meets it at both, so a module whose
+  rules reach back down it runs once per reference.
 - `select-starts-with-double-slash` — the leading `//` of a `@select` is
   anchored as `.//`: `select="//title"` becomes `select=".//title"`, scanning
   descendants of the context node instead of the whole document.
--- a/README.md
+++ b/README.md
@@ -321,11 +321,11 @@ Today this covers:
   puts it at two precedence levels, and `xsl:apply-imports` (like 2.0's
   `xsl:next-match`) walks that chain and meets it at both, so a module whose
   rules reach back down it runs once per reference. An `xsl:include` creates no
-  level of its own but copies in the includee's own `xsl:import` elements, so a
-  repeated include duplicates those the same way. A module reached *both*
-  ways — imported and also included — is reported without a fix at all, since
-  which of the two to drop is a decision about precedence rather than a
-  deletion the tool can make.
+  level of its own but copies in the included module's own `xsl:import`
+  elements, so a repeated include duplicates those the same way. A module
+  reached *both* ways — imported and also included — is reported without a fix
+  at all, since which of the two to drop is a decision about precedence rather
+  than a deletion the tool can make.
 - `select-starts-with-double-slash` — the leading `//` of a `@select` is
   anchored as `.//`: `select="//title"` becomes `select=".//title"`, scanning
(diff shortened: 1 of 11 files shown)


maxonfjvipon:
You are right, and the branch as it stood was wrong for the reason you give. Fixed in `f48438e`.

Reproduced exactly, with both modules ending their rule in `xsl:apply-imports`:

```text
alpha              A
alpha, alpha       AA
alpha, beta, alpha ABA
beta,  alpha       AB      <- what the branch's --fix produced from ABA
gamma, gamma       G       <- include: no level, no change
```

"Fully shadowed and had no effect to preserve" was the sentence that carried the error. An earlier reference is shadowed for template *selection*; it is not shadowed for the traversal, and `xsl:apply-imports` walks down the chain and meets the module at every level it occupies. The `alpha, alpha` case makes the point without any of the three-import shape the branch was built around, exactly as you say.

I took the first remedy — fork the tier on the mechanism — and against the second, for a reason worth recording. The gate on "the corpus holds an `xsl:apply-imports`" is unsound, because the duplicated module need not be in the corpus at all:

```text
  <xsl:import href="/opt/libs/outside.xsl"/>
  <xsl:import href="/opt/libs/outside.xsl"/>
```

Plain `--fix` cuts that today, and xslint never reads `/opt/libs/outside.xsl`, so the gate would answer "no apply-imports anywhere" and call the deletion safe while the library uses one internally. That is a false safe on the tier that runs with no opt-in, which is worse than a demotion — and it would also need a complete list of every precedence-sensitive construct, with `xsl:next-match` only the one we know to name.

So `removal` reads the mechanism, which as you note is already in hand:

```js
suggestion: node.localName === 'import',
```

Through the CLI, against xsltproc:

```text
                     before   plain --fix   --fix-suggestions
alpha, alpha         AA       AA            A
alpha, beta, alpha   ABA      ABA           AB
include gamma twice  G        G             G
```

Plain `--fix` moves no output, and still deletes the duplicate include — `gg.xsl` goes 2 includes to 1 in that same run while `aa.xsl` stays at 2 imports.

The "cut the earlier reference" half of the branch stays, since a suggestion should still be the least damaging correct edit: cutting the last one changes which module wins on top of everything above. `repeated-include.{xsl,fixed.xsl}` is new so the safe half is pinned by a fixture rather than only by the demoted one, the three import rows moved to `--fix-suggestions`, and an `UNCHANGED` row now asserts plain `--fix` leaves a duplicate import alone. The motive's "nothing changes level" claim is gone and it teaches the traversal; `README.md` splits the check across both tier lists.

1362 passing, 100% branch coverage, xcop 265, lint clean.

maxonfjvipon:
Confirmed, and the fixture I added was exactly the shape that cannot show it. `9fd1638` demotes both.

```text
  include gamma once   GD
  include gamma twice  GDD
```

`gamma.xsl` imports `delta.xsl`, both rules end in `xsl:apply-imports`. So "an include creates no precedence level" was true of the includee and false of what it carries — the children copied in include its own `xsl:import` elements. One indirection out, same class, as you say.

I took the second remedy rather than the conservative gate, and the measurement is what decided it. The gate is sound — not-knowing costing a suggestion cannot produce a false safe, and `importsOf`/`graphOf` would answer it — but it would fire nowhere. None of this repository's duplicate-reference fixtures resolves: `common.xsl`, `shared.xsl`, `alpha.xsl`, `beta.xsl`, `other.xsl`, `library.xsl` are all hrefs pointing outside the corpus, `common.xsl` included, so `repeated-include.xsl` itself would demote under it. A safe tier with no test reaching it is closure machinery, cycle handling and a corpus-membership test kept alive for nothing.

The second reason is worse than the first, and it is what settled it: whether a module resolves depends on which files were passed. `xslint sheet.xsl` and `xslint .` would then classify one stylesheet's fix differently — the tier would describe the invocation rather than the code, and a user tidying one file at a time would get the safe deletion precisely where least is known about it.

So `removal` carries `suggestion: true` outright, and the docblock says why for both mechanisms rather than forking. Through the CLI, against xsltproc:

```text
                          before   plain --fix   --fix-suggestions
import alpha twice        AA       AA            A
import alpha, beta, alpha ABA      ABA           AB
include gamma twice       GDD      GDD           GD
```

Plain `--fix` moves no output at all now. `repeated-include` moved to `--fix-suggestions` and gained an `UNCHANGED` row proving plain `--fix` leaves it, beside the one for the import. The motive's "a repeated include carries none of this" paragraph is replaced by the `GDD` case, and `README.md` drops the check from the safe list — nine there now, not ten — with one entry in the suggestions list covering both mechanisms.

Worth stating plainly, since two rounds of this branch have claimed otherwise: `redundant-import` has no safe fix, and the "cut the earlier reference" half stays only because a suggestion should still be the least damaging correct edit — cutting the last one changes which module wins on top of everything else.

1363 passing, 100% branch coverage, xcop 265, lint clean.

PR #716, opened 2026-08-08
Give each stage a directory, so the linter boundary is a rule and not a habit

### LINKED ISSUE #715: The linter boundary `CLAUDE.md` states cannot be enforced, since `src/` holds linters beside the core
`CLAUDE.md:188` states a rule that holds today and that nothing can check:

> No linter imports another.

It is true — no `src/*-linter.js` requires another one — and it will stay true only
for as long as somebody remembers it. The reason it cannot be enforced is the
layout. `src/` is one flat directory of 42 entries where the 16 code-based
linters and the 2 validators sit among the 24 core modules they consume, so
`require('./count-linter')` and `require('./comparisons')` are the same shape of
string, written from the same place, and no ESLint selector can tell a stage from
a shared module. The companion rule is in the same position: the staging is wired
only in `src/xslint.js`, and nothing stops a second file from wiring one too.

What the flat directory looks like from the one file that legitimately reads
every stage, `src/xslint.js:8`:

```js
const {allFilesFrom} = require('./helpers')
const {validate: validateXsls, names: xslChecks} = require('./xsl-validator')
const {lintByXpath, names: xpathChecks} = require('./xpath-linter')
const {lintByCorpus, names: corpusChecks} = require('./corpus-linter')
...
const {fixed} = require('./fixer')
const {logger, levels} = require('./logger')
```

Four kinds of module, one prefix. The pipeline this project is built around —
validators partition the input, linters run over what passed — is drawn in
`CLAUDE.md` and nowhere in the tree.

Move the 16 `*-linter.js` into `src/linters/` and `xsl-validator.js` and
`xpath-validator.js` into `src/validators/`, then add two `no-restricted-syntax`
selectors that the directories make expressible: one under
`files: ['src/linters/**']` banning a `require` of any `-linter`, one under
`files: ['src/*.js']` with `src/xslint.js` ignored, banning a `require` of
`./linters/*`. The `-linter` suffix stays, since dropping it would put
`linters/xpath.js` one word from `src/xpath.js`, the fontoxpath environment.

The change is mechanical and the suite covers all of it: 86 relative requires
inside the moved files gain a `../`, 18 paths in `src/xslint.js` and 19 in
`test/` grow one segment, and nothing else moves — `.c8rc.json`,
`eslint.config.mjs` and `Gruntfile.js` all glob `src/**` recursively, `main` and
`bin` point at files that stay, and no linter is ever discovered by filename.
Worth doing before Phase 4 of #644, which rewrites nine of these linters one pull
request at a time and would otherwise land every one of them in a directory they
are about to leave.

maxonfjvipon:
`src/` was one flat directory of 42 entries, with the 16 code-based linters and
the 2 validators sitting among the 24 core modules they consume. Two rules in
`CLAUDE.md` describe that arrangement — no linter imports another, and only
`src/xslint.js` wires one — and both held on habit alone. They could not be
enforced, because `require('./count-linter')` and `require('./comparisons')` are
the same shape of string written from the same place, and no selector can tell a
stage from a shared module.

The linters move to `src/linters/` and the validators to `src/validators/`, and
each rule becomes one selector:

```js
{
  selector:
    "CallExpression[callee.name='require'][arguments.0.value=/-(linter|validator)$/]",
  message:
    "No stage reads another: a linter and a validator take the corpus and the
     shared modules, and the order they run in is src/xslint.js's business alone"
}
```

The second is scoped to `src/*.js` with `src/xslint.js` ignored, and bans a
require of `./linters/` or `./validators/` — so a core module cannot reach back
into the pipeline it exists to serve. Both were probed by writing the violation
and watching ESLint refuse it: a linter requiring a linter, a validator requiring
a validator, a core module requiring a linter, and `src/xslint.js` requiring one
and staying clean. The shared restriction list is a `const` the scoped blocks
spread, so a file under `src/linters/` is still held to the `\s` ban and the rest
rather than quietly losing them to a later block.

The `-linter` suffix stays inside the directory. Dropping it would put
`linters/xpath.js` one word from `src/xpath.js`, the fontoxpath environment, and
those two are already the near-miss in this tree.

Nothing else moved. `.c8rc.json`, `eslint.config.mjs` and `Gruntfile.js` all glob
`src/**` recursively, `main` and `bin` point at files that stayed, and no linter
is ever discovered by filename — so the rest of the change is 86 relative
requires gaining a `../`, 18 paths in `src/xslint.js`, and 19 in `test/`. Worth
landing before Phase 4 of #644, which rewrites nine of these linters one pull
request at a time and would otherwise put each of them in a directory it is about
to leave.

Verified: 1355 passing (898 fast, 457 deep), coverage 100% of statements (7461),
branches (1189), functions (240) and lines, ESLint clean, `--version` and a real
run over `test/resources/fix` both answering as before.

Closes #715.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -161,11 +161,11 @@ src/index.mjs             CLI entry (commander.js, ESM)
 ```
 
 The two stages have a directory each, and everything else in `src/` is the core
-they consume. That is not filing: it is what makes the two rules below
-expressible, which they were not while a stage and a shared module were the same
-kind of string written from the same place (#715). The `-linter` suffix stays
-inside `src/linters/` for one reason — dropping it would put `linters/xpath.js`
-one word from `src/xpath.js`, the fontoxpath environment.
+they consume. That is not filing: it is what makes the rule below expressible,
+which it was not while a stage and a shared module were the same kind of string
+written from the same place (#715). The `-linter` suffix stays inside
+`src/linters/` for one reason — dropping it would put `linters/xpath.js` one word
+from `src/xpath.js`, the fontoxpath environment.
 
 `src/xslint.js` exposes the whole staging as a pure function,
 `lint(sources, {suppress, overrides}) => defects`: no file I/O, prints nothing,
@@ -194,12 +194,17 @@ test pack:
 | `validation` | `severity` + `message` | code (well-formedness, XPath syntax) | in code |
 | `format` | `severity` + `message` | code (a `src/linters/*-linter.js`) | in code |
 
-No linter imports another, and no validator imports another — a
-`no-restricted-syntax` selector scoped to those two directories bans a `require`
-of anything ending in `-linter` or `-validator`, so the rule fails a build rather
-than a reader. The reverse holds too: only `src/xslint.js` may reach into either
-directory, which a second selector enforces over `src/*.js` with that one file
-ignored. Both were true for a year and enforced by nothing (#715). The
+No linter imports another, no validator imports another, and only
+`src/xslint.js` reaches into either directory. That is one rule rather than
+three: a `no-restricted-syntax` selector over all of `src/`, with that one file
+ignored, refuses a `require` or `import` of any path ending in `-linter` or
+`-validator`. It asks what is being required, never where the requiring file
+sits, which is what makes it hold — the first draft anchored on the requirer's
+directory and three spellings walked past it, an extension (`./name-linter.js`),
+a longer way to the same file (`../src/linters/name-linter`, which resolves), and
+any new `src/` subdirectory at all, since `src/*.js` does not recurse (#715). A
+barrel cannot get around it either: an `src/linters/index.js` would have to
+require the linters it re-exports, and that is the same violation. The
 declarative loaders (`xpath-linter`,
 `corpus-linter`) and the token/DOM linters all share `src/xpath.js` (the
 fontoxpath environment: prefixes, evaluators, `isValid`), `src/tokens.js` (the
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -96,9 +96,10 @@ call that omits it.
 
 A function likewise leaves through exactly one `return`, and the branching — not
 the exit — decides what it carries: a lookup keyed on the deciding values
-(`collapses` in `src/count-linter.js`), a binding an `if`/`else if` chain settles
-before the exit (`defectsOf` in `src/corpus-linter.js`, which picks the strategy
-rather than the answer), or a sentinel a scan assigns before its loop ends
+(`collapses` in `src/linters/count-linter.js`), a binding an `if`/`else if`
+chain settles before the exit (`defectsOf` in `src/linters/corpus-linter.js`,
+which picks the strategy rather than the answer), or a sentinel a scan assigns
+before its loop ends
 (`closes` in `src/expressions.js`). A `return` counts against the nearest
 enclosing function, so a `map`/`every` callback carrying its own single `return`
 is fine, and an arrow with an expression body has none at all.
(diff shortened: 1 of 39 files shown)


maxonfjvipon:
The move is clean and the mechanical claims all hold — checked rather than read,
below. One finding against the enforcement, which is the part the ticket is
about.

**The rule is still a habit in three spellings.**

```text
in src/linters/  require("./count-linter.js")            NOT FLAGGED
in src/foo.js    require("../src/linters/count-linter")  NOT FLAGGED
in src/newdir/   require("../linters/count-linter")      NOT FLAGGED
```

The first walks through the `-(linter|validator)$` anchor by writing the
extension, the second through the `^\./(linters|validators)/` anchor by
spelling the path another way, and the third is covered by neither block: the
core rule is `src/*.js`, which does not recurse, and the stage rule only reaches
`src/linters/**` and `src/validators/**`. A file in a new `src/` subdirectory
inherits the shared bans and none of the isolation — which is worth closing
before Phase 4 of #644 rather than after, since that is when a new subdirectory
is most likely to appear.

All three close together, and the result is smaller than what is there now: one
block rather than two, banning the stage require everywhere under `src/` and
exempting the one file allowed to wire a stage. The two anchors become one
suffix, and the question stops being *where the requiring file sits* and becomes
*what it requires*:

```js
  {
    files: ["src/**/*.js", "src/**/*.mjs"],
    ignores: ["src/xslint.js"],
    rules: {
      "no-restricted-syntax": ["error", ...RESTRICTED,
        {
          selector:
            "CallExpression[callee.name='require'][arguments.0.value=/-(linter|validator)(\\.js)?$/], ImportDeclaration[source.value=/-(linter|validator)(\\.js)?$/], ImportExpression[source.value=/-(linter|validator)(\\.js)?$/]",
          message:
            "Only src/xslint.js wires a stage: no linter or validator reads another, and no core module reaches into the pipeline it exists to serve (#715)"
        }
      ]
    }
  },
```

Every one of the five probes is flagged with it — the three above plus the two
already caught — and the whole tree still lints clean, tests included, since a
test requiring a linter direct sits outside `src/` and is untouched.

---

- **The move is pure.** Each of the eighteen moved files, compared against its
  master counterpart with `require('./` rewritten to `require('../`: zero differ
  beyond that. Nothing rode along.
- **Nothing fell out of coverage**, which was the quieter risk — a non-recursive
  include would have dropped sixteen linters from the 100% gate without a word.
  `src/**` with `all` recurses, and the report lists every `src/linters/*` and
  both `src/validators/*` at 100%.
- **The shared restrictions really are inherited**, which is the thing flat
  config makes easy to lose: a later block redefining `no-restricted-syntax`
  replaces the earlier list rather than extending it. Spreading `RESTRICTED`
  holds — a `\s` regex, a bare `//@name` and a `documentElement.getAttribute(
  'version')` are each still refused inside `src/linters/` and
  `src/validators/`.
- **Both new rules fire** where they claim: a linter requiring a linter, a
  validator requiring a validator, a core module requiring `./linters/`. And
  `src/xslint.js` wires eighteen stages and stays clean.
- 898 fast, 457 deep, 1355 passing, statements 7461 and branches 1189 at 100%,
  ESLint clean, eighteen checks green — every figure as stated.

maxonfjvipon:
Right on all three, and I reproduced each before touching anything rather than
taking the list on faith. The count is `no-restricted-syntax` hits, so `0` is the
evasion:

```text
require('./name-linter.js')          inside src/linters/   0
require('../src/linters/count-...')  from src/checks.js    0
require('../linters/count-linter')   from src/newdir/      0
```

The second is the one I would have argued was theoretical, so it is worth
recording that it is not — from inside `src/`, that path resolves:

```text
$ node -e "console.log(require.resolve('../src/linters/count-linter'))"
.../src/linters/count-linter.js
```

Your block is in, verbatim. Six spellings now flagged — those three plus the
three that were already caught — `src/xslint.js` still wiring eighteen stages
clean, and the shared list still inherited (a `\s` regex inside `src/linters/` is
refused). Two blocks became one, net eight lines fewer.

The framing is the part I had wrong, and it is worth saying rather than just
taking the patch: anchoring on the requirer's directory means the rule has to
predict where the next file will sit, and a rule that predicts layout is a rule
with a hole per directory nobody has created yet. Asking what is being required
has no such surface. It also closes one I had not thought of — a barrel. An
`src/linters/index.js` would have to require the linters it re-exports, so that
is the same violation, and `require('./linters')` can never resolve to anything.

`CLAUDE.md` says one rule now instead of two, and names the three spellings that
defeated the first draft, so the next reader learns why it is written against the
required path rather than the requiring one.

1355 passing, coverage 100% of statements (7461), branches (1189), functions
(240) and lines, ESLint clean.

PR #733, opened 2026-08-10
Count a call's arguments before rewriting it

### LINKED ISSUE #576: Three checks rewrite a call without counting its arguments
Three code-based checks take the text between a call's brackets as *the*
argument without ever asking whether the call holds one. `fn:count`, `fn:not`
and `fn:boolean` each take exactly one argument in every XSLT version, so a
`count()` is not a count of a node-set and a `not(not())` is not a double
negation — but each check reports one anyway, and attaches a **safe**-tier fix,
applied by plain `--fix` rather than `--fix-suggestions`.

On a `version="2.0"` sheet, one `<xsl:if>` per row:

```text
count() = 0       ->  empty()      fn:empty has no arity-0 signature
count(a,b) = 0    ->  empty(a,b)   nor an arity-2 one
not(not())        ->  <empty>
boolean()         ->  <empty>
not(not(a,b))     ->  a,b
boolean(a,b)      ->  a,b
```

and on `version="1.0"`, `count() = 0` becomes `not()`. Every row goes in
unloadable and comes out unloadable a different way. The two that empty the
attribute are the sharpest, because `<xsl:if test="">` is not an XPath at all
and xslint reports it itself on the very next run:

```text
[ERROR] ...(4:18) An XPath expression is malformed and cannot be parsed.
        Fix its syntax. (invalid-xpath-expression)
```

One run manufactures, in silence, the defect the next run reports.

The argument is sliced without a count in four places. `src/comparisons.js:71`,
the scan `count-linter` shares, hands over whatever stands between the brackets:

```js
const argument = expression.slice(open + 1, close)
```

and `src/linters/count-linter.js:58` takes it at face value, deciding only
whether the *comparison* is an existence test:

```js
const decide = function(operator, zero, argument) {
  const test = collapses(operator, zero)
  let found = null
  if (test) {
    found = {test: test, argument: argument}
  }
  return found
}
```

`src/linters/redundant-double-negation-linter.js:75` does the same for the inner
call, and `src/linters/redundant-boolean-call-linter.js:61` builds its whole
replacement out of it.

A fix is tiered safe on the claim that it is deterministic and
semantics-preserving. On a call of the wrong arity it preserves nothing, because
there was no meaning there to preserve — and the report is wrong before the fix
is, since the construct the check names is not present.

Give each of the three the arity of the one function it is about, and decline
both the report and the fix when the call does not hold it. Counting top-level
arguments is a split of the already-blanked inner text on commas at depth zero,
so a nested `count(f(a,b))` still reads as one. No function catalogue is needed:
each check knows its own function by name.

## Why this ticket was rewritten

It was filed as *"Unregistered function calls pass the XPath validator in
`src/xpath.js`"*, and proposed a second static check reading the `XPST0017`
that fontoxpath already raises, to partition unknown-name and wrong-arity calls
out of the corpus before any fixer reached them. The harm above was its evidence.

That mechanism does not work, and the check it describes must not be built.
`evaluateXPath` does raise `XPST0017` statically — an unknown call in a branch
never taken still throws — but against a registry that is not the XSLT one:

```text
XSLT 3.0:                26 of 28 functions unregistered
  current()  document()  key()  format-number()  system-property()
  element-available()  function-available()  unparsed-entity-uri()
  current-group()  regex-group()  accumulator-before()  json-to-xml()  ...
XPath 3.1 fn/map/array:  14 of 29 unregistered
  sort()  analyze-string()  format-dateTime()  transform()  parse-ietf-date()
  contains-token()  environment-variable()  serialize-adjacent...  ...
```

A check reading that verdict would report `current()`, `key()` and `document()`
— three of the commonest calls in any XSLT 1.0 stylesheet — as errors, along
with every user-defined `xsl:function`. That is inventing a defect against
working code, the direction this repository treats as the dangerous one.

Reporting an unregistered or wrong-arity call *in general* therefore needs a
real function catalogue: the XPath F&O and XSLT function sets per version with
their arity ranges, plus the stylesheet's own `xsl:function` declarations and
any extension functions it binds. That is a substantial piece of work with an
ongoing maintenance cost, it is not what this ticket was scoped or estimated
for, and nothing in the tree needs it today — because the whole user-visible
half of the original ticket, the damage `--fix` does, is closed by counting
arguments in the three checks that rewrite a call, with no catalogue at all.
The remaining half is a reporting gap: a wrong-arity call stays in the corpus
and goes unreported. It is recorded against Phase 3 of #644 rather than left
open here, so the backlog carries one ticket for this rather than two.

Not to be confused with #572, where the `string-length-compared-to-zero` fixer
emits `" = ''"`: `fn:string-length` does have a registered arity-zero signature,
so that input is legal XPath and a different defect.

maxonfjvipon:
`fn:count`, `fn:not` and `fn:boolean` each take exactly one argument in every
XSLT version. The three checks reading them took whatever stood between the
brackets as *the* argument without ever asking whether one was there. So
`count()` was reported as a count of a node-set and `not(not())` as a double
negation, and each attached a **safe**-tier fix — the tier plain `--fix` applies,
not `--fix-suggestions`:

```text
count() = 0     ->  empty()      fn:empty has no arity-0 signature
count(a,b) = 0  ->  empty(a,b)   nor an arity-2 one
not(not())      ->  <empty>
boolean()       ->  <empty>
not(not(a,b))   ->  a,b
boolean(a,b)    ->  a,b
```

and `count() = 0` becomes `not()` on 1.0. Every row goes in unloadable and comes
out unloadable a different way. The two that empty the attribute leave
`<xsl:if test="">`, which xslint reports itself on the very next run as
`invalid-xpath-expression`: one run manufactures, in silence, the defect the
next one reports.

```js
const lone = function(inner) {
  const tokens = tokenized(inner)
  let depth = 0
  let alone = tokens.some((token) => !TRIVIA.includes(token.type))
  for (const token of tokens) {
    if (NESTED.opens.includes(token.type)) {
      depth++
    } else if (NESTED.shuts.includes(token.type)) {
```

It reads the **tokens**, and so takes the text as the source spells it rather
than blanked. Both halves of the question need that. A comma divides only as
`TOKENS.COMMA`, so one inside a literal or a comment is a kind of its own and no
separator — where a character walk had to be handed masked text first. And a
bracket holding only a literal is not an empty bracket, where masking turns
`count('abc')` into a gap and no emptiness test can tell an absent argument from
a blanked one. A first pass did walk characters over masked text and lost exactly
that: `count('abc') = 0`, `boolean('abc')` and `not(not('abc'))` stopped being
reported, all three of which had been reported and correctly fixed. So a bracket
is empty when nothing but `TRIVIA` stands in it, which separates `()` and
`( (: c :) )` from `('')`.

That wanted a notion of which kinds are trivia, and it was **already spelled
twice** — `TRIVIA` in `src/grammar.js` and a `&&` chain inside `tokenized`
deciding the last solid token. Spelling it a third time is the #708 defect one
concept over, so it moves into `src/tokens.js` beside `OPAQUE`, with its own
`no-restricted-syntax` selector refusing a second copy. **The selector found a
fourth**, a `filter` in `src/linters/predicate-position-linter.js` that the
review and I had both missed, and it is load-bearing: removing its filtering
fails 2 tests.

## Why no existing gate could do this

`defect` in `src/checks.js` already withholds a fix where the expression does not
parse (#636), but every one of these parses: `isValid` goes to fontoxpath's
`compileXPathToJavaScript`, which resolves no function signature at all. Asking
the engine one call further does not work either, and that is what rescoped the
ticket. `evaluateXPath` raises `XPST0017` statically — an unknown call in a branch
never taken still throws — but against a registry that is not XSLT's: 26 of 28
XSLT 3.0 functions and 14 of 29 XPath 3.1 ones are absent from it, `current()`,
`key()`, `document()` and `format-number()` among them. A check reading that
verdict would report the commonest calls in any 1.0 stylesheet as errors, along
with every user-defined `xsl:function`. So arity is per function and each check
knows its own by name; a general unregistered-call check needs a real function
catalogue, recorded against Phase 3 of #644 rather than left as a second ticket.

## What this gives up

Four shapes master reports **and fixes correctly** now go quiet, and they should
be read before merging. A binding clause puts its commas at bracket depth zero
inside one argument, so they read as separators:

```text
count(for $x in a, $y in b return $x) = 0        master: empty(...)   here: silent
count(let $x := a, $y := b return $x) = 0        "
count(some $x in a, $y in b satisfies $x) = 0    "
count(every $x in a, $y in b satisfies $x) = 0   "
```

and `boolean(...)`/`not(not(...))` around the same clause with them. The real rule
is that a comma in front of `return` or `satisfies` binds while one behind it
separates — two arguments really do stand in `f(for $x in a return $x, b)` — which
no bracket count can tell apart and `parsed` can by construction.

Deliberately not patched with a keyword scan: `for`, `let`, `some`, `every`,
`return` and `satisfies` all lex as plain `TOKENS.NAME`, identical to the node
test in `a/for`, so a counter opening a binding scope on the word would open on
`count(a/for, b)`, never close, and then rewrite a genuine two-argument call —
trading a withheld report for a bad fix, which is the wrong direction. It is a
note on Phase 4 of #644, where all three checks migrate onto the tree, and it is
pinned meanwhile by `BINDINGS` in `test/expressions.test.js`: four rows that turn
red the moment the gap closes, the way `GAPS` pins the grammar's own
disagreements.

## Tests

A 19-row `ARGUMENTS` table in `test/expressions.test.js` — the empty, gap-only,
comment-only, two- and three-argument cases beside the nested call, predicate, map
constructor, literal and commented comma that must each read as one argument;
`BINDINGS` for the known narrowing; negative rows on all three packs (`count()`,
`count(x, y)`, `0 < count()`, `count( )`, `not(not())`, `not(not( ))`,
`not(not(x, y))`, `boolean()`, `boolean(@a, @b)`); and a nested-comma and a
literal-argument positive added to each of the three positive packs, so a `lone`
that is too strict turns them red.

Verified on the merge with master (`89a3c17`, so #740 and #743 are in): 1301
fast and 470 deep passing with nothing pending, 100% of statements (8549),
branches (1365), functions (274) and lines, ESLint clean at the bumped 10.8.1,
265 xcop assertions passing with none pending, and `grunt checks`/`grunt docs`
leaving the tree unchanged. The fixture report is byte-identical to master's — 340 defects each
side, zero gone and zero new — which is why the suite alone could never have
caught the original bug: it narrows onto shapes no committed fixture holds.

Every new assertion mutation-probed: handing the call sites blanked text again
fails 3, dropping `TRIVIA` from the emptiness test fails 4, forcing `lone` true
fails 6, removing the depth-zero comma fails 8, keeping trivia in the
predicate-position scan fails 2, and restoring `src/grammar.js`'s own `TRIVIA`
turns the new selector red.

Closes #576


maxonfjvipon:
Both findings reproduce, both were mine, and one is now fixed. Verified each
through the CLI against master rather than taking the reading — the shapes and
the direction are exactly as described.

**The literal-only argument is fixed**, and by a slightly different route than
suggested. Rather than have `lone` mask internally, it walks the **token stream**,
which answers both halves at once: a comma divides only as `TOKENS.COMMA`, so one
inside a literal or a comment needs no masking to be harmless, and a bracket is
empty when nothing but `TRIVIA` stands in it. That separates `()` and
`( (: c :) )` from `('')` without the two texts ever having to be carried
side by side.

```js
const lone = function(inner) {
  const tokens = tokenized(inner)
  let depth = 0
  let alone = tokens.some((token) => !TRIVIA.includes(token.type))
  for (const token of tokens) {
```

All five rows are back to master's answers — `count('abc') = 0` -> `empty('abc')`,
`boolean('abc')` -> `'abc'`, `not(not('abc'))` -> `'abc'`, and `count('')` and
`count('a, b')` with them — while the #576 shapes stay silent and the nested-comma
positives still fire.

That needed a notion of "which kinds are trivia", and it turned out to be
**already spelled twice**: `TRIVIA` in `src/grammar.js:17`, and as a `&&` chain
inside `tokenized` deciding the last solid token. Spelling it a third time is the
#708 defect one concept over, so it moves into `src/tokens.js` beside `OPAQUE`,
with its own `no-restricted-syntax` selector refusing a second copy. **The
selector found a fourth**, a `filter` in
`src/linters/predicate-position-linter.js:101` that neither of us had spotted.

**The binding clause is not fixed, and I agree it wants the tree.** Worth
recording *why* a cheap patch is wrong rather than merely deferred: `for` is a
`TOKENS.NAME` like any other, so a counter opening a binding scope on the word
would open on the node test of `count(a/for, b)`, never close, and then rewrite a
genuine two-argument call — trading a withheld report for a bad fix. So no new
ticket: it is a note on Phase 4 of #644, where `src/comparisons.js`,
`redundant-double-negation` and `redundant-boolean-call` all migrate onto
`parsed`, and it is **pinned** meanwhile by `BINDINGS` in
`test/expressions.test.js` — four rows that turn red the moment the gap closes,
the way `GAPS` pins the grammar's own disagreements.

Re-verified: 1116 fast and 470 deep passing with nothing pending, 100% of
statements (8188), branches (1293), functions (262) and lines, ESLint clean,
`grunt checks`/`grunt docs` no drift, and the fixture diff against master still
340 rows to 340, zero gone and zero new. Every new assertion mutation-probed:
handing the call sites blanked text again fails 3, dropping `TRIVIA` from the
emptiness test fails 4, forcing `lone` true fails 6, removing the depth-zero
comma fails 8, and keeping trivia in the predicate-position scan fails 2.

Also confirming your last point independently: no over-acceptance. Nothing in the
sweep, the packs, or the 340-row fixture diff has `lone` calling a genuine
multi-argument call lone.

